Issue #152, opened 2022-12-07
nifgen_standard_function TestStand sequence should split up initiate/abort between different TS steps


bkeryan:
Here's how I intended to use the `nifgen_standard_function` service in TestStand. It involves multiple steps:

Generate a waveform with duration=0 abort_when_done=false
Wait 10 sec (using Synchronization >> Wait step) - placeholder for other measuremnets
Generate a 0 V DC waveform with duration=0 abort_when_done=true

However, now that I can finally try this out, I see it doesn't work.

```
2022-12-07 15:41:27,358 INFO: Starting generation: pin_name=Pin1 waveform_type=DC frequency=1e+06 amplitude=0
2022-12-07 15:41:27,404 ERROR: Exception calling application: -1074126847: The operation cannot be completed because the device is not configurable while it is generating a signal.

Property: Output Mode
``` 

[AB#2267378](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2267378)

bkeryan:
I still think this would be a useful capability, but it currently doesn't fit very well into a MeasurementLink measurement service because it requires keeping the session running after unreserving it, which is only possible when the session lifetime is managed by a TestStand sequence. When calling the measurement from InstrumentStudio, each call to measure() creates and destroys a temporary session, so the session cannot keep running after the measure function returns.

PR #212, opened 2023-01-19
Removing abort_when_done from FGEN example.

DelpireNI:

Removes the abort_when_done parameter from our FGEN example measurement. We tested with hardware and determined that this parameter did not have any effect. 

I also updated the measui to remove the control for this parameter, and I tried updating the seq file. **Can someone else pull this down and validate it on their machine?** The diff is gnarly.

We shouldn't ship an example with a non-functional parameter.

Successfully ran TestStand sequence using simulated PXIe-5423 (2CH).

[AB#2273840](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2273840)

### DIFF
--- a/examples/nifgen_standard_function/NIFgenStandardFunction.measui
+++ b/examples/nifgen_standard_function/NIFgenStandardFunction.measui
@@ -1,12 +1,12 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<SourceFile Checksum="781ED04CA16AF5E767AFDCBC0E9E947C69A3A5949C2908D7BE68B1087E3053FCB5824887C9C1382004956EC0D4B226C7823D99C14AA36630E86DC0AEA78DD2C5" Timestamp="1D9069B020F3E87" xmlns="http://www.ni.com/PlatformFramework">
+<SourceFile Checksum="474328318FDB50FC3F99B7181C2B6594B56F0EAC83C80A81294CE37750F1F978E11E6AFED7D35AB729CAFF1FF7C9790DD5BB07D972A657F0B5907BC24BF1964E" Timestamp="1D92C1D54FB2F64" xmlns="http://www.ni.com/PlatformFramework">
 	<SourceModelFeatureSet>
-		<ParsableNamespace AssemblyFileVersion="9.7.0.2171" FeatureSetName="Configuration Based Software Core" Name="http://www.ni.com/ConfigurationBasedSoftware.Core" OldestCompatibleVersion="6.3.0.49152" Version="6.3.0.49152" />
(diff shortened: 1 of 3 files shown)


bkeryan:
We should decide whether to close https://github.com/ni/measurementlink-python/issues/152 . This doesn't make my intended use case work, but it at least removes my non-working implementation of that use case.

bkeryan:
When I diff the TS sequence I get

```

DelpireNI:
> When I diff the TS sequence I get
> 
> ```
> Type 'NI_MeasurementParameter' in NIFgenStandardFunction.seq (D:\dev\measurementlink-python\examples\nifgen_standard_function\NIFgenStandardFunction.seq) is different from the currently loaded type 'NI_MeasurementParameter'.
> 
> Currently Loaded Type
> Version 21.0.5.4
> Last modified: 11/21/2022 8:48:19 AM (Older)
> Category: UserTypes
> Location(s): NIFgenStandardFunction.seq
> 
> Type in File
> Version 21.0.5.4 (Modified)
> Last modified: 1/11/2023 1:24:55 PM (Newer)
> Category: UserTypes
> Location: NIFgenStandardFunction.seq
> 
> Summary of Differences:
> 
> NI_MeasurementParameter (Value, Flag, or Representation Difference)
> ```

@bkeryan I updated the seq file using the latest build which included the change to lock the step type with a password. It looks like that resulted in a change in the seq file. I think that's why its saying `NI_MeasurementParameter ` is different. I was talking to @dixonjoel  about this previous. I think we should probably save all the .seq files with our final build once that's available.

dixonjoel:
> We should decide whether to close #152 . This doesn't make my intended use case work, but it at least removes my non-working implementation of that use case.

So does the generation just keep running indefinitely?

DelpireNI:
> > We should decide whether to close #152 . This doesn't make my intended use case work, but it at least removes my non-working implementation of that use case.
> 
> So does the generation just keep running indefinitely?

The opposite. Whether checked or unchecked, I always lost my sine wave after the time elapsed.

bkeryan:
Other than the type version, the TS diff looks ok:
image

bkeryan:
> Edit: I guess if that's the only change, should I just revert it? Its unclear to me whether the boolean appears because its saved to the seq file, or if TestStand will just remove it because it is no longer a parameter.

TS won't automatically remove measurement service parameters. You had to click the "refresh parameters" button to do that. Updating the sequence is good because it prevents the TS measurement edit panel from displaying a mismatched parameters error.

bkeryan (approved):

dixonjoel (approved):

PR #216, opened 2023-01-20
Replacing DriverError with Error. Adding additional error code for timeout exceeded error.

DelpireNI:

1. `nidcpower.DriverError` does not exist. ~It should be `nidcpower.Error`.~ It should be `nidcpower.errors.DriverError`. (Fixed fgen too)
2. When testing with real hardware and adding a source delay, the error code I got was `-1074097933` (`kErrorTimeoutExceeded`).

Fixed the error handling and added the additional error code check.

Current example does not support source delay with hardware.

Tested on a machine with an actual SMU device.

### DIFF
--- a/examples/nidcpower_source_dc_voltage/measurement.py
+++ b/examples/nidcpower_source_dc_voltage/measurement.py
@@ -127,7 +127,7 @@ def measure(
                 try:
                     channels.wait_for_event(nidcpower.enums.Event.SOURCE_COMPLETE, timeout=0.1)
                     break
-                except nidcpower.Error as e:
+                except nidcpower.errors.DriverError as e:
                     """
                     There is no native way to support cancellation when taking a DCPower
                     measurement. To support cancellation, we will be calling WaitForEvent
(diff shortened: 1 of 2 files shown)


examples/nidcpower_source_dc_voltage/measurement.py
```diff
@@ -126,15 +127,18 @@ def cancel_callback():
                 try:
                     channels.wait_for_event(nidcpower.enums.Event.SOURCE_COMPLETE, timeout=0.1)
                     break
-                except nidcpower.DriverError as e:
+                except nidcpower.Error as e:
```
bkeryan, line 130: `DriverError` is defined here: https://github.com/ni/nimi-python/blob/master/generated/nidcpower/nidcpower/errors.py

but it's not aliased into the `nidcpower` module: https://github.com/ni/nimi-python/blob/master/generated/nidcpower/nidcpower/__init__.py

`nidcpower.Error` isn't guaranteed to have a `code` field, so change this to `nidcpower.errors.DriverError`

bkeryan (changes requested):

examples/nidcpower_source_dc_voltage/measurement.py
bkeryan: The same is true for `nifgen`, so please update the error check you added yesterday to use `nifgen.errors.DriverError`

DelpireNI: Fixed. Tested the DCPower example again and `nidcpower.errors.DriverError` worked.

bkeryan (approved):

Issue #202, opened 2023-01-11
Single-source `MeasurementInfo` and `ServiceInfo` in `.serviceconfig` file


bkeryan:
## Problem to Solve

All of the data in the `.serviceconfig` file is duplicated in the `MeasurementInfo` and `ServiceInfo` objects in `measurement.py`. If these files are not in sync, then the measurement service will work correctly when you start it manually but not when you statically register it, or vice versa.

## Proposed Solution

In the `MeasurementService` constructor, allow specifying the path to the `.serviceinfo` file instead of specifying the `measurement_info` and `service_info` parameters. When this is specified, the `MeasurementService` object parses the `.serviceinfo` and automatically constructs `MeasurementInfo` and `ServiceInfo` objects.

This may require additional parameters for `ui_file_paths` and `version` because these are not in the `.serviceconfig` file, or it might require adding these parameters to the `.serviceconfig` file.

[AB#2267348](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2267348)

dixonjoel:
@bkeryan @DelpireNI This issue should be closed now, right?

PR #219, opened 2023-01-26
Updating MeasurementService __init__ to parse .serviceconfig.

DelpireNI:

Rather than duplicating the service info, measurements should just parse their .serviceconfig file so that information can be single sourced.

Current this PR updates the `MeasurementService` `__init__` method to take in the .serviceconfig file path, version, and UI path. I have only updated 1 example so that we can discuss if this is the correct approach or not.

#202 

Once we agree on an implementation, this should be merge to prevent duplication of information between the .serviceconfig and measurement.py files.

Ran the measurement and successfully connected to it via MeasurementLink UI Editor.

### DIFF
--- a/examples/nidaqmx_analog_input/measurement.py
+++ b/examples/nidaqmx_analog_input/measurement.py
@@ -9,11 +9,11 @@ import nidaqmx
 
 import ni_measurementlink_service as nims
 
-parent_directory = pathlib.Path(__file__).resolve().parent
+service_directory = pathlib.Path(__file__).resolve().parent
 measurement_service = nims.MeasurementService(
-    service_config_path=parent_directory / "NIDAQmxAnalogInput.serviceconfig",
+    service_config_path=service_directory / "NIDAQmxAnalogInput.serviceconfig",
     version="0.1.0.0",
-    ui_file_paths=[parent_directory / "NIDAQmxAnalogInput.measui"],
+    ui_file_paths=[service_directory / "NIDAQmxAnalogInput.measui"],
 )
 
 
--- a/examples/nidaqmx_analog_input/measurement.py
+++ b/examples/nidaqmx_analog_input/measurement.py
@@ -9,19 +9,13 @@ import nidaqmx
 
 import ni_measurementlink_service as nims
 
-measurement_info = nims.MeasurementInfo(
-    display_name="NI-DAQmx Analog Input (Py)",
+parent_directory = pathlib.Path(__file__).resolve().parent
+measurement_service = nims.MeasurementService(
+    service_config_path=parent_directory / "NIDAQmxAnalogInput.serviceconfig",
     version="0.1.0.0",
-    ui_file_paths=[pathlib.Path(__file__).resolve().parent / "NIDAQmxAnalogInput.measui"],
+    ui_file_paths=[parent_directory / "NIDAQmxAnalogInput.measui"],
 )
-
-service_info = nims.ServiceInfo(
-    service_class="ni.examples.NIDAQmxAnalogInput_Python",
-    description_url="",
-)
-
-measurement_service = nims.MeasurementService(measurement_info, service_info)
-
+service_options = ServiceOptions()
 
 @measurement_service.register_measurement
 @measurement_service.configuration("physical_channel", nims.DataType.String, "Dev1/ai0")
--- a/examples/nidaqmx_analog_input/measurement.py
+++ b/examples/nidaqmx_analog_input/measurement.py
@@ -15,7 +15,7 @@ measurement_service = nims.MeasurementService(
     version="0.1.0.0",
     ui_file_paths=[parent_directory / "NIDAQmxAnalogInput.measui"],
 )
-service_options = ServiceOptions()
+
 
 @measurement_service.register_measurement
 @measurement_service.configuration("physical_channel", nims.DataType.String, "Dev1/ai0")
--- a/ni_measurementlink_service/measurement/service.py
+++ b/ni_measurementlink_service/measurement/service.py
@@ -3,7 +3,9 @@
 from __future__ import annotations
 
 from threading import Lock
-from typing import Any, Callable, Dict, TypeVar
+from typing import Any, Callable, Dict, TypeVar, List
+import json
+from os import path
 
 import grpc
 
@@ -128,18 +130,35 @@ class MeasurementService:
 
     """
 
-    def __init__(self, measurement_info: MeasurementInfo, service_info: ServiceInfo) -> None:
-        """Initialize the Measurement Service object with measurement info and service info.
+    def __init__( self, service_config_path: str, version: str, ui_file_paths: List[str]) -> None:
+        """Initialize the Measurement Service object with the .serviceconfig file, version, and UI file paths.
 
         Args:
         ----
-            measurement_info (MeasurementInfo): Measurement Info
+            service_config_path (str): Measurement Info
 
-            service_info (ServiceInfo): Service Info
+            version (str): Version of the measurement service.
+
+            ui_file_paths (List[str]): Service Info
 
         """
-        self.measurement_info: MeasurementInfo = measurement_info
-        self.service_info: ServiceInfo = service_info
+        if not path.exists(service_config_path):
+            raise Exception(f"File does not exist. {service_config_path}")
+
+        with open(service_config_path) as service_config_file:
+            service_config = json.load(service_config_file)
+
+        self.measurement_info = MeasurementInfo(
+            display_name=service_config["services"][0]["displayName"],
+            version=version,
+            ui_file_paths=ui_file_paths,
+        )
+
+        self.service_info = ServiceInfo(
+            service_class=service_config["services"][0]["serviceClass"],
+            description_url=service_config["services"][0]["descriptionUrl"],
+        )
+
         self.configuration_parameter_list: list = []
         self.output_parameter_list: list = []
         self.grpc_service = GrpcService()
@@ -326,3 +345,4 @@ class MeasurementService:
         )
 
         return self.channel_pool.get_channel(target=service_location.insecure_address)
+        
--- a/ni_measurementlink_service/measurement/service.py
+++ b/ni_measurementlink_service/measurement/service.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json
 from os import path
+from pathlib import Path
 from threading import Lock
 from typing import Any, Callable, Dict, TypeVar, List
 
--- a/ni_measurementlink_service/measurement/service.py
+++ b/ni_measurementlink_service/measurement/service.py
@@ -136,7 +136,7 @@ class MeasurementService:
         version: str,
         ui_file_paths: List[str],
         service_class: str = None,
-        ) -> None:
+    ) -> None:
         """Initialize the Measurement Service object.
 
         Uses the specified .serviceconfig file, version, and UI file paths
@@ -151,7 +151,7 @@ class MeasurementService:
             ui_file_paths (List[str]): List of paths to supported UIs.
 
             service_class (str): The service class from the .serviceconfig to use.
-            Default value is None, which will use the first service in the 
+            Default value is None, which will use the first service in the
             .serviceconfig file.
 
         """
@@ -161,11 +161,11 @@ class MeasurementService:
         with open(service_config_path) as service_config_file:
             service_config = json.load(service_config_file)
 
-        try: 
+        try:
             service = next(
-                s for s in service_config["services"] 
-                if service_class is None 
-                or s["serviceClass"] == service_class
+                s
+                for s in service_config["services"]
+                if service_class is None or s["serviceClass"] == service_class
             )
             self.measurement_info = MeasurementInfo(
                 display_name=service["displayName"],
@@ -178,7 +178,9 @@ class MeasurementService:
                 description_url=service["descriptionUrl"],
             )
         except StopIteration:
-            raise RuntimeError(f"Service class '{service_class}' not found in '{service_config_file}'")
+            raise RuntimeError(
+                f"Service class '{service_class}' not found in '{service_config_file}'"
+            )
 
         self.configuration_parameter_list: list = []
         self.output_parameter_list: list = []
--- a/ni_measurementlink_service/measurement/service.py
+++ b/ni_measurementlink_service/measurement/service.py
@@ -135,7 +135,7 @@ class MeasurementService:
         service_config_path: str,
         version: str,
         ui_file_paths: List[str],
-        service_class: str = None
+        service_class: str = None,
         ) -> None:
         """Initialize the Measurement Service object.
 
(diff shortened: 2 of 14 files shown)


ni_measurementlink_service/measurement/service.py
```diff
@@ -128,18 +130,35 @@ class MeasurementService:
...
-        self.measurement_info: MeasurementInfo = measurement_info
-        self.service_info: ServiceInfo = service_info
+        if not path.exists(service_config_path):
+            raise Exception(f"File does not exist. {service_config_path}")
+
+        with open(service_config_path) as service_config_file:
+            service_config = json.load(service_config_file)
+
+        self.measurement_info = MeasurementInfo(
+            display_name=service_config["services"][0]["displayName"],
```
DelpireNI, line 152: `.serviceconfig` files can support multiple services. So simply indexing here doesn't seem ideal. As an alternative, we could create a ServiceConfigParser class and then the user can select the particular service they want and pass that to `__init__` instead.

Alternative to that, we could have a parameter on `__init__` that takes the index we want to use. This could even have a default value of 0. This could cover the majority of cases, but still provide a way to specify a particular service. This would come with the cost of potentially parsing the json file multiple times.

ni_measurementlink_service/measurement/service.py
```diff
@@ -3,7 +3,9 @@
 from __future__ import annotations
 
 from threading import Lock
-from typing import Any, Callable, Dict, TypeVar
+from typing import Any, Callable, Dict, TypeVar, List
```
bkeryan: [O.1.4] ✔️ DO List imports in alphabetical order 💻

https://github.com/ni/python-styleguide/blob/main/docs/Coding-Conventions.md

ni_measurementlink_service/measurement/service.py
```diff
@@ -128,18 +130,35 @@ class MeasurementService:
 
     """
 
-    def __init__(self, measurement_info: MeasurementInfo, service_info: ServiceInfo) -> None:
-        """Initialize the Measurement Service object with measurement info and service info.
+    def __init__( self, service_config_path: str, version: str, ui_file_paths: List[str]) -> None:
```
bkeryan: Extra space (ni-python-styleguide also complained)
```suggestion
    def __init__(self, service_config_path: str, version: str, ui_file_paths: List[str]) -> None:
```

ni_measurementlink_service/measurement/service.py
bkeryan: Some more alternatives.

**Add a JSON pointer/path parameter** 
There's a standard way to specify a JSON sub-object. Two standards, actually: JSON Pointer ("/services/0") and JSON Path ("$.services[0]"). There are 3rd party modules for these. However, I don't think we need the extra flexibility provided by this approach.

**Continue supporting the old API.**
If we had already shipped this module, we would have to keep support for `measurement_info` and `service_info`. This could also be used as a fallback to support more advanced use cases.

**Optionally specify the service class.**
Add `service_class=None`. This is more maintainable than passing an index because you can see whether it's correct without looking at the `.serviceconfig` file and you don't need to shift all of the other service indices by one when you add/remove a service.
- If `service_class is None`, use the 1st service in the `.serviceconfig` file. This could even validate that there is only one service.
- Otherwise, find the service in the `.serviceconfig` file that matches this service class.

I think I like "Optionally specify the service class" best. I'm not too concerned with the overhead of reading/parsing the `.serviceconfig` file multiple times.

bkeryan (changes requested):

ni_measurementlink_service/measurement/service.py
dixonjoel: I think we should continue to support passing `measurement_info` directly like @bkeryan said to
> be used as a fallback to support more advanced use cases.

DelpireNI: AFAIK you cannot overload `__init__`. We could have it take a generic `args` parameter and then try to figure out whats being passed in, but that seems very error prone.

We could remove the arguments and create a couple of methods `from_measurement_info` or `from_service_config` to initialize them. That allows for someone forgetting to initialize.

If we want to support the current way of creating the object, I think the best solution would be to create a new class that creates the `measurement_info` and `service_info` objects. These can then be passed into `__init__`.

I think overall the best solution is to prevent the user from dual sourcing the information all together by doing what I am doing now with @bkeryan 's suggestion.

dixonjoel: Yeah, I guess the .serviceconfig approach seems clean. This may be unfounded, but it seems kind of limiting in a broader sense to make them pass a 'path' to an object. I'm thinking about the overall [Configuration](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-7.0) approach that ASP.NET Core takes in order to be very flexible about these kinds of configuration. Obviously that's overkill for what we need, but maybe there are elements of the hierarchy of configuration methods that we could have.

Anyway, I'm ok with going ahead with the .serviceconfig approach, but we either need to make this change quickly before 1.0 or allow both ways and add this after 1.0.

bkeryan: > AFAIK you cannot overload __init__. We could have it take a generic args parameter and then try to figure out whats being passed in, but that seems very error prone.

Correct.

My recommendation for supporting both approaches is to make the parameters optional and specify them by name. You can also declare keyword-only arguments by adding a `*` to the parameter list.

```
def __init__( 
        self, 
        *,
        measurement_info: MeasurementInfo = None, 
        service_info: ServiceInfo = None,
        service_config_path: str = "", 
        version: str = "", 
        ui_file_paths: List[str] = []
        ) -> None:

Comment (bkeryan) [reply ni_measurementlink_service/measurement/service.py:158]: > We could remove the arguments and create a couple of methods from_measurement_info or from_service_config to initialize them. That allows for someone forgetting to initialize.

Yeah, you could define a class method `MeasurementInfo.from_service_config(service_config_path)`

Comment (DelpireNI) [reply ni_measurementlink_service/measurement/service.py:158]: What happens then if you specify both the objects and the service config file path?

Comment (bkeryan) [reply ni_measurementlink_service/measurement/service.py:158]: > This may be unfounded, but it seems kind of limiting in a broader sense to make them pass a 'path' to an object. I'm thinking about the overall [Configuration](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-7.0) approach that ASP.NET Core takes in order to be very flexible about these kinds of configuration.

`*.serviceconfig` and `appsettings.json` have different roles:
- `*.serviceconfig` tells the discovery service when/how to launch the measurement service.
- `appsettings.json` tells the measurement service about itself.

In our C# services, we are storing the `.serviceconfig` filename in `appsettings.json` instead of hardcoding it in the source code. If we had an equivalent configuration file in Python, we could do that here too.

Comment (bkeryan) [reply ni_measurementlink_service/measurement/service.py:158]: > What happens then if you specify both the objects and the service config file path?

Whatever you decide.

Comment (DelpireNI) [reply ni_measurementlink_service/measurement/service.py]: IMO having both is a bit confusing and complicates the API. Additionally, I don't think this prevents breaking compatibility. In all our examples we pass the `MeasurementInfo` and `ServiceInfo` directly without named parameters. I think we'd have to keep those as positional parameters to prevent breaking changes, but I also feel like that would imply that those parameters are preferred over the other optional parameters.

If we are concerned about backwards compatibility, I can add them back, and we can try to document this as best as possible, but I still think a factory class that takes the `.serviceconfig`, version, and UIs, and returns a `MeasurementInfo` and `ServiceInfo` would be a cleaner approach and maintain compatibility.

File: ni_measurementlink_service/measurement/service.py [hunk 4]
```diff
@@ -154,16 +161,28 @@ def __init__(
...
-        self.service_info = ServiceInfo(
-            service_class=service_config["services"][0]["serviceClass"],
-            description_url=service_config["services"][0]["descriptionUrl"],
-        )
+        service_found = False
+        for service in service_config["services"]:
+            if (
+                service_class is None
+                or service["serviceClass"] == service
+            ):
```
bkeryan, line 169: `service["serviceClass"] == service` looks wrong.

Consider extracting the search into a helper method.

Consider using a more specific exception class like `RuntimeError`.

FYI, here's a way to eliminate the loop:
```
try: 
   service = next(s for s in service_config["services"] if service_class is None or service["serviceClass"] == service_class)
except StopIteration:
   raise RuntimeError(...)
```

DelpireNI: Oops. Yeah that is meant to be `service_class`.

I think your loop elimination is going to trigger the line length analyzer.

DelpireNI:
> I'm assuming this is still an RFC and you won't check it in. If you are going to, please re-set me and I'll review it again.

Yes. Once we're happy with the solution I'll update all the examples.

ni_measurementlink_service/measurement/service.py
bkeryan: > IMO having both is a bit confusing and complicates the API.

I agree. If we want to remove the old interface, we should do it before we release 1.0.

> I think we'd have to keep those as positional parameters to prevent breaking changes, but I also feel like that would imply that those parameters are preferred over the other optional parameters.

Right, making the existing parameters keyword-only would break compatibility. 

You could make the existing parameters optional, though.

```
def __init__( 
        self, 
        measurement_info: MeasurementInfo = None, 
        service_info: ServiceInfo = None,
        version: str = "", 
        ui_file_paths: List[str] = []
        *,
        service_config_path: str = "", 
        ) -> None:
```

ni_measurementlink_service/measurement/service.py
mshafer-NI: consider using filter https://docs.python.org/3/library/functions.html#filter

DelpireNI: @mshafer-NI Something like this?
```python
service = next(
    filter(
        lambda s: service_class is None or s["serviceClass"] == service_class,
        service_config["services"],
    ),
    None,
)

if not service:
    raise RuntimeError(
        f"Service class '{service_class}' not found in '{service_config_file}'"
    )

self.measurement_info = MeasurementInfo(
    display_name=service["displayName"],
    version=version,
    ui_file_paths=ui_file_paths,
)

self.service_info = ServiceInfo(
    service_class=service["serviceClass"],
    description_url=service["descriptionUrl"],
)
```

mshafer-NI: or:
```python
services = list(
    filter(
        lambda s: service_class is None or s["serviceClass"] == service_class,
        service_config["services"],
    ),
)

if not services:
    raise RuntimeError(
        f"Service class '{service_class}' not found in '{service_config_file}'"
    )

service = services[0]
```

mshafer-NI: or, since service_class isn't changing:
```python
if service_class is None:
   service = next(service in service_config["services"])
else:
   service = next((s for s in service_config["services"] if s["serviceClass"] == service_class), None)
   if not service :
       pass # TODO: raise

self.measurement_info = MeasurementInfo(
    display_name=service["displayName"],
    version=version,
    ui_file_paths=ui_file_paths,
)

self.service_info = ServiceInfo(
    service_class=service["serviceClass"],
    description_url=service["descriptionUrl"],
)
```

bkeryan:
```
if service_class is None:
   service = next(service_config["services"], None)
else:
   service = next((s for s in service_config["services"] if s["serviceClass"] == service_class), None)
if not service:
   raise RuntimeError(...)
```

ni_measurementlink_service/measurement/service.py
DelpireNI: Yeah that would work. I'm not convinced of the usefulness of letting people dual source their measurement/service info, nor why this approach would be better than serviceconfig parsing class.

bkeryan: I agree. Better to single source it now.

bkeryan (approved):
I approve of the idea in this RFC but please don't merge it into main :)

examples/nidaqmx_analog_input/measurement.py
```diff
@@ -9,19 +9,13 @@
 
 import ni_measurementlink_service as nims
 
-measurement_info = nims.MeasurementInfo(
-    display_name="NI-DAQmx Analog Input (Py)",
+parent_directory = pathlib.Path(__file__).resolve().parent
```
dixonjoel, line 12: I think I'd prefer a name like 'root_directory' or 'service_directory'. What do you think?

bkeryan:
Also remove [RFC] from the PR title

DelpireNI:
> Also remove [RFC] from the PR title

I was waiting until I had successfully PR checks, which it looks like I do now. :) Removed.

tests/unit/test_service.py
```diff
@@ -248,4 +240,9 @@ def _fake_measurement_function():
 @pytest.fixture
 def measurement_service() -> MeasurementService:
     """Create a MeasurementService."""
-    return MeasurementService(_TEST_MEASUREMENT_INFO, _TEST_SERVICE_INFO)
+    assets_directory = pathlib.Path(__file__).resolve().parent.parent
```
bkeryan, line 243: This isn't the assets directory. You need to add "assets" on the end.

DelpireNI:
> > Also remove [RFC] from the PR title
> 
> I was waiting until I had successfully PR checks, which it looks like I do now. :) Removed.

tests/integration/test_service_manager.py
```diff
@@ -48,8 +48,8 @@ def test___grpc_service_without_discovery_service___start_service___service_host
 def test___grpc_service_started___stop_service___service_stopped(grpc_service: GrpcService):
     """Test to validate if measurement service is stopped."""
     port_number = grpc_service.start(
-        measurement.measurement_info,
-        measurement.service_info,
+        measurement.sample_measurement_service.measurement_info,
+        measurement.sample_measurement_service.service_info,
```
bkeryan, line 52: ok, these were already public

tests/unit/test_service.py
DelpireNI: Moved `"assets"` from the line below to here to make the variable name make sense.

tests/integration/test_service_manager.py
DelpireNI: Should they be?

dixonjoel (approved):

tests/integration/test_service_manager.py
bkeryan: I guess not. The only references to those fields seem to be in this class and its tests. The same is true for `configuration_parameter_list`, `output_parameter_list`, and `grpc_service`.

What about types? `MeasurementInfo` and `ServiceInfo` are used in `DiscoveryService`, which is in an `_internal` module but is used by the example code.

Consider cleaning this up in a separate PR. Don't add it to this one.

bkeryan (approved):

tests/integration/test_service_manager.py
DelpireNI: @bkeryan We're actually documenting these types:
https://ni.github.io/measurementlink-python/autoapi/ni_measurementlink_service/index.html#ni_measurementlink_service.MeasurementInfo

bkeryan: I know. We're also documenting these fields: https://ni.github.io/measurementlink-python/autoapi/ni_measurementlink_service/index.html#ni_measurementlink_service.MeasurementService

DelpireNI: Looks like they're being generated automatically. So I guess I don't need to take any action. If we do hide them, I imagine that'll be reflected in the doc generation as well.

Issue #220, opened 2023-02-02
The examples should use a different timeout when reserving sessions in the session manager


pbirkhol-ni:
# Bug Report

Currently, the examples use a timeout of -1 when reserving sessions in the session manager (`reserve_sessions` and `reserve_all_registered_sessions`). In the LabVIEW repo we are using a timeout of 0 when reserving sessions in ProcessSetup or ProcessCleanup and a timeout of 1 minute when reserving sessions in the measurement. We should do the same in python.

[AB#2288787](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2288787)

PR #222, opened 2023-02-02
Change timeouts to 60000 in measurements and 0 in TestStand fixtures

dixonjoel:
Signed-off-by: Joel Dixon <joel.dixon@ni.com>

Change timeouts for reserving sessions to 60 seconds in the measurement and to 0 in TestStand usage.

#220 
Matches the LabVIEW repo examples.

No manual testing.

### DIFF
--- a/examples/nidcpower_source_dc_voltage/measurement.py
+++ b/examples/nidcpower_source_dc_voltage/measurement.py
@@ -72,7 +72,7 @@ def measure(
                 context=measurement_service.context.pin_map_context,
                 pin_or_relay_names=pin_names,
                 instrument_type_id=nims.session_management.INSTRUMENT_TYPE_NI_DCPOWER,
-                timeout=-1,
+                timeout=60000,
             )
         )
 
--- a/examples/nidcpower_source_dc_voltage/measurement.py
+++ b/examples/nidcpower_source_dc_voltage/measurement.py
@@ -15,7 +15,6 @@ from _helpers import ServiceOptions
 
 import ni_measurementlink_service as nims
 
-
 NIDCPOWER_WAIT_FOR_EVENT_TIMEOUT_ERROR_CODE = -1074116059
 NIDCPOWER_TIMEOUT_EXCEEDED_ERROR_CODE = -1074097933
 
@@ -72,6 +71,9 @@ def measure(
                 context=measurement_service.context.pin_map_context,
                 pin_or_relay_names=pin_names,
                 instrument_type_id=nims.session_management.INSTRUMENT_TYPE_NI_DCPOWER,
+                # If another measurement is using the session, wait for it to complete.
+                # Specify a timeout to aid in debugging missed unreserve calls.
+                # Long measurements may require a longer timeout.
                 timeout=60,
             )
         )
--- a/examples/nidcpower_source_dc_voltage/measurement.py
+++ b/examples/nidcpower_source_dc_voltage/measurement.py
@@ -72,7 +72,7 @@ def measure(
                 context=measurement_service.context.pin_map_context,
                 pin_or_relay_names=pin_names,
                 instrument_type_id=nims.session_management.INSTRUMENT_TYPE_NI_DCPOWER,
-                timeout=60000,
+                timeout=60,
             )
         )
 
--- a/examples/nivisa_dmm_measurement/teststand_fixture.py
+++ b/examples/nivisa_dmm_measurement/teststand_fixture.py
@@ -1,12 +1,13 @@
 """Functions to set up and tear down NI-VISA DMM sessions in NI TestStand."""
 
-from _helpers import GrpcChannelPoolHelper
-from _helpers import PinMapClient
-from _visa_helpers import INSTRUMENT_TYPE_DMM_SIMULATOR
-from _visa_helpers import create_visa_resource_manager
-from _visa_helpers import create_visa_session
-from _visa_helpers import log_instrument_id
-from _visa_helpers import reset_instrument
+from _helpers import GrpcChannelPoolHelper, PinMapClient
+from _visa_helpers import (
+    INSTRUMENT_TYPE_DMM_SIMULATOR,
+    create_visa_resource_manager,
+    create_visa_session,
+    log_instrument_id,
+    reset_instrument,
+)
 
 import ni_measurementlink_service as nims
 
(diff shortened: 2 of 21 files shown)


examples/nidcpower_source_dc_voltage/measurement.py
```diff
@@ -72,7 +72,7 @@ def measure(
                 context=measurement_service.context.pin_map_context,
                 pin_or_relay_names=pin_names,
                 instrument_type_id=nims.session_management.INSTRUMENT_TYPE_NI_DCPOWER,
-                timeout=-1,
+                timeout=60000,
```
bkeryan, line 75: In Python, timeouts are specified in seconds, so use 60.

Please test one of the examples in both InstrumentStudio and TestStand, including the timeout case.

bkeryan (changes requested):

examples/nidcpower_source_dc_voltage/measurement.py
bkeryan: FYI: the type annotation says `Optional[float]` and `60` is an `int`, but that's ok: https://mypy.readthedocs.io/en/latest/duck_type_compatibility.html

examples/nivisa_dmm_measurement/teststand_fixture.py
```diff
@@ -1,13 +1,12 @@
...
 
-from _helpers import GrpcChannelPoolHelper, PinMapClient
-from _visa_helpers import (
-    INSTRUMENT_TYPE_DMM_SIMULATOR,
-    create_visa_resource_manager,
-    create_visa_session,
-    log_instrument_id,
-    reset_instrument,
-)
+from _helpers import GrpcChannelPoolHelper
```
pbirkhol-ni, line 3: Can you revert this change? [PEP8 ](https://peps.python.org/pep-0008/#imports)has no problem with importing multiple objects from the same file. I think this will also fix the PR build.

pbirkhol-ni (approved):

examples/nivisa_dmm_measurement/teststand_fixture.py
pbirkhol-ni: I meant to highlight lines 3-9.

mshafer-NI: @dixonjoel, you may need to update the version of ni-python-styleguide in the lock file / your venv (0.4.0 keeps these together now)

bkeryan:
Let's also add comments explaining the timeouts. Something like this:
- teststand_fixture.py: This code module sets up the sessions, so error immediately if they are in use.
- measurement.py: If another measurement is using the session, wait for it to complete. Specify a timeout to aid in debugging missed unreserve calls. Long measurements may require a longer timeout.

examples/nivisa_dmm_measurement/teststand_fixture.py
dixonjoel: I updated the pyproject.toml to have >=0.4.0 for ni-python-styleguide. This required me to run poetry lock, so please check the lock file and pyproject file.

I have Poetry (version 1.3.1). I can't remember the correct version, but I remember @mshafer-NI saying I had to have a certain version to get the .lock file right.

dixonjoel: Also, I'm getting a bunch of D417 errors for docstrings. I'm not sure why since the methods mostly have documentation on them. Unless I need to document the 'self' argument or the docs are formatted incorrectly.
> ./examples/niswitch_control_relays/teststand_fixture.py:[10](https://github.com/ni/measurementlink-python/actions/runs/4098613934/jobs/7067851096#step:9:11):1: D417 Missing argument descriptions in the docstring
./examples/niswitch_control_relays/_helpers.py:52:1: D417 Missing argument descriptions in the docstring
./examples/niswitch_control_relays/_helpers.py:100:1: D417 Missing argument descriptions in the docstring
./examples/nivisa_dmm_measurement/teststand_fixture.py:19:1: D417 Missing argument descriptions in the docstring

examples/nidcpower_source_dc_voltage/measurement.py
dixonjoel: Any good ideas about how to test the timeout case? My thought was to build in a 60 second delay in the session manager service in my local build before it checks the timeout so it would always time out.

I figured if I just killed or made the session manager service missing, it would error out before it had a chance to timeout during the reserve session call.

examples/nivisa_dmm_measurement/teststand_fixture.py
mshafer-NI: maybe related?
https://github.com/PyCQA/pydocstyle/issues/443

> I have Poetry (version 1.3.1). I can't remember the correct version, but I remember @mshafer-NI saying I had to have a certain version to get the .lock file right.

@bkeryan this project is using Poety 1.2, right?

examples/nidcpower_source_dc_voltage/measurement.py
bkeryan: Run another measurement service that uses DCPower. Add a delay to that service or attach a debugger to it.

examples/nivisa_dmm_measurement/teststand_fixture.py
bkeryan: I think the D417 errors occur because we're mixing numpy and Google style.

numpy uses dashes with no colon: https://numpydoc.readthedocs.io/en/latest/format.html
```
Parameters
----------
x : type
    Description of parameter `x`.
y
    Description of parameter `y` (with type not specified).
```

Google style uses a colon with no dashes: https://google.github.io/styleguide/pyguide.html
```
    """Connects to the next available port.

    Args:
      minimum: A port value greater or equal to 1024.

    Returns:
      The new minimum port.

    Raises:
      ConnectionError: If no available port is found.
    """
```

bkeryan: I think you should split the ni-python-styleguide update into a separate PR.

dixonjoel: I'm fine with that. That just means these lines will have to be separate imports until we migrate the style guide.

bkeryan: You don't have to revert the imports. `ni-python-styleguide lint` never required these imports to be on separate lines.

examples/nidcpower_source_dc_voltage/measurement.py
dixonjoel: Tested with delay in InstrumentStudio and TestStand. InstrumentStudio - errors out after the specified delay (60 seconds). TestStand - errors out immediately.

bkeryan (approved):

PR #230, opened 2023-02-21
Make examples specifically depend on 1.0.0 to avoid future breakages

### LINKED ISSUE #209: Examples should depend on a specific version of ni-measurementlink-service
# Bug Report

The examples currently depend on ni-measurementlink-service version "*" (latest). There are at least two problems with this:
- If we make incompatible changes (preferably in a 2.0 version), the old examples and any services based on them will break.
- When we introduce new features, the examples in the `main` branch use the latest released package, which does not have those features until a new package is uploaded to PyPI. During that time, users may get errors like `AttributeError: module 'ni_measurementlink_service.session_management' has no attribute 'ChannelMapping'`. Specifying the required version (which is not released yet) will still result in errors in this workflow, but it might make this situation easier to debug.

I think we should also update documentation to recommend obtaining examples from the release branch.

## Repro or Code Sample

Run nidcpower example using current main branch.

## Expected Behavior

Example works.

## Current Behavior

`AttributeError: module 'ni_measurementlink_service.session_management' has no attribute 'ChannelMapping'`

## Workaround

Uncomment the ni-measurementlink-service dev-dependency in the example's pyproject.toml file.

## Context

N/A

## Your Environment

N/A

[AB#2270783](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2270783)

dixonjoel:

Make the examples depend on a specific version of ni-measurement-service, 1.0.0 as of now instead of "*" which would always take the latest.
fixes #209 

We know the examples won't break as we add new potentially breaking code to the repo.

None.

### DIFF
--- a/examples/nidaqmx_analog_input/pyproject.toml
+++ b/examples/nidaqmx_analog_input/pyproject.toml
@@ -7,7 +7,7 @@ authors = ["National Instruments"]
 [tool.poetry.dependencies]
 python = "^3.8"
 nidaqmx = ">=0.6.3"
-ni-measurementlink-service = "*"
+ni-measurementlink-service = "1.0.0"
 click = ">=7.1.2"
 
 [tool.poetry.dev-dependencies]
--- a/examples/nidaqmx_analog_input/pyproject.toml
+++ b/examples/nidaqmx_analog_input/pyproject.toml
@@ -7,7 +7,7 @@ authors = ["National Instruments"]
 [tool.poetry.dependencies]
 python = "^3.8"
 nidaqmx = ">=0.6.3"
-ni-measurementlink-service = "1.0.0"
+ni-measurementlink-service = "^1.0.0"
 click = ">=7.1.2"
 
 [tool.poetry.dev-dependencies]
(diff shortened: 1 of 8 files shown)


examples/nidaqmx_analog_input/pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = ["National Instruments"]
 [tool.poetry.dependencies]
 python = "^3.8"
 nidaqmx = ">=0.6.3"
-ni-measurementlink-service = "*"
+ni-measurementlink-service = "1.0.0"
```
bkeryan, line 10: Let's specify "^1.0.0" so that the examples use the latest 1.x.

Later, when we update an example to use a new feature that is not supported in 1.0.0, we should update the example to use "^1.1" or whatever it needs.

If we have to make compatibility-breaking changes, then we should update ni-measurementlink-service to 2.x to signal that there was a compatibility break. Clients that request "^1.0.0" will keep using the latest 1.x.

(This is basically semantic versioning.)

bkeryan (approved):

PR #236, opened 2023-03-14
README.md: Add link to manual and make API reference link more prominent

bkeryan:

Add link to MeasurementLink manual.

Make API reference link more prominent. 

Make documentation easier to find.

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,8 @@
 
 - [MeasurementLink™ Support for Python](#measurementlink--support-for-python)
   - [Introduction](#introduction)
-  - [Abbreviations](#abbreviations)
   - [Dependencies](#dependencies)
+  - [Documentation](#documentation)
   - [Examples](#examples)
   - [Developing Measurements: Quick Start](#developing-measurements-quick-start)
     - [Installation](#installation)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,6 +34,13 @@ MeasurementLink Support for Python (`ni-measurementlink-service`) is a Python fr
 
 ---
 
+## Documentation
+
+- [MeasurementLink Manual](https://www.ni.com/docs/en-US/bundle/measurementlink)
```
bkeryan, line 39: @nstokes-nati Is this the right terminology? And is this a stable URL that won't change?

DelpireNI (approved):

pbirkhol-ni (approved):

bkeryan:
> This might be better for another PR, but I think we should get rid of the dependencies section. I was helping Jason earlier today get python set up on his machine, and the first thing he did after installing python was to manually install each dependency in that list, which is unnecessary since poetry manages dependencies for you. Also, the list in the README is almost certainly going to go out of date the next time we add a dependency.

I agree we should reconsider the dependencies section.

I also agree it would be better for another PR :)

README.md
nstokes-nati: Yes, that's fine terminology. And yes, that's the best format for a direct link and I don't think this really merits a redirect.

nstokes-nati (approved):
Yeah, that's the best for a direct link and I don't think this really merits a redirect when we can just update the readme if we need to

PR #241, opened 2023-04-12
Update static registration readme and script

### LINKED ISSUE #235: Static registration instructions recommend relocating virtualenvs
# Bug Report

The static registration instructions (and `scripts/install_examples.py`) say to copy the entire measurement service directory, including the `.venv` subdirectory. However, a virtualenv's scripts and/or `pyvenv.cfg` contain the absolute path to the virtualenv, so the copied `.venv` directory refers to the original `.venv` directory.

https://docs.python.org/3/library/venv.html has a warning about this:

> Warning: Because scripts installed in environments should not expect the environment to be activated, their shebang lines contain the absolute paths to their environment’s interpreters. Because of this, environments are inherently non-portable, in the general case. You should always have a simple means of recreating an environment (for example, if you have a requirements file requirements.txt, you can invoke pip install -r requirements.txt using the environment’s pip to install all of the packages needed by the environment). If for any reason you need to move the environment to a new location, you should recreate it at the desired location and delete the one at the old location. If you move an environment because you moved a parent directory of it, you should recreate the environment in its new location. Otherwise, software installed into the environment may not work as expected.

## Repro or Code Sample

Statically register a measurement service following the instructions (or by running `scripts/install_examples.py`).

## Expected Behavior

The installed service contains no references to the original service path.

You can delete the original service directory and the installed copy continues to work properly.

## Current Behavior

The installed service contains references to the original service path.

[AB#2331982](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2331982)

dixonjoel:

Update instructions and script which installs examples. This is a follow-up to https://github.com/ni/measurementlink-python/commit/007f087a018e3614f0370a05f9abc58d240d320a which was mistakenly submitted directly to main.

The .venv should not be copied when copying an example or measurement to statically register it. The venv must be re-created in place to be fully valid.

Fixes #235 

Tested the script locally and ensured that the .venv created has no references to the original directory from the repo.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -48,8 +48,9 @@ The example measurements are *Poetry-based* projects. Follow the steps below to 
 
 To statically register an example measurement service with the MeasurementLink discovery service, do the following:
 
-1. Copy the example measurement's directory (do not copy the `.venv` subdirectory - it must be re-created in the new location), `.serviceconfig` file, and startup batch file to a subdirectory of `C:\ProgramData\National Instruments\MeasurementLink\Services`.
-2. Open a command prompt to the new example directory under `C:\ProgramData` and run `poetry install`.
+1. Copy the example measurement's directory (including the `.serviceconfig` file and startup batch file) to a subdirectory of `C:\ProgramData\National Instruments\MeasurementLink\Services`.
+2. NOTE: Do not copy the `.venv` subdirectory &mdash; the virtual environment must be re-created in the new location.
+3. Create a virtual environment in the new location. Open a command prompt to the new example directory under `C:\ProgramData` and run `poetry install`.
 
     ``` cmd
     cd <path_of_example_measurement>
--- a/examples/README.md
+++ b/examples/README.md
@@ -49,8 +49,9 @@ The example measurements are *Poetry-based* projects. Follow the steps below to 
 To statically register an example measurement service with the MeasurementLink discovery service, do the following:
 
 1. Copy the example measurement's directory (including the `.serviceconfig` file and startup batch file) to a subdirectory of `C:\ProgramData\National Instruments\MeasurementLink\Services`.
-2. NOTE: Do not copy the `.venv` subdirectory &mdash; the virtual environment must be re-created in the new location.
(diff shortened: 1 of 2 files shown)


DelpireNI (approved):

examples/README.md
```diff
@@ -48,8 +48,9 @@ The example measurements are *Poetry-based* projects. Follow the steps below to
 
 To statically register an example measurement service with the MeasurementLink discovery service, do the following:
 
-1. Copy the example measurement's directory (do not copy the `.venv` subdirectory - it must be re-created in the new location), `.serviceconfig` file, and startup batch file to a subdirectory of `C:\ProgramData\National Instruments\MeasurementLink\Services`.
-2. Open a command prompt to the new example directory under `C:\ProgramData` and run `poetry install`.
+1. Copy the example measurement's directory (including the `.serviceconfig` file and startup batch file) to a subdirectory of `C:\ProgramData\National Instruments\MeasurementLink\Services`.
+2. NOTE: Do not copy the `.venv` subdirectory &mdash; the virtual environment must be re-created in the new location.
```
bkeryan, line 52: This note is part of step 1 and should not be a separate numbered step.

Here's how to define a note in GitHub-flavored markdown: https://github.com/orgs/community/discussions/16925

Remove the spaces around the em dash.

examples/README.md
```diff
@@ -48,8 +48,9 @@ The example measurements are *Poetry-based* projects. Follow the steps below to
...
+3. Create a virtual environment in the new location. Open a command prompt to the new example directory under `C:\ProgramData` and run `poetry install`.
```
bkeryan, line 53: Clarify that "create a virtual environment" and "open a command prompt..." are the same thing:

```suggestion
3. Create a virtual environment in the new location by opening a command prompt to the new example directory under `C:\ProgramData` and running `poetry install`.
```

bkeryan (changes requested):

bkeryan (approved):

PR #245, opened 2023-04-18
examples: Create nidigital_spi example

bkeryan:

Add an NI-Digital Pattern Driver example, based on the LabVIEW `Getting Started >> SPI` example.

Implements AB#2264935

Tested with persistent simulated devices (created with nisimdev / Offline Mode).


bkeryan:
UI:
image

examples/nidigital_spi/measurement.py
```diff
@@ -0,0 +1,184 @@
...
+)
+@measurement_service.configuration(
+    "specification_file_path", nims.DataType.Path, "Specifications.specs"
+)
+@measurement_service.configuration("levels_file_path", nims.DataType.Path, "PinLevels.digilevels")
+@measurement_service.configuration("timing_file_path", nims.DataType.Path, "Timing.digitiming")
+@measurement_service.configuration("pattern_file_path", nims.DataType.Path, "Pattern.digipat")
+# TODO: MeasurementLink UI Editor doesn't support arrays of Booleans
+@measurement_service.output("passing_sites", nims.DataType.Int32Array1D)
+@measurement_service.output("failing_sites", nims.DataType.Int32Array1D)
```
bkeryan, line 39: The LabVIEW example displays an array of clusters:
image

My first inclination was to display parallel arrays of sites and pass/fail, similar to how the NI-DCPower Source DC Voltage example uses parallel arrays to return per-site information. However, displaying an array of Booleans is not supported yet.

Cc: @rfriedma

examples/nidigital_spi/measurement.py
```diff
@@ -0,0 +1,184 @@
...
+                grpc.StatusCode.INVALID_ARGUMENT,
+                f"Unsupported number of sessions: {len(reservation.session_info)}",
+            )
+
+        session_info = reservation.session_info[0]
+        session = stack.enter_context(_create_nidigital_session(session_info))
+
+        pin_map_context = measurement_service.context.pin_map_context
+        selected_sites_string = ",".join(f"site{i}" for i in pin_map_context.sites)
+        selected_sites = session.sites[selected_sites_string]
```
bkeryan, line 81: In InstrumentStudio and MeasurementLink UI Editor, selecting "all sites" works great: it makes one call with sites=[0,1,2,3].

In TestStand, the site number is based on the test socket, so you have to enable the batch process model to do multi-site, and then it runs 4 separate measurements, one site at a time.

Cc: @rfriedma @jonathanmendez

examples/nidigital_spi/teststand_fixture.py
```diff
@@ -0,0 +1,120 @@
...
+                        str(_resolve_relative_path(service_directory, path))
+                        for path in levels_file_paths
+                    ],
+                    [
+                        str(_resolve_relative_path(service_directory, path))
+                        for path in timing_file_paths
+                    ],
+                )
+                for path in pattern_file_paths:
+                    session.load_pattern(str(_resolve_relative_path(service_directory, path)))
```
bkeryan, line 77: NI-Digital returns an error if I try to load the specifications more than once without unloading, which seems like a pretty clear hint that this should only be done once per TestStand execution.

```
nidigital.errors.DriverError: -1074118560: An error occurred while loading a specifications file.

Device: DigitalPattern1
spml_kUsageTypeErrorMessage: Failed to load specifications file D:\dev\measurementlink-python\examples\nidigital_spi\Specifications.specs. Found duplicate symbol 'dc.vcc'
```

In TestStand, I'm currently specifying the filenames as file globals:
image

Cc: @jonathanmendez @rfriedma

examples/nidigital_spi/teststand_fixture.py
```diff
@@ -0,0 +1,120 @@
...
+                grpc_options = nidigital.GrpcSessionOptions(
+                    grpc_channel_pool.get_grpc_device_channel(
+                        nidigital.GRPC_SERVICE_INTERFACE_NAME
+                    ),
+                    session_name=session_info.session_name,
+                    initialization_behavior=nidigital.SessionInitializationBehavior.INITIALIZE_SERVER_SESSION,
+                )
+
+                session = nidigital.Session(
+                    resource_name=session_info.resource_name, grpc_options=grpc_options
```
bkeryan, line 58: @rfriedma: Creating a persistent simulated NI-Digital instrument requires nisimdev / Offline Mode. Should I add a constant to enable session-based simulation, like I did for the `nivisa_dmm_measurement` example? This could also make sense for the `nifgen_standard_function` example.

examples/nidigital_spi/teststand_fixture.py
```diff
@@ -0,0 +1,120 @@
...
+                )
+                # Preload the pin map, specifications, levels, timing, and pattern files.
+                session.load_pin_map(pin_map_id)
+                session.load_specifications_levels_and_timing(
+                    [
+                        str(_resolve_relative_path(service_directory, path))
```
bkeryan, line 64: The TestStand sequence is currently passing in just the filenames, but it probably ought to resolve the absolute paths using TestStand's search path like we're doing for the pin map path.

In the non-TestStand case, I still want to be able to resolve the relative path against the service directory in order to avoid saving absolute paths as default values.

examples/nidigital_spi/measurement.py
```diff
@@ -0,0 +1,184 @@
...
+
+        if not session_info.session_exists:
+            # When running the measurement from TestStand, teststand_fixture.py should have
+            # already loaded the pin map, specifications, levels, timing, and patterns.
+            session.load_pin_map(pin_map_context.pin_map_id)
```
bkeryan, line 86: This example currently assumes that `pin_map_id` is the absolute path. `teststand_fixture.py` has the same assumption: https://github.com/ni/measurementlink-python/blob/main/examples/nidmm_measurement/_helpers.py#L51

dixonjoel (approved):

pbirkhol-ni:
I am getting an error for the teststand code modules:
image
When I force upgrade pywin32 to version 306 (from 303) it worked. However, pip gave an error when I did the upgrade:
```
 ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
ni-measurementlink-service 1.0.1 requires pywin32<304,>=303; sys_platform == "win32", but you have pywin32 306 which is incompatible.
```
It looks like you submitted a PR yesterday that changes the pywin32 dependency and that should in theory fix this issue. However, we would need to make a release.

pbirkhol-ni:
I think we are planning on deleting ProcessSetup and ProcessCleanup sequences in our examples and instead putting that code in Setup/Teardown of the main sequence. You should do that here.

pbirkhol-ni:
> I am getting an error for the teststand code modules: image When I force upgrade pywin32 to version 306 (from 303) it worked. However, pip gave an error when I did the upgrade:
> 
> ```
>  ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> ni-measurementlink-service 1.0.1 requires pywin32<304,>=303; sys_platform == "win32", but you have pywin32 306 which is incompatible.
> ```
> 
> It looks like you submitted a PR yesterday that changes the pywin32 dependency and that should in theory fix this issue. However, we would need to make a release.

Brad and I discussed this offline. The issue is that I had pywin32 306 installed in my native python environment. For some reason, that was causing issues when pywin32 303 was installed in the venv.

examples/nidigital_spi/teststand_fixture.py
```diff
@@ -0,0 +1,120 @@
...
+        pin_map_client.update_pin_map(pin_map_id)
+
+
+def create_nidigital_sessions(
+    pin_map_id: str,
+    specifications_file_paths: Iterable[str],
+    levels_file_paths: Iterable[str],
+    timing_file_paths: Iterable[str],
+    pattern_file_paths: Iterable[str],
+) -> None:
```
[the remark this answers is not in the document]
bkeryan: I created separate steps for loading files into NI-Digital, updated it to use the sequence context to get the active pin map id and resolve file paths, and refactored `teststand_fixture.py`.

bkeryan:
> I think we are planning on deleting ProcessSetup and ProcessCleanup sequences in our examples and instead putting that code in Setup/Teardown of the main sequence. You should do that here.

Done

examples/nidigital_spi/teststand_fixture.py
bkeryan: Fixed: `timing_file_abs_paths = [teststand_support.resolve_file_path(p) for p in timing_file_paths]` etc.

examples/nidigital_spi/measurement.py
bkeryan: Fixed: `pin_map_id = teststand_support.get_active_pin_map_id()`

examples/nidigital_spi/teststand_fixture.py
bkeryan: I haven't gotten any feedback suggesting that we do otherwise, so I'm marking this as resolved.

examples/nidigital_spi/measurement.py
bkeryan: This was an FYI, so I'm marking it resolved.

examples/nidigital_spi/measurement.py
bkeryan: This seems like the best approach for now, marking resolved.

bkeryan:
Updated sequence: 
image

dixonjoel (approved):

examples/nidigital_spi/teststand_fixture.py
bkeryan: @rfriedma and I discussed this offline. I'm going to do this in a separate PR.

PR #251, opened 2023-04-27
RFC: service: Allow services to return incremental updates using a generator

bkeryan:

Currently, measurement services send outputs to the client when the measurement is complete, by returning a list/tuple of output values. This PR allows measurement services to send outputs to the client during the measurement, by yielding a list/tuple of output values.

When accessed via the `ni.measurementlink.measurement.v1.MeasurementService` interface, only the last update is sent to the client.

The final update may use either a `yield` or `return` statement, so that adding a `yield` statement to a previously working measurement service doesn't cause the return value to be ignored. This is why this code catches `StopIteration`: to inspect the return value stored in the exception's `value` field.

AB#2371309

Manually tested `long_running_measurement` with @jasonmreding's ASW dev branch.


dixonjoel:
This seems like a pretty good architecture. The measurement.py looks a little more complicated with the user of the Generator. Is this pretty common Python knowledge? Any concern that the measurement author won't get it?

The other big concern I have is backwards compatibility. Does this mean all existing measurement services will need to update to return Generators? Or did I miss some conversion / adapter code that would mean they don't have to change anything?

bkeryan:
> This seems like a pretty good architecture. The measurement.py looks a little more complicated with the user of the Generator. Is this pretty common Python knowledge? Any concern that the measurement author won't get it?

The only complication is in the type annotations. If you don't specify type annotation, the only change is to replace `return (x, y, z)` with `yield (x, y, z)`.

> The other big concern I have is backwards compatibility. Does this mean all existing measurement services will need to update to return Generators? Or did I miss some conversion / adapter code that would mean they don't have to change anything?

In grpc_servicer.py, checking `isinstance(return_value, collections.abc.Generator)` makes it backwards compatible. If the measurement function uses `return (x, y, z)` or `return [x, y, z]` and never calls `yield`, then `return_value` is a tuple/list and the else-case handles it appropriately.

dixonjoel (approved):

bkeryan:
This is superseded by:
- https://github.com/ni/measurementlink-python/pull/261
- https://github.com/ni/measurementlink-python/pull/262

PR #252, opened 2023-04-27
[RFC] Added pin map and session management helper functions to the python measurement template

Tharun-Sundar:

Note: This PR only contains a preliminary implementation with the intention of getting quick and iterative feedback. The implementation will be updated and cleaned up in the upcoming commits.

Include the helper methods used in the TestStand example in the measurement template. 

This will enable measurement authors to reuse the helpers and perform pin map manager and session manager related operations with minimal coding.
This PR is based on the feedback from this [discussion](https://teams.microsoft.com/l/message/19:36f0e963feb84c0c8122e6ddfbe7df7b@thread.tacv2/1675196998185?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=a092c504-be1c-4d21-9ef4-068a40ca6974&parentMessageId=1675196998185&teamName=MeasurementLink%20%26%20ETW%20Global%20Collaboration&channelName=Roadmap%20Input%20and%20Feature%20Ideas&createdTime=1675196998185).

- The ni-measurementlink-generator tests are passing.
- Built the ni-measurementlink-generator, installed the .whl file and created a measurement.


ni_measurementlink_generator/ni_measurementlink_generator/templates/_helpers.py
```diff
@@ -0,0 +1,313 @@
...
+            The resource id of the pin map to register as a pin map resource. By
+            convention, the pin_map_id is the .pinmap file path.
+
+    """
+    with GrpcChannelPoolHelper() as grpc_channel_pool:
+        pin_map_client = PinMapClient(grpc_channel=grpc_channel_pool.pin_map_channel)
+        pin_map_client.update_pin_map(pin_map_id)
+
+
+def create_nidcpower_sessions(pin_map_id: str) -> None:
```
dixonjoel, line 32: I can see why these methods need to be different since the session types are distinct for the drivers, but is there not any way we can share common code between all these by parameterizing things in common function calls?

dixonjoel (approved):

pbirkhol-ni:
Why is _helpers.py in the template folder but _internals.py is in the internal folder? It seems like both of these should be in the same folder, maybe called "support" (or something similar).

pbirkhol-ni:
I feel like _internal.py contains fundamental classes that should be part of the ni-measurement-service package rather than added to the generator measurement.

ni_measurementlink_generator/ni_measurementlink_generator/templates/_helpers.py
Tharun-Sundar: Yes Joel. I am working on updating the create and destroy sessions method to take in instrument type as a parameter.

ni_measurementlink_generator/ni_measurementlink_generator/templates/internal/_internal.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        pin_map_path = pathlib.Path(pin_map_id)
+        request = pin_map_service_pb2.UpdatePinMapFromXmlRequest(
+            pin_map_id=pin_map_id, pin_map_xml=pin_map_path.read_text(encoding="utf-8")
+        )
+        response: pin_map_service_pb2.PinMap = self._client.UpdatePinMapFromXml(request)
+        assert response.pin_map_id == pin_map_id
+
+
+class GrpcChannelPoolHelper(GrpcChannelPool):
```
rfriedma, line 69: How about GrpcChannelPoolManager ?

ni_measurementlink_generator/ni_measurementlink_generator/templates/internal/_internal.py
bkeryan, line 1: This isn't internal. Example measurements use ServiceOptions and str_to_enum directly. Example TestStand code modules use PinMapClient and GrpcChannelPoolHelper directly.

ni_measurementlink_generator/ni_measurementlink_generator/templates/_helpers.py
bkeryan, line 1: @sreekantham When is it appropriate to add some of this functionality to the official API?

ni_measurementlink_generator/ni_measurementlink_generator/templates/internal/_internal.py
bkeryan: We should also consider moving this functionality into GrpcChannelPool.

bkeryan (changes requested):

bkeryan (changes requested):

bkeryan:
mypy fails for ni_measurementlink_generator because it is trying to type-check the templates, but ni_measurementlink_generator doesn't depend on the packages used by the templates.

```
2s
Run poetry run mypy -p ni_measurementlink_generator
ni_measurementlink_generator/templates/internal/_internal.py:7: error: Cannot find implementation or library stub for module named "grpc"  [import]
ni_measurementlink_generator/templates/internal/_internal.py:8: error: Cannot find implementation or library stub for module named "ni_measurementlink_service"  [import]
ni_measurementlink_generator/templates/internal/_internal.py:9: error: Cannot find implementation or library stub for module named "ni_measurementlink_service._internal.discovery_client"  [import]
ni_measurementlink_generator/templates/internal/_internal.py:[1](https://github.com/ni/measurementlink-python/actions/runs/4820178497/jobs/8584315992?pr=252#step:9:1)0: error: Cannot find implementation or library stub for module named "ni_measurementlink_service._internal.stubs.ni.measurementlink.pinmap.v1"  [import]
ni_measurementlink_generator/templates/internal/_internal.py:14: error: Cannot find implementation or library stub for module named "ni_measurementlink_service.measurement.service"  [import]
ni_measurementlink_generator/templates/_helpers.py:3: error: Cannot find implementation or library stub for module named "ni_measurementlink_service"  [import]
ni_measurementlink_generator/templates/_helpers.py:3: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
ni_measurementlink_generator/templates/_helpers.py:4: error: Cannot find implementation or library stub for module named "nidcpower"  [import]
ni_measurementlink_generator/templates/_helpers.py:5: error: Cannot find implementation or library stub for module named "nidmm"  [import]
ni_measurementlink_generator/templates/_helpers.py:6: error: Cannot find implementation or library stub for module named "nifgen"  [import]
ni_measurementlink_generator/templates/_helpers.py:7: error: Cannot find implementation or library stub for module named "niscope"  [import]
ni_measurementlink_generator/templates/_helpers.py:8: error: Cannot find implementation or library stub for module named "niswitch"  [import]
Found [11](https://github.com/ni/measurementlink-python/actions/runs/4820178497/jobs/8584315992?pr=252#step:9:12) errors in 2 files (checked 7 source files)
```

Potential solutions:
1. Rename `_helpers.py` to `_helpers.py.mako`
2. Update `[tool.mypy]` in `pyproject.toml` to set `exclude = "templates/"` or something similar that works.

rfriedma:
General feedback. For validating with users we need to show what the impact is on the resulting measure.py. It's hard to see from the template.py changes. If we aren't ready for that then can we show the impact it would make on a shipping example?

ni_measurementlink_generator/ni_measurementlink_generator/templates/_helpers.py
bkeryan, line 1: One of the problems with `_helpers.py` is that it is treated as a top-level module, which leads to problems if you try to use multiple examples' `teststand_fixture.py` code modules in the same TestStand sequence.

If measurement services followed either of the standard layouts for Python packages, this problem would go away because each measurement service would be a top-level module and `_helpers.py` would be a submodule.

https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

bkeryan (changes requested):

ni_measurementlink_generator/ni_measurementlink_generator/templates/_helpers.py
bkeryan: https://github.com/ni/measurementlink-python/pull/255

bkeryan: https://github.com/ni/measurementlink-python/pull/256 seems to work better.

ni_measurementlink_generator/ni_measurementlink_generator/templates/_helpers.py
bkeryan: FYI, I refactored the NI-Digital version: https://github.com/ni/measurementlink-python/pull/245

bkeryan:
Are we done with this PR? If so, please close it.

PR #257, opened 2023-05-02
examples: Refactor command line options

bkeryan:

`_helpers.py`:
- Add `configure_logging(verbosity)`
- Add shared command line option decorators
- Specify that `--verbose` has a Python argument name of `verbosity`
- Add `get_service_options(**kwargs)` to build a `ServiceOptions` using keyword arguments 

`sample_measurement`:
- Add `_configure_logging(verbosity)`
- Specify that `--verbose` has a Python argument name of `verbosity`
- This example is used in automated tests, so it doesn't have a `_helpers.py`.

Reduce the amount of duplicated code in `measurement.py`.

Manually tested all examples.


dixonjoel (approved):

bkeryan:
Offline, @mshafer-NI pointed out two alternative approaches:
- Use an option mixin class like `niclickutils` does
- Define a wrapper function with `@functools.wraps` and `@click.option(...)` directly to the wrapper

I tried out the 2nd approach and decided not to go with it for two reasons:
1. It adds more boilerplate, especially for decorators that take arguments:
```
def use_simulation_option(default: bool) -> Callable:
    """Decorator for --use-simulation command line option."""

    def decorator(func: Callable) -> Callable:
        @click.option(
            "--use-simulation/--no-use-simulation",
            default=default,
            is_flag=True,
            help="Use simulated instruments.",
        )
        @functools.wraps(func)
        def wrapper(*args, **kwargs):
            return func(*args, **kwargs)

        return wrapper

    return decorator
```
2. It's harder to get the type annotations right, and I didn't want to spend much more time on it.

PR #258, opened 2023-05-02
examples: Use session-based simulation by default

bkeryan:

Update examples for modular instruments to use session-based simulation by default.

Make examples run "out of the box".

Enable simulation for instruments that either do not support persistent simulation in NI MAX or require TSM offline mode to enable persistent simulation.

Manually tested all examples in InstrumentStudio and TestStand.


examples/nifgen_standard_function/measurement.py
```diff
@@ -185,12 +195,15 @@ def _create_nifgen_session(
             initialization_behavior=nifgen.SessionInitializationBehavior.AUTO,
         )
 
-    return nifgen.Session(session_info.resource_name, session_info.channel_list, **session_kwargs)
+    return nifgen.Session(
+        session_info.resource_name, session_info.channel_list, options=options, **session_kwargs
```
bkeryan, line 199: FYI @Tharun-Sundar @dixonjoel This Session constructor takes different arguments than the rest.

examples/nifgen_standard_function/measurement.py
```diff
@@ -163,6 +168,11 @@ def cancel_callback():
...
     session_info: nims.session_management.SessionInformation,
 ) -> nifgen.Session:
+    options = {}
+    if service_options.use_simulation:
+        options["simulate"] = True
+        options["driver_setup"] = {"Model": "5423 (2CH)"}
+
     session_kwargs = {}
     if service_options.use_grpc_device:
         session_grpc_address = service_options.grpc_device_address
```
bkeryan, line 178: FYI @Tharun-Sundar @dixonjoel If we extract a shared helper function for this duplicated GrpcSessionOptions code, the rest of this function is a lot more manageable.

examples/nidcpower_source_dc_voltage/teststand_fixture.py
```diff
@@ -6,6 +6,9 @@
 
 import ni_measurementlink_service as nims
 
+# To use a physical NI SMU instrument, set this to False.
+USE_SIMULATION = True
```
bkeryan, line 10: It's annoying that this constant is duplicated between `measurement.py` and `teststand_fixture.py`.

`from measurement import USE_SIMULATION` might work in this PR, but it wouldn't work in the `multiple_measurements` example from https://github.com/ni/measurementlink-python/pull/256

Moving the constant to `_helpers.py` might work, but I don't want to encourage people to customize `_helpers.py`.

Moving the TestStand version to a sequence file global might make sense.

dixonjoel:
So with this setup, is it true that the user doesn't have to do anything in NI-MAX? Or do you still need to add a simulated device in NI-MAX?

bkeryan:
> So with this setup, is it true that the user doesn't have to do anything in NI-MAX? Or do you still need to add a simulated device in NI-MAX?

Yes, you no longer have to add simulated devices in NI MAX.

If you do add a simulated device in NI MAX, you have to set USE_SIMULATION=False in order to use it.

dixonjoel (approved):

examples/niswitch_control_relays/measurement.py
```diff
@@ -101,12 +111,13 @@ def _create_niswitch_session(
         )
 
     # This uses the topology configured in MAX.
-    return niswitch.Session(session_info.resource_name, **session_kwargs)
+    return niswitch.Session(session_info.resource_name, options=options, **session_kwargs)
```
bkeryan, line 114: `niswitch.Session` is different from the others. Instead of `options`, I think we need to pass `simulate=True` and `topology="2567/Independent"`.

bkeryan: https://github.com/ni/nimi-python/issues/1959

PR #262, opened 2023-05-04
service: Add support for updating outputs during the measurement

bkeryan:

Currently, measurement services send outputs to the client when the measurement is complete, by returning a list/tuple of output values. This PR allows measurement services to send outputs to the client during the measurement, by yielding a list/tuple of output values.

When accessed via the `ni.measurementlink.measurement.v1.MeasurementService` interface, only the last update is sent to the client.

The final update may use either a `yield` or `return` statement, so that adding a `yield` statement to a previously working measurement service doesn't cause the return value to be ignored. This is why this code catches `StopIteration`: to inspect the return value stored in the exception's `value` field.

Also:
- Update the `ni-measurementlink-service` version to 1.1.0a0
 
AB#2371309

- Ran `ni-python-styleguide`, `mypy`, and `pytest`.
- Manually tested `ui_progress_updates` example (with latest dev build of InstrumentStudio.
   - Used logging to test that the **STOP** button makes the measurement stop.
   - This example will be added in a separate PR.

### DIFF
--- a/ni_measurementlink_service/_internal/grpc_servicer.py
+++ b/ni_measurementlink_service/_internal/grpc_servicer.py
@@ -1,6 +1,7 @@
 """Contains Measurement Service Implementation class and method to host the service.
 """
 import collections.abc
+import contextlib
 import inspect
 import pathlib
 from contextvars import ContextVar
@@ -235,7 +236,18 @@ class MeasurementServiceServicerV1(v1_measurement_service_pb2_grpc.MeasurementSe
         )
         try:
             return_value = self.measure_function(**mapping_by_variable_name)
-            return self._serialize_response(return_value)
+            if isinstance(return_value, collections.abc.Generator):
+                with contextlib.closing(return_value) as output_iter:
+                    outputs = None
+                    try:
+                        while True:
+                            outputs = next(output_iter)
+                    except StopIteration as e:
+                        if e.value is not None:
+                            outputs = e.value
+                    return self._serialize_response(outputs)
+            else:
+                return self._serialize_response(return_value)
         finally:
             measurement_service_context.get().mark_complete()
             measurement_service_context.reset(token)
@@ -368,7 +380,17 @@ class MeasurementServiceServicerV2(v2_measurement_service_pb2_grpc.MeasurementSe
         )
         try:
             return_value = self.measure_function(**mapping_by_variable_name)
-            yield self._serialize_response(return_value)
+            if isinstance(return_value, collections.abc.Generator):
+                with contextlib.closing(return_value) as output_iter:
+                    try:
+                        while True:
+                            outputs = next(output_iter)
+                            yield self._serialize_response(outputs)
+                    except StopIteration as e:
+                        if e.value is not None:
+                            yield self._serialize_response(e.value)
+            else:
+                yield self._serialize_response(return_value)
         finally:
             measurement_service_context.get().mark_complete()
             measurement_service_context.reset(token)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ exclude = '''
 
 [tool.poetry]
 name = "ni_measurementlink_service"
-version = "1.0.1"
+version = "1.1.0a0"
 description = "MeasurementLink Support for Python"
 authors = ["NI <opensource@ni.com>"]
 readme = "README.md" # apply the repo readme to the package as well


bkeryan:
Screenshot:
image

dixonjoel:
You may want to test whether the STOP button also still works correctly to stop a measurement that is in the progress of returning incremental updates.

examples/long_running_measurement/_helpers.py
```diff
@@ -0,0 +1,233 @@
...
+        return mapping[value]
+    except KeyError as e:
+        logging.error("Unsupported enum value %s", value)
+        raise grpc.RpcError(
+            grpc.StatusCode.INVALID_ARGUMENT,
+            f'Unsupported enum value "{value}"',
+        ) from e
+
+
+class PinMapClient(object):
```
dixonjoel, line 52: I think many of these helpers are not used by this example. Is the thought just to keep them all the same in case anyone wants to tweak this example and might make use of them later?

examples/long_running_measurement/measurement.py
```diff
@@ -0,0 +1,83 @@
...
+
+        if cancellation_event.wait(10e-3):
+            logging.info("Canceling measurement")
+            measurement_service.context.abort(
+                grpc.StatusCode.CANCELLED, "Client requested cancellation"
+            )
+
+    logging.info("Completed measurement")
+    status = "Measurement complete."
+    return (random_numbers, status, overlimit, elapsed_time)
```
dixonjoel, line 69: It seems to me like this last set of outputs returned would always be the same as the last yield? That's not necessarily a problem, definitely not a big problem. But I'm curious what the rules are here. Does the measure method have to have a return at the end? Does it have to yield at least once if there isn't a return statement?

examples/long_running_measurement/poetry.toml
```diff
@@ -0,0 +1,2 @@
+[virtualenvs]
+in-project = true
```
dixonjoel, line 2: What does this file and this 'in-project' line do for us again?

examples/long_running_measurement/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+
+[tool.poetry.dependencies]
+python = "^3.8"
+ni-measurementlink-service = {version = "^1.1.0a0", allow-prereleases = true}
+click = ">=7.1.2"
+
+[tool.poetry.group.dev.dependencies]
+mypy = ">=1.0"
+# Uncomment to use prerelease dependencies (requires poetry>=1.2).
+ni-measurementlink-service = {path = "../..", develop = true}
```
dixonjoel, line 15: Is this supposed to be uncommented for now until we actually release 1.1.0a0? Or should it be commented out?

examples/long_running_measurement/start.bat
```diff
@@ -0,0 +1,6 @@
+@echo off
+REM The discovery service uses this script to start the measurement service.
+REM You can customize this script for your Python setup. The -v option logs
+REM messages with level INFO and above.
+
+call python "%~dp0measurement.py" -v
```
dixonjoel, line 6: Looks like most / all of our other examples use `.venv\Scripts\python.exe measurement.py -v`. Is there a reason this is different?

bkeryan:
> You may want to test whether the STOP button also still works correctly to stop a measurement that is in the progress of returning incremental updates.

I originally overlooked that. It did not work, so I fixed it in commit 4 and then implemented a slightly better fix in commit 5.

dixonjoel (changes requested):

examples/long_running_measurement/_helpers.py
bkeryan: `_helpers.py` is shared code, but we want each example to be a standalone program, so it is "vendored" in each example.

Maintaining N identical copies of the same file is not great, but it's better than maintaining N slightly different copies of the same file.

examples/long_running_measurement/measurement.py
bkeryan: > It seems to me like this last set of outputs returned would always be the same as the last yield?

The `status` field is different.

> But I'm curious what the rules are here. Does the measure method have to have a return at the end?

No. The `return` is optional. Normally, a generator function in Python only uses `yield`. However, it would be confusing if adding a `yield` to your measurement function caused the `return` to be ignored.

> Does it have to yield at least once if there isn't a return statement?

Yes, it must yield or return a result.

examples/long_running_measurement/poetry.toml
bkeryan: It tells Poetry to create a virtualenv in `.venv` rather than `C:\Users\username\AppData\Local\pypoetry\Cache\virtualenvs\long-running-measurement-asdfghjk-py3.9`.

examples/long_running_measurement/measurement.py
bkeryan: I added comments explaining this.

examples/long_running_measurement/measurement.py
```diff
@@ -0,0 +1,103 @@
...
+        status = f"Loop index: {loop_index}"
+        loop_index += 1
+
+        elapsed_time = time.monotonic() - start_time
+
+        desired_len = int(elapsed_time * 1000.0 / 10.0)
+        while len(random_numbers) < desired_len:
+            random_number = random.randint(1, 1000)
+            random_numbers.append(random_number)
+            if random_number > 800:
```
pbirkhol-ni, line 55: I think the limit should be an input to the measurement. At a minimum, I think there should be a comment on the UI that the limit is hardcoded to 800.

pbirkhol-ni (approved):
Does python support partial updates? If so, I think it would be useful to create an example that shows partial updates. Or modify this example to use partial updates.

bkeryan:
> Does python support partial updates? If so, I think it would be useful to create an example that shows partial updates. Or modify this example to use partial updates.

Not yet.

examples/long_running_measurement/measurement.py
```diff
@@ -0,0 +1,103 @@
...
+    version="0.5.0.0",
+    ui_file_paths=[service_directory / "LongRunningMeasurement.measui"],
+)
+
+
+Outputs = Tuple[List[int], str, List[int], float]
+
+
+@measurement_service.register_measurement
+@measurement_service.configuration("time_in_seconds", nims.DataType.Int32, 10)
```
pbirkhol-ni, line 26: Why is this an Int32? What would it mean to run this measurement for negative time?  :D   Same goes for random number and overlimit. As far as I can tell, those can only be positive integers and should probably be UInt32.

bkeryan: I would prefer to use float to represent fractional seconds, but I copied the data types from @dixonjoel's C# version.

Random numbers are not inherently positive or integral, and I would be happy to replace them with a +/-10.0 sine wave or something.

pbirkhol-ni: Yeah, that's fair. The random numbers that this particular example creates are always positive but that isn't guaranteed to always be true. But time in seconds should probably be a U32. Although if we change to a sine wave then we might want to be using periods instead, which I guess would also be positive.

pbirkhol-ni:
We should consider renaming this example. LongRunningMeasurement isn't very descriptive, especially when I can make the measurement be as long or short as I want. Perhaps we should call it ContinuousUpdateMeasurement or ContinuouslyUpdatingMeasurement or something similar.

examples/long_running_measurement/measurement.py
bkeryan: I'll post a separate PR with the revised example.

examples/long_running_measurement/measurement.py
bkeryan: I'll post a separate PR with the revised example.

PR #263, opened 2023-05-04
examples: Added common helper method to reserve and create session for DC Power Example

vigkre:

- Session creation and reservation codes are common in the `measurement.py`. Hence added separate functions for them in `_nidcpower_helpers.py`. These functions are also used by `teststand_fixture.py` for creating session.
- By doing this, we have reduced unnecessary code duplications in the examples.
- User can easily understand and add functionality to the `measurement.py` code.

- This PR is created to address the thought shared by @rfriedma 

image

- In this PR, we are updating only `nidcpower` example to showcase the idea and make the code review easier.

- Created `nidcpower` measurement and verified no changes to the existing functionality


examples/nidcpower_source_dc_voltage/_helpers.py
```diff
@@ -231,3 +232,111 @@ def use_simulation_option(default: bool) -> Callable[[F], F]:
...
+) -> Tuple[Any, List[session_management.SessionInformation]]:
+    """Create and reserve driver sessions."""
+    session_management_client = nims.session_management.Client(
+        grpc_channel=measurement_service.get_channel(
+            provided_interface=nims.session_management.GRPC_SERVICE_INTERFACE_NAME,
+            service_class=nims.session_management.GRPC_SERVICE_CLASS,
+        )
+    )
+
+    with contextlib.ExitStack() as stack:
```
bkeryan, line 252: This destroys the driver session before returning. 

FYI, I am planning to minimize the use of ExitStack: https://github.com/ni/measurementlink-python/issues/260

examples/nidcpower_source_dc_voltage/_helpers.py
```diff
@@ -231,3 +232,111 @@ def use_simulation_option(default: bool) -> Callable[[F], F]:
...
+
+def create_session(
+    session_info: session_management.SessionInformation,
+    instrument_type_module,
+    service_options: Optional[ServiceOptions] = None,
+    measurement_service: Optional[MeasurementService] = None,
+) -> Any:
+    """Create driver session based on the instrument type and reserved session."""
+    session_kwargs = {}
+    # To support teststand_fixture functions since they dont have service options
```
bkeryan, line 290: I am adding session-based simulation code to these functions.

examples/nidcpower_source_dc_voltage/_helpers.py
```diff
@@ -231,3 +232,111 @@ def use_simulation_option(default: bool) -> Callable[[F], F]:
...
+            )
+        else:
+            session_grpc_channel = measurement_service.channel_pool.get_channel(
+                target=session_grpc_address
+            )
+        session_kwargs["grpc_options"] = instrument_type_module.GrpcSessionOptions(
+            session_grpc_channel,
+            session_name=session_info.session_name,
+            initialization_behavior=instrument_type_module.SessionInitializationBehavior.AUTO,
+        )
```
bkeryan, line 320: As I pointed out in a comment on one of my other PRs, it would make sense to extract **this** into a helper method.

examples/nidcpower_source_dc_voltage/measurement.py
```diff
@@ -63,96 +64,76 @@ def measure(
     """Source and measure a DC voltage with an NI SMU."""
     logging.info("Executing measurement: pin_names=%s voltage_level=%g", pin_names, voltage_level)
 
-    session_management_client = nims.session_management.Client(
-        grpc_channel=measurement_service.get_channel(
-            provided_interface=nims.session_management.GRPC_SERVICE_INTERFACE_NAME,
-            service_class=nims.session_management.GRPC_SERVICE_CLASS,
-        )
+    [session, reserved_session_info] = create_driver_session(
```
bkeryan, line 67: Destructuring a tuple using list syntax seems weird to me.

bkeryan (changes requested):
I think this is the wrong path forward. Instead of trying to write a large, one-size-fits-all helper method, I think we will be more successful in writing smaller helper methods to reduce the amount of boilerplate while still keeping the driver-specific things separate.

examples/nidcpower_source_dc_voltage/_helpers.py
bkeryan: Also, the `session_grpc_address` variable doesn't add any value. I think it's a remnant of the code before `measurement_service.get_channel` was added.

bkeryan:
We should also consider moving USE_SIMULATION to _nidcpower_helpers.py, etc.

https://github.com/ni/measurementlink-python/pull/258#discussion_r1182059967

examples/nidcpower_source_dc_voltage/_helpers.py
vigkre: Removed ExitStack code block

examples/nidcpower_source_dc_voltage/_helpers.py
```diff
@@ -231,3 +232,111 @@ def use_simulation_option(default: bool) -> Callable[[F], F]:
...
+    session_info: session_management.SessionInformation,
+    instrument_type_module,
+    service_options: Optional[ServiceOptions] = None,
+    measurement_service: Optional[MeasurementService] = None,
+) -> Any:
+    """Create driver session based on the instrument type and reserved session."""
+    session_kwargs = {}
+    # To support teststand_fixture functions since they dont have service options
+    if service_options is None:
+        with GrpcChannelPoolHelper() as grpc_channel_pool:
```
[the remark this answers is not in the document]
vigkre: Removed

examples/nidcpower_source_dc_voltage/_helpers.py
vigkre: Included session based simulation code in helper methods.

examples/nidcpower_source_dc_voltage/_helpers.py
vigkre: Removed `session_grpc_address`. Refactored the code block to separate get channel method and setting grpc options method.

examples/nidcpower_source_dc_voltage/measurement.py
vigkre: Updated.

examples/nidcpower_source_dc_voltage/_helpers.py
```diff
@@ -231,3 +232,111 @@ def use_simulation_option(default: bool) -> Callable[[F], F]:
...
+                target=session_grpc_address
+            )
+        session_kwargs["grpc_options"] = instrument_type_module.GrpcSessionOptions(
+            session_grpc_channel,
+            session_name=session_info.session_name,
+            initialization_behavior=instrument_type_module.SessionInitializationBehavior.AUTO,
+        )
+    return instrument_type_module.Session(
+        resource_name=session_info.resource_name, **session_kwargs
+    )
```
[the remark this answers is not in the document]
vigkre: Created `_nidcpower_helpers.py` to share NI-DC Power session creation code between `measurement.py` and `testand_fixture.py`

examples/nidcpower_source_dc_voltage/_helpers.py
```diff
@@ -231,3 +232,111 @@ def use_simulation_option(default: bool) -> Callable[[F], F]:
         is_flag=True,
         help="Use simulated instruments.",
     )
+
+
+def create_driver_session(
```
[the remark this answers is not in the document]
vigkre: Created `_nidcpower_helpers.py` to share NI-DC Power session creation code between `measurement.py` and `testand_fixture.py`

dixonjoel (approved):

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
```diff
@@ -0,0 +1,104 @@
+"""nidcpower Helper classes and functions for MeasurementLink examples."""
+
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, TypeVar
+
+import click
+import grpc
+import nidcpower
+
+import ni_measurementlink_service as nims
+from ni_measurementlink_service import session_management
```
bkeryan, line 10: Delete this and use `nims.session_management`.

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+from ni_measurementlink_service.measurement.service import MeasurementService
```
bkeryan, line 11: `ni_measurementlink_service/__init__.py` aliases this so you don't have to reference submodules.
```suggestion
from ni_measurementlink_service import MeasurementService
```

examples/nidcpower_source_dc_voltage/_helpers.py
bkeryan: You removed a lot more than that.

This is what removing `session_grpc_address` looks like:

```diff
-       session_grpc_address = service_options.grpc_device_address
-
        if not session_grpc_address:
            session_grpc_channel = measurement_service.get_channel(
                provided_interface=instrument_type_module.GRPC_SERVICE_INTERFACE_NAME,
                service_class="ni.measurementlink.v1.grpcdeviceserver",
            )
        else:
            session_grpc_channel = measurement_service.channel_pool.get_channel(
-               target=session_grpc_address
+               target=service_options.grpc_device_address
            )
        session_kwargs["grpc_options"] = instrument_type_module.GrpcSessionOptions(
            session_grpc_channel,
            session_name=session_info.session_name,
            initialization_behavior=instrument_type_module.SessionInitializationBehavior.AUTO,
        )
```

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+            service_class="ni.measurementlink.v1.grpcdeviceserver",
+        )
+
+        # Leave session open.
+        session = create_session(reservation.session_info[0], session_grpc_channel)
+
+    return (session, reservation.session_info)
+
+
+def create_session(
```
bkeryan, line 53: `create_nidcpower_session` and `create_session` sound the same. How will users choose between them?

bkeryan (changes requested):

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+    measurement_service: MeasurementService,
+    pin_names: Iterable[str],
+) -> Tuple[Any, List[session_management.SessionInformation]]:
+    """Create and reserve nidcpower sessions."""
+    session_management_client = nims.session_management.Client(
+        grpc_channel=measurement_service.get_channel(
+            provided_interface=nims.session_management.GRPC_SERVICE_INTERFACE_NAME,
+            service_class=nims.session_management.GRPC_SERVICE_CLASS,
+        )
+    )
```
bkeryan, line 28: Rather than duplicating this session management client boilerplate in 7 different _driver_helpers.py files, we should directly address it.

Ideas (mutually exclusive):
- Extract get_session_management_channel()
- Extract create_session_management_client(measurement_service)
- Update the Client class to accept either a grpc.Channel **OR** a MeasurementService

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+        if len(reservation.session_info) != 1:
+            measurement_service.context.abort(
+                grpc.StatusCode.INVALID_ARGUMENT,
+                f"Unsupported number of sessions: {len(reservation.session_info)}",
+            )
+
+        session_grpc_channel = measurement_service.get_channel(
+            provided_interface=nidcpower.GRPC_SERVICE_INTERFACE_NAME,
+            service_class="ni.measurementlink.v1.grpcdeviceserver",
+        )
```
bkeryan, line 45: I think we should refactor this too.

You could write a `get_grpc_device_channel(measurement_service, driver_module)`. Usage: `get_grpc_device_channel(measurement_service, nidcpower)`. Inside, it would use `getattr(driver_module, "GRPC_SERVICE_INTERFACE_NAME")`.

bkeryan (changes requested):

examples/nidcpower_source_dc_voltage/_helpers.py
```diff
@@ -221,13 +221,3 @@ def grpc_device_options(func: F) -> F:
...
-
-
-def use_simulation_option(default: bool) -> Callable[[F], F]:
-    """Decorator for --use-simulation command line option."""
-    return click.option(
-        "--use-simulation/--no-use-simulation",
-        default=default,
-        is_flag=True,
-        help="Use simulated instruments.",
-    )
```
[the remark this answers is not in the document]
vigkre: Updated.

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+        # Leave session open.
+        session = create_session(reservation.session_info[0], session_grpc_channel)
+
+    return (session, reservation.session_info)
+
+
+def create_session(
+    session_info: session_management.SessionInformation,
+    session_grpc_channel: grpc.Channel,
+) -> Any:
```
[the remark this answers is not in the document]
vigkre: updated return type

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
vigkre: Refactored this code as suggested

examples/nidcpower_source_dc_voltage/_helpers.py
```diff
@@ -231,3 +232,20 @@ def use_simulation_option(default: bool) -> Callable[[F], F]:
...
+
+
+def get_grpc_device_channel(measurement_service: MeasurementService, driver_module) -> grpc.Channel:
+    """Returns driver specific grpc device channel."""
+    return measurement_service.get_channel(
+        provided_interface=getattr(driver_module, "GRPC_SERVICE_INTERFACE_NAME"),
+        service_class="ni.measurementlink.v1.grpcdeviceserver",
+    )
+
+def create_session_management_client(measurement_service: MeasurementService):
```
bkeryan, line 244:
```
Run poetry run ni-python-styleguide lint
./examples/nidcpower_source_dc_voltage/_helpers.py:244:1: BLK100 Black would make changes.
./examples/nidcpower_source_dc_voltage/measurement.py:85:1: BLK[10](https://github.com/ni/measurementlink-python/actions/runs/4976613516/jobs/8904847290?pr=263#step:8:11)0 Black would make changes.
./examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py:44:63: BLK100 Black would make changes.
Error: Process completed with exit code 255.
```

examples/nidcpower_source_dc_voltage/_helpers.py
```diff
@@ -231,3 +232,20 @@ def use_simulation_option(default: bool) -> Callable[[F], F]:
         is_flag=True,
         help="Use simulated instruments.",
     )
+
+
+def get_grpc_device_channel(measurement_service: MeasurementService, driver_module) -> grpc.Channel:
```
bkeryan, line 237: Type annotation: `driver_module: types.ModuleType`

examples/nidcpower_source_dc_voltage/_helpers.py
bkeryan, line 244: Type annotation: `-> nims.session_management.Client`

examples/nidcpower_source_dc_voltage/measurement.py
```diff
@@ -68,96 +66,91 @@ def measure(
...
     logging.info("Executing measurement: pin_names=%s voltage_level=%g", pin_names, voltage_level)
 
-    session_management_client = nims.session_management.Client(
-        grpc_channel=measurement_service.get_channel(
-            provided_interface=nims.session_management.GRPC_SERVICE_INTERFACE_NAME,
-            service_class=nims.session_management.GRPC_SERVICE_CLASS,
-        )
+    session_management_client = create_session_management_client(measurement_service)
+
+    reservation = reserve_session(
```
bkeryan, line 71: This leaks the reservation, making the next measurement hang forever. Use `with reserve_session(...) as reservation:` to manage the lifetime of the reservation.

examples/nidcpower_source_dc_voltage/measurement.py
```diff
@@ -68,96 +66,91 @@ def measure(
...
         )
 
-        if len(reservation.session_info) != 1:
-            measurement_service.context.abort(
-                grpc.StatusCode.INVALID_ARGUMENT,
-                f"Unsupported number of sessions: {len(reservation.session_info)}",
-            )
+    session_grpc_channel = get_grpc_device_channel(measurement_service, nidcpower)
+    
+    session = create_nidcpower_session(
```
bkeryan, line 86: This leaks the session, which may leave driver resources reserved. Use `with create_session(..) as session` to manage the lifetime of the session.

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
```diff
@@ -0,0 +1,54 @@
...
+        session_name=session_info.session_name,
+        initialization_behavior=nidcpower.SessionInitializationBehavior.AUTO,
+    )
+
+    return nidcpower.Session(
+        resource_name=session_info.resource_name, options=options, **session_kwargs
+    )
+
+
+def reserve_session(
```
bkeryan, line 38: The create session and reserve session function names should be consistent about whether they include the driver name. 

I'm inclined to remove it, since the module name already contains the driver name.

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
```diff
@@ -0,0 +1,54 @@
...
+) -> nims.session_management.Reservation:
+    """Reserve the session based on the instrument type id."""    
+    return session_management_client.reserve_sessions(
+        context=pin_map_context,
+        pin_or_relay_names=pin_names,
+        instrument_type_id=nims.session_management.INSTRUMENT_TYPE_NI_DCPOWER,
+        # If another measurement is using the session, wait for it to complete.
+        # Specify a timeout to aid in debugging missed unreserve calls.
+        # Long measurements may require a longer timeout.
+        timeout=timeout,
```
bkeryan, line 52: This comment describes `timeout=60` and does not make sense here.

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
```diff
@@ -0,0 +1,54 @@
...
+
+    return nidcpower.Session(
+        resource_name=session_info.resource_name, options=options, **session_kwargs
+    )
+
+
+def reserve_session(
+    session_management_client: nims.session_management.Client,
+    pin_map_context: nims.session_management.PinMapContext,
+    timeout: int,
```
bkeryan, line 41: `timeout` is an optional float, not a required int: https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/session_management.py#L173

(You're allowed to pass an int like 60 to a function that expects a float.)

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
```diff
@@ -0,0 +1,54 @@
...
+    pin_names: Optional[Iterable[str]] = None,
+) -> nims.session_management.Reservation:
```
bkeryan, line 43: The order of this function's parameters is shuffled compared to `nims.session_management.Client.reserve_sessions`. This is because you changed `timeout` to required. If you make `timeout` optional, then both functions can have the same parameter order except for the removal of instrument_type_id.

examples/nidcpower_source_dc_voltage/teststand_fixture.py
```diff
@@ -50,32 +52,12 @@ def create_nidcpower_sessions(sequence_context: Any) -> None:
         pin_map_id = teststand_support.get_active_pin_map_id()
 
         pin_map_context = nims.session_management.PinMapContext(pin_map_id=pin_map_id, sites=None)
-        with session_management_client.reserve_sessions(
-            context=pin_map_context,
-            instrument_type_id=nims.session_management.INSTRUMENT_TYPE_NI_DCPOWER,
-            # This code module sets up the sessions, so error immediately if they are in use.
```
bkeryan, line 56: Don't delete the comment explaining `timeout=0`.

examples/nidcpower_source_dc_voltage/measurement.py
```diff
@@ -68,96 +66,91 @@ def measure(
...
-        grpc_channel=measurement_service.get_channel(
-            provided_interface=nims.session_management.GRPC_SERVICE_INTERFACE_NAME,
-            service_class=nims.session_management.GRPC_SERVICE_CLASS,
-        )
+    session_management_client = create_session_management_client(measurement_service)
+
+    reservation = reserve_session(
+        session_management_client,
+        measurement_service.context.pin_map_context,
+        timeout=60,
```
bkeryan, line 74: Don't delete the comment explaining `timeout=60`.

examples/nidcpower_source_dc_voltage/measurement.py
```diff
@@ -68,96 +66,91 @@ def measure(
...
+            grpc.StatusCode.INVALID_ARGUMENT,
+            f"Unsupported number of sessions: {len(reservation.session_info)}",
         )
 
-        if len(reservation.session_info) != 1:
-            measurement_service.context.abort(
-                grpc.StatusCode.INVALID_ARGUMENT,
-                f"Unsupported number of sessions: {len(reservation.session_info)}",
-            )
+    session_grpc_channel = get_grpc_device_channel(measurement_service, nidcpower)
```
bkeryan, line 84: `session_grpc_channel` sounds a lot like "session management grpc channel".

`grpc_device_channel` would be clearer, but this encourages sharing the channel for multiple drivers, which works today but may not always work in the future.

`nidcpower_grpc_channel` would work. Or you could move it into the create session call below:
```
with create_nidcpower_session(
    reservation.session_info[0],
    get_grpc_device_channel(measurement_service, nidcpower)
) as session:
```

examples/nidcpower_source_dc_voltage/teststand_fixture.py
```diff
@@ -50,32 +52,12 @@ def create_nidcpower_sessions(sequence_context: Any) -> None:
...
-                )
-
-                # Leave session open
-                nidcpower.Session(
-                    resource_name=session_info.resource_name,
-                    options=options,
-                    grpc_options=grpc_options,
+                grpc_channel = grpc_channel_pool.get_grpc_device_channel(
+                    nidcpower.GRPC_SERVICE_INTERFACE_NAME
                 )
```
bkeryan, line 59: Hmm, you can't reuse `_helpers.get_grpc_device_channel` because it expects a `MeasurementService`.

What if it took a `Union[MeasurementService, ChannelPool]`? Then you could use it in both places.

bkeryan (changes requested):

examples/nidcpower_source_dc_voltage/_helpers.py
```diff
@@ -231,3 +232,111 @@ def use_simulation_option(default: bool) -> Callable[[F], F]:
...
+                timeout=60,
+                pin_names=pin_names,
+            )
+        )
+
+        if len(reservation.session_info) != 1:
+            measurement_service.context.abort(
+                grpc.StatusCode.INVALID_ARGUMENT,
+                f"Unsupported number of sessions: {len(reservation.session_info)}",
+            )
```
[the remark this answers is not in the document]
vigkre: Sure. Moved to `measurement.py` code

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+) -> Tuple[Any, List[session_management.SessionInformation]]:
+    """Create and reserve nidcpower sessions."""
+    session_management_client = nims.session_management.Client(
+        grpc_channel=measurement_service.get_channel(
+            provided_interface=nims.session_management.GRPC_SERVICE_INTERFACE_NAME,
+            service_class=nims.session_management.GRPC_SERVICE_CLASS,
+        )
+    )
+
+    with reserve_session(
```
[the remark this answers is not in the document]
vigkre: I have split the reserve and create session functions with separate with statement, hence removing the context manager complications.

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
vigkre: Removed `create_nidcpower_session` . Now only `create_session` method is available inside driver helpers.

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
vigkre: Extracted `create_session_management_client(measurement_service)` into helpers.

examples/nidcpower_source_dc_voltage/_helpers.py
vigkre: Updated.

examples/nidcpower_source_dc_voltage/_helpers.py
vigkre: Updated.

examples/nidcpower_source_dc_voltage/_helpers.py
vigkre: Updated.

examples/nidcpower_source_dc_voltage/measurement.py
vigkre: Enclosed reservation code inside `with`.

examples/nidcpower_source_dc_voltage/measurement.py
vigkre: Enclosed session creation code inside `with`

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
vigkre: Removed driver name in functions.

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
vigkre: Removed comment here and added it to caller.

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
vigkre: Updated the type annotation to `Optional[float]`

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
vigkre: Function param Order updated.

examples/nidcpower_source_dc_voltage/teststand_fixture.py
vigkre: Updated the comment for `timeout=0`

examples/nidcpower_source_dc_voltage/measurement.py
vigkre: Updated the comment for `timeout=60`

examples/nidcpower_source_dc_voltage/measurement.py
vigkre: Moved the code into create_session call as suggested

examples/nidcpower_source_dc_voltage/teststand_fixture.py
vigkre: you are right Brad. We can't reuse `_helpers.get_grpc_device_channel` in teststand code.

The reason for that, teststand code uses the `get_grpc_device_channel` from ChannelPool, and that inturn calls the method below.

 ```
def get_grpc_device_channel(self, provided_interface: str) -> grpc.Channel:
        """Return gRPC channel to specified NI gRPC Device service.

        Args:
            provided_interface (str): The gRPC Full Name of the service.

        """
        return self.get_channel(
            self._discovery_client.resolve_service(
                provided_interface=provided_interface,
                service_class="ni.measurementlink.v1.grpcdeviceserver",
            ).insecure_address
        )


```

Comment (bkeryan) [reply examples/nidcpower_source_dc_voltage/teststand_fixture.py]: The TestStand code uses ChannelPoolHelper, not ChannelPool, so let's update that to expect a driver module.

```diff
diff --git a/examples/nidcpower_source_dc_voltage/_helpers.py b/examples/nidcpower_source_dc_voltage/_helpers.py
index d0bdc94..1ef4fe7 100644
--- a/examples/nidcpower_source_dc_voltage/_helpers.py
+++ b/examples/nidcpower_source_dc_voltage/_helpers.py
@@ -109,16 +109,16 @@ class GrpcChannelPoolHelper(GrpcChannelPool):
             ).insecure_address
         )

-    def get_grpc_device_channel(self, provided_interface: str) -> grpc.Channel:
+    def get_grpc_device_channel(self, driver_module: types.ModuleType) -> grpc.Channel:
         """Return gRPC channel to specified NI gRPC Device service.

         Args:
-            provided_interface (str): The gRPC Full Name of the service.
+            driver_module: The driver module (e.g. nidcpower)

         """
         return self.get_channel(
             self._discovery_client.resolve_service(
-                provided_interface=provided_interface,
+                provided_interface=getattr(driver_module, "GRPC_SERVICE_INTERFACE_NAME"),
                 service_class="ni.measurementlink.v1.grpcdeviceserver",
             ).insecure_address
         )
diff --git a/examples/nidcpower_source_dc_voltage/teststand_fixture.py b/examples/nidcpower_source_dc_voltage/teststand_fixture.py
index 73aa3ee..aa0ad9b 100644
--- a/examples/nidcpower_source_dc_voltage/teststand_fixture.py
+++ b/examples/nidcpower_source_dc_voltage/teststand_fixture.py
@@ -83,9 +83,7 @@ def destroy_nidcpower_sessions() -> None:

             for session_info in reservation.session_info:
                 grpc_options = nidcpower.GrpcSessionOptions(
-                    grpc_channel_pool.get_grpc_device_channel(
-                        nidcpower.GRPC_SERVICE_INTERFACE_NAME
-                    ),
+                    grpc_channel_pool.get_grpc_device_channel(nidcpower),
                     session_name=session_info.session_name,
                     initialization_behavior=nidcpower.SessionInitializationBehavior.ATTACH_TO_SERVER_SESSION,
                 )
```

examples/nidcpower_source_dc_voltage/teststand_fixture.py
```diff
@@ -52,12 +52,17 @@ def create_nidcpower_sessions(sequence_context: Any) -> None:
...
+            # This code module sets up the sessions, so error immediately if they are in use.
+            timeout=0,
+        ) as reservation:
             for session_info in reservation.session_info:
-                grpc_channel = grpc_channel_pool.get_grpc_device_channel(
+                grpc_device_channel = grpc_channel_pool.get_grpc_device_channel(
                     nidcpower.GRPC_SERVICE_INTERFACE_NAME
                 )
-                create_nidcpower_session(session_info, grpc_channel)
+                create_session(session_info, grpc_device_channel)
```
bkeryan, line 65: This no longer specifies `grpc_option.initialization_behavior`.

examples/nidcpower_source_dc_voltage/teststand_fixture.py
bkeryan: Oops, teststand_fixture.py has two calls that create the session. We should update both.

examples/nidcpower_source_dc_voltage/teststand_fixture.py
```diff
@@ -52,12 +52,17 @@ def create_nidcpower_sessions(sequence_context: Any) -> None:
...
 
             session_management_client.register_sessions(reservation.session_info)
```
bkeryan, line 68: Update destroy_nidcpower_sessions too.

bkeryan (changes requested):

examples/nidcpower_source_dc_voltage/teststand_fixture.py
vigkre: Updated destroy method to use `create_session` function.

examples/nidcpower_source_dc_voltage/teststand_fixture.py
```diff
@@ -50,32 +52,17 @@ def create_nidcpower_sessions(sequence_context: Any) -> None:
...
-
-                # Leave session open
-                nidcpower.Session(
-                    resource_name=session_info.resource_name,
-                    options=options,
-                    grpc_options=grpc_options,
+                grpc_device_channel = grpc_channel_pool.get_grpc_device_channel(
+                    nidcpower.GRPC_SERVICE_INTERFACE_NAME
                 )
+                create_session(session_info, grpc_device_channel)
```
bkeryan, line 65: `initialization_behavior=nidcpower.SessionInitializationBehavior.INITIALIZE_SERVER_SESSION`

bkeryan (approved):
Approved with suggestions.

Thank you for your patience 😀

examples/nidcpower_source_dc_voltage/teststand_fixture.py
subash-suresh: updated as per suggestion

jarnoldNI:
I like the updates. This comment does not apply specifically to this example since it only uses a single instrument, but commonly measurements will use more than 1 instrument. In this case, it would be really nice to have a helper function like this:

def get_session_for_pin(sessions, pin)
        DO STUFF TO FIND WHICH SESSION IS ASSOCIATED WITH PIN
return pin_session

bkeryan:
> I like the updates. This comment does not apply specifically to this example since it only uses a single instrument, but commonly measurements will use more than 1 instrument. In this case, it would be really nice to have a helper function like this:
> 
> def get_session_for_pin(sessions, pin) DO STUFF TO FIND WHICH SESSION IS ASSOCIATED WITH PIN return pin_session

Thanks for the feedback.

The information about which session is associated with which pin/site combo is in `reservation.session_info[i].channel_mappings[j]`. I think it would make sense to provide helpers so that common use cases don't require interacting directly with the channel mappings.

How are the examples using it so far?
- The nidcpower example uses it to look up each channel's channel string, pin, and site: https://github.com/ni/measurementlink-python/blob/cc3c8fa6cf6406d8b409fef5735196f7e94c7bb8/examples/nidcpower_source_dc_voltage/measurement.py#LL156C63-L156C63
- The niscope example uses it to build a pin_to_channel dictionary, which it uses to map the trigger source from a pin to a channel: https://github.com/ni/measurementlink-python/blob/cc3c8fa6cf6406d8b409fef5735196f7e94c7bb8/examples/niscope_acquire_waveform/measurement.py#L143

bkeryan:
@jarnoldNI Something like this? (untested)

```
def get_session_and_channel_string(reservation, sessions, pin, site):
    for session_index, session_info in enumerate(reservation.session_info):
        for channel_mapping in session_info.channel_mappings:
            if pin == channel_mapping.pin_or_relay_name and site == channel_mapping.site:
                return (sessions[session_index], channel_mapping.channel)
    raise KeyError(f"Pin {pin} and site {site} not found")
```

examples/nidcpower_source_dc_voltage/_helpers.py
```diff
@@ -231,3 +236,40 @@ def use_simulation_option(default: bool) -> Callable[[F], F]:
...
+    """Return created session management client."""
+    return nims.session_management.Client(
+        grpc_channel=measurement_service.get_channel(
+            provided_interface=nims.session_management.GRPC_SERVICE_INTERFACE_NAME,
+            service_class=nims.session_management.GRPC_SERVICE_CLASS,
+        )
+    )
+
+
+def get_session_and_channel_string(pin, sessions):
```
bkeryan, line 263: The function name implies that it returns a tuple containing the session and the channel string, but it actually returns a list of sessions and no channel string.

If you want the function to return one session, it must take the site number as a parameter.

If you want the function to not take the site number, it must return more than one session. Also, rename to `get_sessions_and_channel_strings` and add a list of channel strings.

bkeryan, line 263: If sessions was an object, you would call `sessions.get_session_and_channel_string(pin)`:
- Swapping the parameters would make calling this more natural.
- There is an object containing the `SessionInfo` objects. It's called `reservation` and it was in the code snippet I posted.

bkeryan, line 263: Add type annotations.

examples/nidcpower_source_dc_voltage/_helpers.py
```diff
@@ -231,3 +236,40 @@ def use_simulation_option(default: bool) -> Callable[[F], F]:
...
+    """Returns the session information for a given pin"""
+    pin_based_sessions = []
+    for session_info in sessions:
```
bkeryan, line 266: It is very surprising for a parameter named `sessions` to contain `SessionInfo` objects and not `Session` objects.

If this doesn't return a `Session` object or an index that the caller can use to get the `Session` object, I don't think it solves Jack's problem.

bkeryan (changes requested):
Please revert `get_session_and_channel_string`, add it in a separate PR, and update an example to use it.

bkeryan (changes requested):
Please revert `get_session_and_channel_string`, add it in a separate PR, and update an example to use it.

examples/nidcpower_source_dc_voltage/_helpers.py
subash-suresh: I moved the method addition to a new PR - #276

subash-suresh: I have moved the implementation to a new PR #276 and method is now present in the reservation class as it made more sense.

subash-suresh: add type annotations , the implementation can be found in #276

examples/nidcpower_source_dc_voltage/_helpers.py
subash-suresh: This API is intended to be used before the creation of the session and the api returns only the session information.
making this return the actual session might make this more driver specific, to make it a generic implementation we are only returning the session information.
The new implementation in #276 , defines a much simpler method.

bkeryan: > making this return the actual session might make this more driver specific

No, you can write a generic function that takes a Sequence[T] and returns a T.

bkeryan (approved):

PR #265, opened 2023-05-07
examples: Update docs to redirect users to the appropriate release

bkeryan:

Update the top-level `README.md` and `examples/README.md` to redirect users to the appropriate release for each MeasurementLink version.

Also delete the screenshot of a Poetry-related error message.

We will soon update examples in the `main` branch to demonstrate features that are not available in older versions of MeasurementLink.

N/A


README.md
```diff
@@ -59,7 +59,16 @@ There are three ways to do this:
 
 ## Examples
 
-The [`examples`](https://github.com/ni/measurementlink-python/tree/main/examples/) directory contains example measurement services. See the [README.md](https://github.com/ni/measurementlink-python/tree/main/examples/README.md) file for more information.
+Example measurements are available in the `examples` directory of the `measurementlink-python` Git repository. You can download the examples by cloning the Git repository or downloading the appropriate source code zip file.
+
+> **Note**
```
bkeryan, line 64: Not sure if this should be a warning or a note

README.md
```diff
@@ -59,7 +59,16 @@ There are three ways to do this:
 
 ## Examples
 
-The [`examples`](https://github.com/ni/measurementlink-python/tree/main/examples/) directory contains example measurement services. See the [README.md](https://github.com/ni/measurementlink-python/tree/main/examples/README.md) file for more information.
+Example measurements are available in the `examples` directory of the `measurementlink-python` Git repository. You can download the examples by cloning the Git repository or downloading the appropriate source code zip file.
+
+> **Warning**
```
pbirkhol-ni, line 64: While I agree with this warning, it is pretty strange. I wonder if we could do development in a private fork (or even a separate, public branch) and then push it public when the next version of IS and ML release. That might be a terrible idea, but I also don't like the idea of our mainline containing changes that customers can't use.

examples/README.md
```diff
@@ -1,5 +1,18 @@
 
-# Example Measurements
+## Example Measurements
+
+### Downloading the Example Measurements
+
+Example measurements are available in the `examples` directory of the `measurementlink-python` Git repository. (That's this directory.) You can download the examples by cloning the Git repository or downloading the appropriate source code zip file.
```
pbirkhol-ni, line 6:
```suggestion
This directory contains example measurements. You can download the examples by cloning the Git repository or downloading the appropriate source code zip file.
```

Feel free to disagree with my suggestion. I think the `That's this directory` statement is awkward so I'm trying to get rid of it.

pbirkhol-ni (approved):

README.md
nstokes-nati: I agree that this seems best. That aside, we typically use Caution for physical injury and Warning for data loss, so in writer-owned docs this would be a Note or Info. You could just drop the label and leave the text.

bkeryan: I added a release asset containing a zip file of examples, changed these docs to directly link to the zip file, reworded the warning, and removed the warning label.

examples/README.md
bkeryan: `examples/README.md` now says "These are example measurements for MeasurementLink 2023 Q3 or later," which is correct no matter whether you are looking at it on GitHub, a local Git clone, or a downloaded copy of `measurementlink-python-examples-1.1.0-dev0.zip`.

examples/README.md
```diff
@@ -1,21 +1,19 @@
...
 
-### Downloading the Example Measurements
+These are example measurements for MeasurementLink 2023 Q3 or later.
 
-Example measurements are available in the `examples` directory of the `measurementlink-python` Git repository. (That's this directory.) You can download the examples by cloning the Git repository or downloading the appropriate source code zip file.
+If you are using a previous version of MeasurementLink, download the appropriate examples:
 
-> **Warning**
-> Example measurements in the `main` branch of the Git repository may require unreleased versions of InstrumentStudio and MeasurementLink, due to new features and `.measui` format changes.
+- MeasurementLink 2023 Q1: [measurementlink-python-examples-1.0.1.zip](https://github.com/ni/measurementlink-python/releases/download/1.0.1/measurementlink-python-examples-1.0.1.zip)
```
pbirkhol-ni, line 8: It looks like you retroactively created the example assets? I can do that for LabVIEW too.

pbirkhol-ni (approved):

PR #272, opened 2023-05-18
Use NXG/fuse controls for LabVIEW measurement UIs

pbirkhol-ni:

This PR uses NXG/fuse controls for all LabVIEW measurement UIs. It modifies the DCPower LabVIEW UI and the Sample Measurement LabVIEW UI. The following were done to the VIs:
- The background color was changed to white.
- All controls/indicators were changed to NXG controls.
- The Editor Options were changed to NXG style controls.

Note that the DCPower UI VI is a copy of the DCPower UI VI in the measurementlink-labview repo. This only change I made was disconnecting it from the library and changing the icon.

#271 

The LabVIEW PO requested that we use these controls.

I ran both examples and made sure the bindings still work.

image
image


pbirkhol-ni:
The boolean input for the Sample Measurement example is awkward. I changed it to a "Switch when Pressed" mechanical action, so it behaves correctly, but it isn't very intuitive. There are other controls that might be better, but none that is obviously better.
image

bkeryan (approved):

PR #224, opened 2023-02-08
Update pinmaps in nidcpower_source_dc_voltage example to have one pin/channel

samchris007:
### What does this Pull Request accomplish?

Updates pinmaps in `nidcpower_source_dc_voltage` example to have one pin/channel.

### Why should this Pull Request be merged?

Fixes [Bug 2275162](https://ni.visualstudio.com/DevCentral/_workitems/edit/2275162): DCPower Example does not work with single channel SMUs (GitHub Issue #215)

### What testing has been done?

Manually tested.

PR #276, opened 2023-05-25
Add helper to get session information based on pin name

subash-suresh:

Currently the reserve session call returns a list of session info and users have to define their own method to get the exact session's information to use the required session on the pins that is needed to be used. A new helper method has been introduced in this PR to allow the users to get the list of session information related to a specific pin or a set of pins.

- Added a two new API to get the sessions information based on the pins namely,
  - `get_sessions_and_channels_for_pins` which takes in a list of pin names and returns a list of (session_info, channel_list).
  - `get_session_and_channel_for_pin` which takes in a single pin name and returns one session information and the corresponding channel list
- Both of these APIs have an optional site input which filters the sessions based on the site. 

Tested the APIs manually.


ni_measurementlink_service/session_management.py
```diff
@@ -138,6 +138,20 @@ def unreserve(self):
         """Unreserve sessions."""
         self._session_manager._unreserve_sessions(self._session_info)
 
+    def get_pin_based_session_info(self, pin: str) -> List[Tuple[SessionInformation, str]]:
```
bkeryan, line 141: I suspect there should be two versions of this:
- Get a single session/channel by pin and optional site. If this matches multiple sessions, throw an exception. Return a single session/channel.
- Get multiple sessions/channels by pin and optional site. Return all of the session/channel pairs that match.

bkeryan (changes requested):

ni_measurementlink_service/session_management.py
```diff
@@ -138,6 +138,20 @@ def unreserve(self):
...
+        for session_info in self.session_info:
+            for channel_mapping in session_info.channel_mappings:
+                if pin == channel_mapping.pin_or_relay_name:
+                    pin_based_session_info.append((session_info, channel_mapping.pin_or_relay_name))
+                    continue
+
+        if len(pin_based_session_info) == 0:
+            raise KeyError(f"Pin {pin} not found")
+
+        return pin_based_session_info
```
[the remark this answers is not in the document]
subash-suresh: I believe this API is supposed to be used after reservation and before calling the `_create_nidcpower_session` api.
The create session api would accept this session information to either create a new session or attach to an existing one.

ni_measurementlink_service/session_management.py
```diff
@@ -138,6 +138,20 @@ def unreserve(self):
         """Unreserve sessions."""
         self._session_manager._unreserve_sessions(self._session_info)
 
+    def get_pin_based_session_info(self, pin: str) -> List[Tuple[SessionInformation, str]]:
+        """Returns the session and channel information based on the given pin name."""
+        pin_based_session_info = []
+        for session_info in self.session_info:
+            for channel_mapping in session_info.channel_mappings:
+                if pin == channel_mapping.pin_or_relay_name:
+                    pin_based_session_info.append((session_info, channel_mapping.pin_or_relay_name))
```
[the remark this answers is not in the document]
subash-suresh: The intention was to return the channel string, i have updated this to return the entire channel mapping object.

ni_measurementlink_service/session_management.py
subash-suresh: The reason I did not have an API which accepts the site information, was because in case of measurement link the pin context always is based on the site selection in instrument studio and usually the site selection is a single site. Only when the setting is `All Sites` multiple sites might come into picture. Even in the multiple site scenario, I assume the measurement operations are going to be repeated and not unique for each of the site. This was the primary reason i felt that the measurement author might not require to provide the site information as this logic might restrict their options in the instrument studio settings.
If we wanted to go with two helper APIs can we have something like this,
1. A method which accepts only pin named 'get_session_info_for_pin` which only accepts pin name and returns an array of session information.
2. A method which accepts both pin and site `get_session_info_for_pin_and_site` which accepts both pin and site and returns a array of session information.

bkeryan: "optional site" is orthogonal to "get single session" vs. "get multiple sessions". If you remove "optional site" from my suggestion, the rest still stands.

If "optional site" is a less important use case, then it would be more appropriate to support it via an optional parameter than via a separate function.

```
if pin == channel_mapping.pin_or_relay_name and (site is None or site == channel_mapping.site):
```

subash-suresh: I have updated the PR and taken the approach of having a single method with site as optional input

examples/nidcpower_source_dc_voltage/measurement.py
```diff
@@ -78,89 +83,97 @@ def measure(
...
-                for index, mapping in enumerate(channel_mappings):
-                    measured_values[index] = measured_values[index]._replace(
-                        in_compliance=session.channels[mapping.channel].query_in_compliance()
-                    )
-            session = None  # Don't abort after this point
-
-    _log_measured_values(channel_mappings, measured_values)
+        for pin_index, pin_name in enumerate(pin_names):
+            for session_info, channel_mapping in reservation.get_pin_based_session_info(pin_name):
+                with create_session(
```
bkeryan, line 94: Before, this example sourced and measured multiple pins & sites in parallel.

Now, it sources and measures one pin+site at a time. This makes the measurement slower for no reason. 

There are two other examples that demonstrate how to create multiple sessions. Note that they loop over session as the **innermost** loop, not the **outermost** loop:
- [nifgen_standard_function](https://github.com/ni/measurementlink-python/blob/main/examples/nifgen_standard_function/measurement.py) 
- [niswitch_control_relays](https://github.com/ni/measurementlink-python/blob/main/examples/niswitch_control_relays/measurement.py)

Multi-session examples should manage a list of sessions, with an ExitStack for cleanup. Initiate could use a separate ExitStack if that makes more sense.

examples/nidcpower_source_dc_voltage/measurement.py
```diff
@@ -78,89 +83,97 @@ def measure(
...
+                                    mapping.channel
+                                ].query_in_compliance()
+                            )
+                    session = None  # Don't abort after this point
+                _log_measured_values(channel_mappings, measured_values)
+                measurement_sites += [m.site for m in channel_mappings]
+                measurement_pin_names += [m.pin_or_relay_name for m in channel_mappings]
+                voltage_measurements += [m.voltage for m in measured_values]
+                current_measurements += [m.current for m in measured_values]
+                in_compliance += [m.in_compliance for m in measured_values]
```
bkeryan, line 169: I thought one of the reasons to write this helper function was so that clients wouldn't need to deal with channel mappings directly.

ni_measurementlink_service/session_management.py
```diff
@@ -138,6 +138,24 @@ def unreserve(self):
         """Unreserve sessions."""
         self._session_manager._unreserve_sessions(self._session_info)
 
+    def get_pin_based_session_info(
+        self, pin: str, site: Optional[str] = None
```
bkeryan, line 142: Site is an integer, not a string.

ni_measurementlink_service/session_management.py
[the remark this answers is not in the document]
bkeryan: I disagree. Using it that way makes the measurement slow for no reason. Measurements that support N homogenous sessions should manage a list of sessions, not create one session at a time.

examples/nidcpower_source_dc_voltage/NIDCPowerSourceDCVoltage.pinmap
```diff
@@ -2,17 +2,20 @@
...
 	<Instruments>
 		<NIDCPowerInstrument name="DCPower1" numberOfChannels="4">
-			<ChannelGroup name="CommonDCPowerChannelGroup" />
+			<ChannelGroup name="Group1" channels="0,2" />
+			<ChannelGroup name="Group2" channels="1,3" />
 		</NIDCPowerInstrument>
 	</Instruments>
 	<Pins>
 		<DUTPin name="Pin1" />
+		<DUTPin name="Pin2" />
```
bkeryan, line 11: This is a regression for #224. This example no longer works with single-channel SMUs.

examples/nidcpower_source_dc_voltage/measurement.py
```diff
@@ -78,89 +83,97 @@ def measure(
         # Long measurements may require a longer timeout.
         timeout=60,
     ) as reservation:
-        if len(reservation.session_info) != 1:
+        if len(pin_names) != len(voltage_levels):
             measurement_service.context.abort(
                 grpc.StatusCode.INVALID_ARGUMENT,
-                f"Unsupported number of sessions: {len(reservation.session_info)}",
+                f"The voltage level input did not match the number of pins configured",
             )
```
bkeryan, line 90: Single-session measurements like this one are the use case for the single-session helper function I have been suggesting.

NI-DCPower supports channel expansion, so I think it's reasonable for the NI-DCPower example to be a single-session measurement.

examples/nidcpower_source_dc_voltage/NIDCPowerSourceDCVoltage.measui
bkeryan, line 1: FYI, I did not review this yet.

bkeryan (changes requested):
I disagree with the direction you're taking this example and I'm not convinced that this is the best example to demonstrate this new helper function.

examples/nidcpower_source_dc_voltage/measurement.py
bkeryan: Here's what takes significant time:
- create_session:
  - In InstrumentStudio, this creates a new NI-DCPower channel expansion session. This has significant overhead, especially for simulated devices. I think it has time complexity of O(Nchannels), so repeatedly creating/destroying the same session for each channel is O(Nchannels^2) with a relatively large multiplier. You should call create_session() once per channel group, and prefer fewer channel groups except when necessary (e.g. for multi-site parallelism).
  - In TestStand, this attaches to the existing session that was created in setup, which is O(1) and fast.
- initiate:
  - This starts the source operation, which is pretty fast.
  - **IMPORTANT**: If you initiate multiple sessions or channels in a row, the SMU hardware will perform source operations **simultaneously**.
  - **ALSO IMPORTANT**: If you initiate multiple channels in a single call to a single session, I think NI-DCPower automatically dispatches to multiple threads. This applies to most of these API calls, not just initialize.
- wait_for_event:
  - This waits for the source operation to complete. Part of this time is hardware overhead, part is the `source_delay` setting. If you set `source_delay` to 5 seconds, the time from the beginning of initiate() to the end of wait_for_event() is a little longer than 5 seconds.
    - Note: simulated devices don't simulate this delay accurately.
  - If you set `source_delay` to 5 seconds, initiate() N times, and wait_for_event() N times, it will take a little longer than 5 seconds.
  - If you set `source_delay` to 5 seconds and call initiate() + wait_for_event() in a single loop, it will take N * 5 seconds.
- measure:
  - This performs a measurement, which has some hardware overhead.
- abort:
  - You don't see this call in the code, because it's handled by the context manager returned from initialize().
  - I think this is pretty fast.
- close session:
  - You don't see this call in the code because it's handled by the context manager returned from create_session.
  - In InstrumentStudio, this probably has significant overhead, like create_session.

ni_measurementlink_service/session_management.py
subash-suresh: I have updated based on suggestion

ni_measurementlink_service/session_management.py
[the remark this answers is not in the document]
subash-suresh: So is the suggestion here to return the actual sessions? This would make the API driver specific, and it would be hard for us to manage it if the pin names correspond to different driver sessions.

examples/nidcpower_source_dc_voltage/measurement.py
subash-suresh: There is some details of channel mapping being added into the session info object we are returning from our new API.- https://github.com/ni/measurementlink-python/blob/f974731efa6ea622802aa060ab481461356c0a5d/ni_measurementlink_service/session_management.py#L152
I think we could include the required details here and not return the entire channel mappings structure.

bkeryan (changes requested):

ni_measurementlink_service/session_management.py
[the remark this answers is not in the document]
bkeryan: Also, `_nifgen_helpers.py` and `_niswitch_helpers.py` could provide multi-session helper functions that create a list of sessions and yield it to the caller.

```
def create_sessions(
    session_infos: Sequence[nims.session_management.SessionInformation],
    session_grpc_channel: grpc.Channel,
    initialization_behavior: Optional[
        nidcpower.SessionInitializationBehavior
    ] = nidcpower.SessionInitializationBehavior.AUTO,
) -> Generator[List[nidcpower.Session], None, None:
    with ExitStack() as stack:
        sessions = [
            stack.enter_context(create_session(session_info, session_grpc_channel, initialization_behavior))
            for session_info in session_infos
        ]
        yield sessions
```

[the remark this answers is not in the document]
subash-suresh: okay, got it. @bkeryan 
Currently the create sessions API is just taking in one session information to return 1 session in python.
So,
1.  This should be changed or a new version of this API should be added that accepts mutliple session information and returns multiple sessions.
2.  There should be a generic set of APIs that return the session based on pin information, one for single and one for multiple sessions. And these APIs will take in sessions array, pin name and session information array as input and return the appropriate session(s) based on the inputs.
3. The create method should open the context of the session internally and yield it so that once the session is used it automatically is destroyed. But one question here is that, won't the yield statement wait until previously yielded session is used/closed. Will this be a problem?.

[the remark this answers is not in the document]
bkeryan: > Currently the create sessions API

There is no create sessions API. There is only a create session API. One is singular and the other is plural.

> This should be changed or a new version of this API should be added that accepts mutliple session information and returns multiple sessions.

I am suggesting that you add a multi-session create sessions API for drivers where it is appropriate (nifgen and niswitch, to start with).

It should be a separate API that delegates to the single-session create session API.

> The create method should open the context of the session internally and yield it so that once the session is used it automatically is destroyed. But one question here is that, won't the yield statement wait until previously yielded session is used/closed. Will this be a problem?.

The create session API (singular) should not yield. It returns a session, which is a context manager, so `with create_session(...) as session:` already works fine.

I am suggesting that the create sessions API (plural) yield a list of sessions. I am not suggesting that it yield each individual session. `with create_sessions(...) as sessions:` will execute `create_sessions` up to the `yield` statement, then bind the yielded list to the sessions variable and execute the rest of the measurement until it exits the `with` block, and then it will execute the rest of `create_sessions` after the `yield` statement.

However, on further thought, this has a different problem: if you expect `create_sessions()` to return a list of sessions, you will be sorely disappointed. You have to use it in a `with` block or iterate through it with `for` or `next()`. So maybe there is a better way than what I have proposed.

> There should be a generic set of APIs that return the session based on pin information, one for single and one for multiple sessions. And these APIs will take in sessions array, pin name and session information array as input and return the appropriate session(s) based on the inputs.

Return the appropiate session(s) **and channel(s)**. There may be multiple channels per session.

I'm not committed to having this API take a sessions array. As I have previously suggested, it could return a session **index** instead, like TSM does. Look up `SemiconductorModuleContext.cs` in ni-central to see what I mean. Note that TSM's .NET API uses method overloading, so `GetNIDCPowerSessions()` returns a different type of pin query context depending on whether you pass in one pin or multiple pins.

I would really like to see this API used to **simplify** some examples (as opposed to making them more complicated or changing them to demonstrate a more complicated use case).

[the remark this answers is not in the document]
subash-suresh: I have updated the new APIs to return the session index and the channel list.
The changes on starting the context in the create sessions method is not yet done, the context is still entered in the measurement logic.
The changes in dcpower example has been reverted and niscope and nifgen examples code has been updated to demonstrate the new APIs

examples/nidcpower_source_dc_voltage/measurement.py
subash-suresh: Added two APIs as mentioned

examples/nidcpower_source_dc_voltage/NIDCPowerSourceDCVoltage.pinmap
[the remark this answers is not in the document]
subash-suresh: reverted the dcpower examples

examples/nidcpower_source_dc_voltage/measurement.py
subash-suresh: This has been resolved and i understand the comment now

examples/nifgen_standard_function/_nifgen_helpers.py
```diff
@@ -0,0 +1,40 @@
...
+# --no-use-simulation on the command line.
+USE_SIMULATION = True
+
+
+def _create_nifgen_session(
+    session_info: nims.session_management.SessionInformation,
+    session_grpc_channel: grpc.Channel,
+    initialization_behavior: Optional[
+        nifgen.SessionInitializationBehavior
+    ] = nifgen.SessionInitializationBehavior.AUTO,
```
bkeryan, line 20: Optional parameters don't have to have a type of Optional[T]. Optional[T] means nullable (can be None) which is different than optional as in "has a default". 

You can remove the type annotation for this parameter and let mypy infer it from the default value.

Did we get this wrong in the DCPower example too?

examples/niscope_acquire_waveform/_niscope_helpers.py
```diff
@@ -0,0 +1,35 @@
+"""niscope Helper classes and functions for MeasurementLink examples."""
+
+from typing import Any, Dict, Optional
+
+import grpc
+import niscope
+
+import ni_measurementlink_service as nims
+
+# To use a physical NI FGen instrument, set this to False or specify
```
bkeryan, line 10: The correct driver name is "NI-SCOPE" with a dash and in all uppercase.

examples/nifgen_standard_function/_nifgen_helpers.py
```diff
@@ -0,0 +1,40 @@
+"""nifgen Helper classes and functions for MeasurementLink examples."""
+
+from typing import Any, Dict, Optional
+
+import grpc
+import nifgen
+
+import ni_measurementlink_service as nims
+
+# To use a physical NI FGen instrument, set this to False or specify
```
bkeryan, line 10: The correct driver name is "NI-FGEN" with a dash and in all uppercase.

examples/niscope_acquire_waveform/measurement.py
```diff
@@ -132,98 +134,77 @@ def cancel_callback():
...
-                grpc.StatusCode.INVALID_ARGUMENT,
-                f"Unsupported number of sessions: {len(reservation.session_info)}",
+        sessions = [
+            stack.enter_context(
+                _create_niscope_session(
+                    session_info, get_grpc_device_channel(measurement_service, niscope)
+                )
+            )
+            for session_info in reservation.session_info
+        ]
```
bkeryan, line 144: This example should not support multiple sessions.
- NI-SCOPE supports channel expansion and automatically synchronizes instruments across the PXI trigger bus.
- With multiple independent sessions, it is impossible to map the trigger source from a DUT pin to a scope channel because each session has different scope channels.

examples/niscope_acquire_waveform/measurement.py
```diff
@@ -132,98 +134,77 @@ def cancel_callback():
...
+                enforce_realtime=True,
             )
 
-        session_info = reservation.session_info[0]
-        channel_names = session_info.channel_list
-        pin_to_channel = {
-            mapping.pin_or_relay_name: mapping.channel for mapping in session_info.channel_mappings
-        }
-        if trigger_source in pin_to_channel:
-            trigger_source = pin_to_channel[trigger_source]
```
bkeryan, line 147: Lines 143-147 are the only part of this example that can benefit from your new function.

ni_measurementlink_service/session_management.py
```diff
@@ -138,6 +138,39 @@ def unreserve(self):
...
+            raise ValueError(f"Unsupported number of sessions: {len(self.session_info)}")
+        return session_and_channel_info[0]
+
+    def get_sessions_and_channels_for_pins(
+        self, pins: Union[str, List[str]], site: Optional[int] = None
+    ) -> List[Tuple[int, List[str]]]:
+        """Returns the session information based on the given pin names."""
+        pin_names = [pins] if isinstance(pins, str) else pins
+        session_and_channel_info = []
+        for session_index in range(len(self.session_info)):
```
bkeryan, line 157: Use enumerate()

ni_measurementlink_service/session_management.py
```diff
@@ -138,6 +138,39 @@ def unreserve(self):
...
+            channel_list = []
+            for pin in pin_names:
+                for channel_mapping in self.session_info[session_index].channel_mappings:
```
bkeryan, line 160: I don't like iterating over channel_mappings multiple times because it may be large. I think it's better to iterate over channel_mappings once and match each element against pin_names.

Also, consider converting pin_names to a set for quicker matching.

Use a list comprehension:

```
channel_list = [
    mapping.channel
    for mapping in channel_mappings
    if mapping.pin_or_relay_name in pin_names and (site is None or mapping.site == site)
]
```

ni_measurementlink_service/session_management.py
```diff
@@ -138,6 +138,39 @@ def unreserve(self):
...
+                    if pin == channel_mapping.pin_or_relay_name and (
+                        site is None or site == channel_mapping.site
+                    ):
+                        channel_list.append(channel_mapping.channel)
+                        continue
+            if len(channel_list) != 0:
+                session_and_channel_info.append((session_index, channel_list))
+
+            if len(session_and_channel_info) == 0:
+                raise KeyError(f"Pin {pin} and site {site} not found")
```
bkeryan, line 170: The 2nd if-statement is at the wrong indentation level, so it will error if the first session does not have a match.

ni_measurementlink_service/session_management.py
bkeryan, line 1: Please write some tests for these new functions. They don't interact with gRPC, so you could write unit tests. Use a mock session_manager "Client" and use a real list of session_infos.

For mocking examples, see:
- https://github.com/ni/nidaqmx-python/blob/master/tests/unit/conftest.py
- https://github.com/ni/nidaqmx-python/blob/master/tests/unit/_task_utils.py
- https://github.com/ni/nidaqmx-python/blob/master/tests/unit/test_task.py

Keep all the mock stuff in tests/unit. Dont put it in tests/conftest.py.

examples/nifgen_standard_function/_nifgen_helpers.py
bkeryan, line 1: I think you should update all of the examples for https://github.com/ni/measurementlink-python/pull/263 in a separate PR. It will go much more quickly than this PR.

bkeryan (changes requested):

ni_measurementlink_service/session_management.py
subash-suresh: Updated

ni_measurementlink_service/session_management.py
subash-suresh: Updated as per suggestions

ni_measurementlink_service/session_management.py
subash-suresh: Updated to use enumerate

examples/nifgen_standard_function/_nifgen_helpers.py
subash-suresh: Updated

examples/nifgen_standard_function/measurement.py
```diff
@@ -106,15 +108,20 @@ def cancel_callback():
         )
 
         sessions = [
-            stack.enter_context(_create_nifgen_session(session_info))
+            stack.enter_context(
+                _create_nifgen_session(
+                    session_info, get_grpc_device_channel(measurement_service, nifgen)
```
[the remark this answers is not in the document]
subash-suresh: Updated as per suggestions

examples/nifgen_standard_function/_nifgen_helpers.py
subash-suresh: I have raised a new PR on the example updates - #284

sreekantham (changes requested):
Similar to my comment in the LV PR.  I don't approve of adding this to the session manager service file with out further discussion as we talked about in the water cooler on June 12th.  The methods themselves are fine.  Can we put them in a helpers method somewhere else? CC @pbirkhol-ni

examples/nifgen_standard_function/measurement.py
```diff
@@ -123,83 +130,48 @@ def cancel_callback():
...
-                    else:
-                        raise
+            is_simulated = session.simulate
+            deadline = time.time() + measurement_service.context.time_remaining
+            stop_time = time.time() + duration
+            while True:
+                if time.time() >= stop_time:
+                    break
+                if time.time() > deadline:
+                    measurement_service.context.abort(
```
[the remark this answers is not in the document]
subash-suresh: I am planning to take the example updates in another PR to make this PR simpler. I will take this comment into consideration in that PR.

ni_measurementlink_service/session_management.py
subash-suresh: Since i moved the APIs out of the sessoin_management.py into the helpers, I believe test are no longer required.

bkeryan (approved):

subash-suresh:
@bkeryan  Can we bypass @sreekantham  due to his unavailability. I have resolved the comments given by Kiran in this PR.

dixonjoel:
> @bkeryan Can we bypass @sreekantham due to his unavailability?. I have resolved the comments given by Kiran in this PR.

@bkeryan @subash-suresh Done.

dixonjoel (approved):

dixonjoel:
@bkeryan @subash-suresh How do we get the checks to pass? Does Subash need to merge main into this branch or something?

bkeryan:
> @bkeryan @subash-suresh How do we get the checks to pass? Does Subash need to merge main into this branch or something?

Yes. Sorry, I posted a comment on Wesley's enum PR but didn't notice this one is still active.

subash-suresh:
> @bkeryan @subash-suresh How do we get the checks to pass? Does Subash need to merge main into this branch or something?

I have resolved the conflicts in this PR

PR #277, opened 2023-05-31
Add pre-commit to prevent someone committing directly to main

dixonjoel:

Adding a pre-commit hook to prevent someone mistakenly committing to main directly.

### DIFF
--- a/.github/hooks/pre-commit
+++ b/.github/hooks/pre-commit
@@ -1,0 +1,16 @@
+#!/bin/bash
+# Stops accidental commits to master/main/develop. https://gist.github.com/stefansundin/9059706
+# Install:
+# cd path/to/git/repo
+# curl -fL -o .git/hooks/pre-commit https://gist.githubusercontent.com/stefansundin/9059706/raw/pre-commit
+# chmod +x .git/hooks/pre-commit
+
+BRANCH=`git rev-parse --abbrev-ref HEAD`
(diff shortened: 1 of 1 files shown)


bkeryan (changes requested):
Git hooks belong in your local `.git/hooks` not checked into `.github/hooks`. For example, if we were using the `pre-commit` Python package, each developer would [run `pre-commit install`](https://[pre-commit](https://pre-commit.com/#3-install-the-git-hook-scripts).com/#3-install-the-git-hook-scripts) to install the Git hooks.

Does GitHub do something useful with Git hooks that I'm not aware of?

dixonjoel:
> Git hooks belong in your local `.git/hooks` directory, not checked into `.github/hooks`. For example, if we were using the `pre-commit` Python package, each developer would [run `pre-commit install`](https://pre-commit.com/#3-install-the-git-hook-scripts) to install the Git hooks.
> 
> Does GitHub do something useful with Git hooks that I'm not aware of?

No, I just misread the Teams comment as .github. So we don't need to check anything in - it's just a local helper to remind a developer. Is that right?

bkeryan:
> No, I just misread the Teams comment as .github. So we don't need to check anything in - it's just a local helper to remind a developer. Is that right?

Right. Install it locally on every machine where you have the repo checked out.

PR #253, opened 2023-04-29
Upgrade ni-python-styleguide to version 0.4.1

bkeryan:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/measurement-services-python/blob/main/CONTRIBUTING.md).

### What does this Pull Request accomplish?

Update `ni-python-styleguide` to version 0.4.1.

Fix syntax of `noqa` comments. To ignore a specific error, you must put a colon after `noqa`, or else it ignores all errors. 

Fix import order in `__init__.py`.

Remove docstrings from test case functions.

### Why should this Pull Request be merged?

Fixes #250 

Removes boilerplate docstrings from tests.

### What testing has been done?

Ran `ni-python-styleguide lint` and `pytest`.

PR #279, opened 2023-06-01
Create Predictable Streaming Measurement example and write automated tests

DelpireNI:

Created a new example called Sample Streaming Measurement (Py) that returns a predictable stream of data based on the inputs. Wrote automated tests against this new example measurement.

Here is what the measui looks like:
image

We wanted to write some automated tests against a v2 measurement that streams data. The current example that does this, UI Progress Updates, returns random data, which won't allow us to write very thorough automated tests. By adding an example that returns predictable responses, we can now have automated tests.

This PR adds new automated tests to ensure v2 measurements that stream multiple responses works in Python.


tests/acceptance/test_v2_measurement_service.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def _get_serialized_measurement_configuration_parameters(
+    name: str,
+    num_responses: int,
+    data_size: int,
+    cumulative_data: bool,
+    response_interval_in_ms: int,
+    error_on_index: int,
+) -> bytes:
```
bkeryan, line 152: Make some or all of these parameters optional so tests don't have to specify parameters they don't care about.
```suggestion
def _get_serialized_measurement_configuration_parameters(
    name: str = "test",
    num_responses: int = 10,
    data_size: int = 1,
    cumulative_data: bool = True,
    response_interval_in_ms: int = 1,
    error_on_index: int = -1,
) -> bytes:
```

tests/acceptance/test_v2_measurement_service.py
```diff
@@ -0,0 +1,187 @@
...
+    name: str,
+    num_responses: int,
+    data_size: int,
+    cumulative_data: bool,
+    response_interval_in_ms: int,
+    error_on_index: int,
+) -> any_pb2.Any:
+    serialized_parameter = _get_serialized_measurement_configuration_parameters(
+        name, num_responses, data_size, cumulative_data, response_interval_in_ms, error_on_index
+    )
```
bkeryan, line 139: You can forward function arguments with `*args` and `**kwargs`.
```suggestion
def _get_configuration_parameters(*args, **kwargs) -> any_pb2.Any:
    serialized_parameter = _get_serialized_measurement_configuration_parameters(*args, **kwargs)
```

It's possible to write type annotations using `ParamSpec`, `ParamSpecArgs`, and `ParamSpecKwargs`, but this requires Python 3.10 and is probably tricky to do in a way that supports older versions.

tests/acceptance/test_v2_measurement_service.py
```diff
@@ -0,0 +1,187 @@
...
+    response_interval_in_ms: int,
+    error_on_index: int,
+) -> bytes:
+    config_params = sample_streaming_measurement_test_pb2.SampleStreamingMeasurementParameter()
+    config_params.name = name
+    config_params.num_responses = num_responses
+    config_params.data_size = data_size
+    config_params.cumulative_data = cumulative_data
+    config_params.response_interval_in_ms = response_interval_in_ms
+    config_params.error_on_index = error_on_index
```
bkeryan, line 159: FYI, protobuf messages accept keyword arguments, but I don't think it makes much of a difference in this case.

tests/acceptance/test_v2_measurement_service.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+
+    response_iterator = stub_v2.Measure(request)
+
+    index = 0
+    with pytest.raises(grpc.RpcError):
+        for response in response_iterator:
+            index += 1
+
+    assert index == error_on_index
```
bkeryan, line 126: You could also check the index in the error message.
```suggestion
    with pytest.raises(grpc.RpcError) as exc_info:
        for response in response_iterator:
            index += 1

    assert index == error_on_index
    assert f"Errored at index: {error_on_index}\n" in exc_info.value.args[0]
```

tests/assets/sample_measurement_test.proto
```diff
@@ -14,9 +14,9 @@ option csharp_namespace = "NationalInstruments.MeasurementServices.Measurements"
 //---------------------------------------------------------------------
 
 message SampleMeasurementParameter{
```
bkeryan, line 16: another style fix:
```suggestion
message SampleMeasurementParameter {
```

examples/sample_streaming_measurement/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+
+[build-system]
+requires = ["poetry-core>=1.2.0"]
+build-backend = "poetry.core.masonry.api"
+
+[[tool.mypy.overrides]]
+module = [
+    "grpc.*",
+]
+ignore_missing_imports = true
```
bkeryan, line 25: Incorporate the changes from https://github.com/ni/measurementlink-python/pull/275 
```suggestion
[tool.poetry.group.dev.dependencies]
mypy = ">=1.0"
grpc-stubs = "^1.49.1"
# Uncomment to use prerelease dependencies (requires poetry>=1.2).
# ni-measurementlink-service = {path = "../..", develop = true}

[build-system]
requires = ["poetry-core>=1.2.0"]
build-backend = "poetry.core.masonry.api"
```

bkeryan (changes requested):

bkeryan:
> Could you post a screenshot of the UI?

Is it the same as the C# version?

dixonjoel:
> Could you post a screenshot of the UI?

I've seen the C# UI so no need unless it's different.

tests/acceptance/test_v2_measurement_service.py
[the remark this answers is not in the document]
DelpireNI: Are you suggesting I move both of these tests classes into the `integration` folder? I would say the feature we are testing is whether customers can successfully write measurements that stream data. Is the data they are sending what consumers would be receiving, and can customers return an error if something goes wrong.

I think the test looks and reads like we are testing the example I wrote, but the example was written to ensure customers can write measurements this way successfully and clients can consume that data.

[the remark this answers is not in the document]
bkeryan: No, I'm suggesting that you name the new file based on the feature name, which I thought was something like "long running measurement" or "progressive updates".

We should also consider renaming `test_measurement_service.py` but that's separate from this PR.

DelpireNI:
> > Could you post a screenshot of the UI?
> 
> @DelpireNI I've seen the C# UI so no need unless it's different.

I posted a screenshot anyway.

tests/acceptance/test_v2_measurement_service.py
DelpireNI: Seems like the build doesn't complain if we don't use type annotations in this case.

tests/acceptance/test_v2_measurement_service.py
```diff
@@ -19,21 +19,8 @@ def test___streaming_measurement_service___request_number_of_responses___receive
     num_responses: int, stub_v2: v2_measurement_service_pb2_grpc.MeasurementServiceStub
 ):
     """End to End Test validating streaming measurement returns the expected number of responses."""
```
bkeryan, line 21: Test cases in given/when/then format don't need docstrings anymore. 

Did you copy this from `test_measurement_service.py`? I thought I deleted all of the test case docstrings in #253, but apparently I missed one.

tests/acceptance/test_v2_measurement_service.py
```diff
@@ -93,62 +72,44 @@ def test___streaming_measurement_service___specify_data_size___receives_expected
...
     request = v2_measurement_service_pb2.MeasureRequest(
-        configuration_parameters=_get_configuration_parameters(
-            name, num_responses, data_size, cumulative_data, response_interval_in_ms, error_on_index
-        )
+        configuration_parameters=_get_configuration_parameters(error_on_index=error_on_index)
     )
 
     response_iterator = stub_v2.Measure(request)
 
     index = 0
```
bkeryan, line 91: FYI: Python functions have a single scope for local variables. `with` and `for` do not begin a new scope. https://realpython.com/python-scope-legb-rule/

```
with pytest.raises(grpc.RpcError, match=f"Errored at index {error_on_index}"):
    for index, response in enumerate(response_iterator):
        pass

assert index == error_on_index
```

tests/acceptance/test_v2_measurement_service.py
bkeryan: Values without type annotations have type `typing.Any`, which means you can treat them as any type, call any method on them, etc. Not to be confused with `any_pb2.Any` :D

tests/acceptance/test_v2_measurement_service.py
DelpireNI: Yes I did. `test___measurement_service_v2___measure___returns_output` still has a docstring.

dixonjoel (approved):

tests/acceptance/test_streaming_data_measurement.py
```diff
@@ -0,0 +1,143 @@
...
+def test___streaming_measurement_service___request_number_of_responses___receives_responses(
+    num_responses: int, stub_v2: v2_measurement_service_pb2_grpc.MeasurementServiceStub
+):
+    request = v2_measurement_service_pb2.MeasureRequest(
+        configuration_parameters=_get_configuration_parameters(num_responses=num_responses)
+    )
+
+    response_iterator = stub_v2.Measure(request)
+    responses = [response for response in response_iterator]
+    assert len(responses) == num_responses
```
bkeryan, line 27: 3 sections: arrange, act, assert
```suggestion
    response_iterator = stub_v2.Measure(request)

    responses = [response for response in response_iterator]
    assert len(responses) == num_responses
```

tests/acceptance/test_streaming_data_measurement.py
```diff
@@ -0,0 +1,143 @@
...
+        configuration_parameters=_get_configuration_parameters(error_on_index=error_on_index)
+    )
+
+    response_iterator = stub_v2.Measure(request)
+
+    with pytest.raises(grpc.RpcError, match=f"Errored at index {error_on_index}"):
+        for index, response in enumerate(response_iterator):
+            pass
+
+    assert (index + 1) == error_on_index
```
bkeryan, line 91: 3 sections: arrange, act, assert
```suggestion
    with pytest.raises(grpc.RpcError, match=f"Errored at index {error_on_index}"):
        for index, response in enumerate(response_iterator):
            pass
    assert (index + 1) == error_on_index
```

bkeryan (approved):

Issue #259, opened 2023-05-02
`nidigital_spi` example returns errors when the service and TestStand sequence are in different directories


bkeryan:
# Bug Report

The `nidigital_spi` example resolves file paths differently in the service vs. the TestStand code module:
- The service resolves file paths with respect to the service directory.
- The TestStand code module resolves file paths with respect to the TestStand search paths, including the directory containing the sequence file.

If you statically register the example by copying it from the `measurementlink-python` Git repo to `C:\ProgramData\National Instruments\MeasurementLink\Services\nidigital_spi`, then the TestStand sequence in `C:\ProgramData` works but the TestStand sequence in the `measurementlink-python` Git repo fails.

## Repro or Code Sample

Run `python scripts\install_examples.py`.
Cd to `C:\dev\measurementlink-python\examples\nidigital_spi` and run `poetry install` to create a virtualenv for the TestStand code module.
Open `C:\dev\measurementlink-python\examples\nidigital_spi\NIDigitalSPI.seq` in TestStand and run it.

## Expected Behavior

The sequence passes.

## Current Behavior

The sequence fails with this error:

```
Error executing substep 'Post'.
Exception calling application: -1074118557: The specified levels or timing sheet name was not found. Ensure that you load levels and timing sheets before you applying them.

Device: DigitalPattern1
Levels Sheet: C:\ProgramData\National Instruments\MeasurementLink\Services\nidigital_spi\PinLevels.digilevels
```

This is because the TestStand code module loaded `C:\dev\measurementlink-python\examples\nidigital_spi\PinLevels.digilevels`, not `C:\ProgramData\National Instruments\MeasurementLink\Services\nidigital_spi\PinLevels.digilevels`.

## Your Environment

* OS & Device: Windows 10
* `ni-measurementlink-service` version: 1.0.1
* MeasurementLink version: 2023 Q2
* Python version: 3.9

AB#2381210

PR #282, opened 2023-06-07
Use timing and levels sheets name when applying rather than path

jonathanmendez:

Fixes Issue #259 

The [apply_levels_and_timing API ](https://nimi-python.readthedocs.io/en/master/nidigital/class.html#apply-levels-and-timing) allows the use of the file name (without extension) or the use of an absolute path. Since the file names are passed in as relative paths, while it is necessary to resolve to an absolute path if calling `load_specifications_levels_and_timing`, the name can be used when calling `apply_levels_and_timing`. This allows for when the files have been loaded from a different location at session initialization time, as is the case with the TestStand sequence.

Make the nidigital_spi example work in cases where the TestStand sequence initializes the session with files that are found in a different location than next to the measurement service.

Followed the steps to reproduce before the fix to see the error. Followed the steps to reproduce after the fix to see the sequence pass.

### DIFF
--- a/examples/nidigital_spi/measurement.py
+++ b/examples/nidigital_spi/measurement.py
@@ -150,17 +150,15 @@ def _create_nidigital_session(
 def _resolve_relative_path(
     directory_path: pathlib.Path, file_path: Union[str, pathlib.Path]
 ) -> pathlib.Path:
-    if not isinstance(file_path, pathlib.Path):
-        file_path = pathlib.Path(file_path)
+    file_path = pathlib.Path(file_path)
     if file_path.is_absolute():
         return file_path
     else:
         return (directory_path / file_path).resolve()
 
 
-def _get_file_name_from_path(file_path: Union[str, pathlib.Path]) -> str:
-    if not isinstance(file_path, pathlib.Path):
-        file_path = pathlib.Path(file_path)
+def _get_file_name_from_path(file_path: Union[str, pathlib.PurePath]) -> str:
+    file_path = pathlib.PurePath(file_path)
     return file_path.stem
 
 
--- a/examples/nidigital_spi/measurement.py
+++ b/examples/nidigital_spi/measurement.py
@@ -104,10 +104,9 @@ def measure(
                 str(_resolve_relative_path(service_directory, pattern_file_path)),
             )
 
-        selected_sites.apply_levels_and_timing(
-            str(_resolve_relative_path(service_directory, levels_file_path)),
-            str(_resolve_relative_path(service_directory, timing_file_path)),
-        )
+        levels_file_name = _get_file_name_from_path(levels_file_path)
+        timing_file_name = _get_file_name_from_path(timing_file_path)
+        selected_sites.apply_levels_and_timing(levels_file_name, timing_file_name)
         selected_sites.burst_pattern(start_label="SPI_Pattern")
         site_pass_fail = selected_sites.get_site_pass_fail()
         passing_sites = [site for site, pass_fail in site_pass_fail.items() if pass_fail]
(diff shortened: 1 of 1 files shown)


bkeryan (approved):

examples/nidigital_spi/measurement.py
```diff
@@ -159,6 +158,12 @@ def _resolve_relative_path(
         return (directory_path / file_path).resolve()
 
 
+def _get_file_name_from_path(file_path: Union[str, pathlib.Path]) -> str:
+    if not isinstance(file_path, pathlib.Path):
+        file_path = pathlib.Path(file_path)
```
mshafer-NI, line 163: hmmmm...

normally, you would just do something like
```python
pathlib_file_path = pathlib.Path(file_path)
return pathlib_file_path.stem
```

however, it is fair to say that always copying is likely (I haven't tested it) not as performant as an if-type.

jonathanmendez: I'm not python-enlightened enough to have an opinion, but I was copying the code pattern from lines 153-154, so I'm guessing this comment applies to that too?

I'm not sure why any of the usages of either function would ever be passing in a pathlib.Path based on my reading of the code. @bkeryan are these ever going to be pathlib.Paths?

Either way, does it maybe make sense to do the pathlib.Path() creation in the measure function before calling either of these, and simplifying these to expect a pathlib.Path always?

bkeryan: Matthew's suggestion is fine. The overhead of constructing an extra `pathlib.Path` is not important, so remove the type check.

Also, since this function doesn't access the filesystem, `pathlib.PurePath` is fine.

> I'm not sure why any of the usages of either function would ever be passing in a pathlib.Path based on my reading of the code. @bkeryan are these ever going to be pathlib.Paths?

I would prefer to represent paths as `pathlib.Path` or `pathlib.PurePath`, and I originally tried to use it in this example's `measure` function, but NI's Python APIs do not support `pathlib` yet so I switched to `str` before I posted the PR to create this example.

https://github.com/ni/measurementlink-python/issues/283
https://github.com/ni/nidaqmx-python/issues/26
https://github.com/ni/nimi-python/issues/1949

bkeryan: > pathlib_file_path = pathlib.Path(file_path)

FYI, I think mypy allows reusing a variable name for type narrowing.

image

mshafer-NI: it does -> I tend to use a new name to and try to communicate to the next reader (myself next week, or team-mate next month) why I'm doing a type conversion. (not saying this name is the best, though)

examples/nidigital_spi/measurement.py
```diff
@@ -104,10 +104,9 @@ def measure(
                 str(_resolve_relative_path(service_directory, pattern_file_path)),
             )
 
-        selected_sites.apply_levels_and_timing(
-            str(_resolve_relative_path(service_directory, levels_file_path)),
-            str(_resolve_relative_path(service_directory, timing_file_path)),
-        )
+        levels_file_name = _get_file_name_from_path(levels_file_path)
+        timing_file_name = _get_file_name_from_path(timing_file_path)
```
bkeryan, line 108: I guess we don't need a helper function for this after all.

```suggestion
        levels_file_name = pathlib.Path(levels_file_path).stem
        timing_file_name = pathlib.Path(timing_file_path).stem
```

bkeryan (approved):
Approved with suggestions

examples/nidigital_spi/measurement.py
jonathanmendez: Removed the helper

jonathanmendez:
I revalidated the fix after my latest changes. Completing this PR.

PR #286, opened 2023-06-13
examples: Update minimum ni-measurementlink-service version

bkeryan:

Update the minimum `ni-measurementlink-service` version for all of the examples to 1.1.

Remove `ni_measurementlink_service.*` from the `ignore_missing_imports` mypy config block.

Fix a mypy error in `nidigital_spi`: we defined `PinMapContext.sites` as optional, so clients must handle the case where it is None.

Preparation for enum support and updating `session_management` API

Ran mypy on each example with "../.." uncommented.

### DIFF
--- a/examples/nidaqmx_analog_input/pyproject.toml
+++ b/examples/nidaqmx_analog_input/pyproject.toml
@@ -7,7 +7,7 @@ authors = ["National Instruments"]
 [tool.poetry.dependencies]
 python = "^3.8"
 nidaqmx = { version = ">=0.8.0-dev0", extras = ["grpc"], allow-prereleases = true }
-ni-measurementlink-service = "^1.1.0-dev1"
+ni-measurementlink-service = {version = "^1.1.0-dev1", allow-prereleases = true}
 click = ">=7.1.2"
 
 [tool.poetry.group.dev.dependencies]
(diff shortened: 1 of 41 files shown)


examples/nidaqmx_analog_input/pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = ["National Instruments"]
 [tool.poetry.dependencies]
 python = "^3.8"
 nidaqmx = { version = ">=0.8.0-dev0", extras = ["grpc"], allow-prereleases = true }
-ni-measurementlink-service = "^1.0.0"
+ni-measurementlink-service = "^1.1.0-dev1"
```
dixonjoel, line 10: Do we not need the 'allow-prereleases' here for ni-measurementlink-service like we have for nidaqmx?

dixonjoel: Same question throughout.

dixonjoel (approved):

examples/nidaqmx_analog_input/pyproject.toml
bkeryan: Good catch. We probably need that to download the package from PyPI but not the local hard disk.

PR #287, opened 2023-06-13
service: Simplify single-session measurements

bkeryan:

`session_management.py`:
- Add `Client.reserve_session` (singular), which returns a `SingleSessionReservation`, which has a `session_info` property that returns a single `SessionInformation` object.
- Change `Client.reserve_sessions` (plural) to return a `MultiSessionReservation`, which has a `session_infos` property that returns a list of `SessionInformation` objects.
- Compatibility:
   - `Reservation` is an alias for `MultiSessionReservation` which reports a `DeprecationWarning`
   - `MultiSessionReservation.session_info` is an alias for `MultiSessionReservation.session_infos` which reports a `DeprecationWarning`

Update tests and examples to use updated APIs.

Add single-session and deprecation test cases.

Simplify single-session measurements by removing the need to check the number of session infos and index the session info list.

Ran updated unit tests.
Manually tested a couple examples in InstrumentStudio and TestStand with an earlier version of this branch.
TODO: re-test all examples
Ran mypy on all examples.

### DIFF
--- a/examples/nidaqmx_analog_input/pyproject.toml
+++ b/examples/nidaqmx_analog_input/pyproject.toml
@@ -7,7 +7,7 @@ authors = ["National Instruments"]
 [tool.poetry.dependencies]
 python = "^3.8"
 nidaqmx = { version = ">=0.8.0-dev0", extras = ["grpc"], allow-prereleases = true }
-ni-measurementlink-service = "^1.0.0"
+ni-measurementlink-service = "^1.1.0-dev1"
 click = ">=7.1.2"
 
 [tool.poetry.group.dev.dependencies]
@@ -24,7 +24,6 @@ build-backend = "poetry.core.masonry.api"
 
 [[tool.mypy.overrides]]
 module = [
-    "ni_measurementlink_service.*", # TODO: remove after 1.0.2 is released with py.typed marker
     "nidaqmx.*",
 ]
 ignore_missing_imports = true
--- a/examples/nidaqmx_analog_input/pyproject.toml
+++ b/examples/nidaqmx_analog_input/pyproject.toml
@@ -7,7 +7,7 @@ authors = ["National Instruments"]
 [tool.poetry.dependencies]
 python = "^3.8"
 nidaqmx = { version = ">=0.8.0-dev0", extras = ["grpc"], allow-prereleases = true }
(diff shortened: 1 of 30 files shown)


dixonjoel (approved):

examples/nidaqmx_analog_input/pyproject.toml
```diff
@@ -16,15 +16,15 @@ grpc-stubs = "^1.53"
 # Uncomment to use prerelease dependencies (requires poetry>=1.2).
 # TODO: Comment out once a prerelease is built.
 nidaqmx = { git = "https://github.com/ni/nidaqmx-python.git", branch = "master", extras = ["grpc"]}
-# ni-measurementlink-service = {path = "../..", develop = true}
+# Uncomment once 1.1.0-dev2 is released
+ni-measurementlink-service = {path = "../..", develop = true}
```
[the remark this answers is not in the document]
bkeryan: No, it should be uncommented until we build the next prerelease.

examples/nidcpower_source_dc_voltage/measurement.py
```diff
@@ -68,7 +68,7 @@ def measure(
 
     session_management_client = create_session_management_client(measurement_service)
 
-    with session_management_client.reserve_sessions(
+    with session_management_client.reserve_session(
```
bkeryan, line 71: FYI @jarnoldNI: this removes some more boilerplate code. If you only want one session, call reserve_session() instead of reserve_sessions(). It will validate that only one session is reserved, and `reservation.session_info` will be a single `SessionInformation` object instead of a list.

ni_measurementlink_service/session_management.py
```diff
@@ -138,27 +201,40 @@ def unreserve(self):
...
+        """Single session information object."""
+        assert len(self._session_info) == 1
+        return _convert_session_info_from_grpc(self._session_info[0])
+
+
+class MultiSessionReservation(BaseReservation):
+    """Manages reservation for multiple sessions."""
+
     @cached_property
     def session_info(self) -> List[SessionInformation]:
```
bkeryan, line 219: I originally wanted to deprecate this property and rename it to `session_infos`, but that would affect 100% of users. @dixonjoel and I discussed this earlier and I'm going to leave this property alone for multi-session reservations.

PR #288, opened 2023-06-15
Enum Annotations Support for MeasurementLink Python

WesleyTangNationalInstruments:

- Updated Configuration and Output decorator parameters to allow users to specify enum types.
- Automatically construct the annotation for enums based on the information provided in these decorators.
- Updated serialization strategy to include mappings for enum encoding and decoding.
- Updated serializer to encode and decode enums and store them as gRPC ints.

[AB#2364515](https://ni.visualstudio.com/DevCentral/_backlogs/backlog/InstrumentStudio%20Platform/Features/?workitem=2364515)

- Updated unit test coverage in `test_metadata.py`, `test_serializer.py` and `test_serialization_strategy.py` to include enums.
- Updated acceptance test coverage in `test_measurement_service.py`.
- Updated examples that already use enums via strings to use enums.
- Updated `sample_measurement` to include an enum input.
- Manually tested with sample_measurement. 

### Open Issues

- MeasurementUIEditor does not recognize enum output parameters for enum indicators and only recognizes them output parameters with the numeric indicator. Note that input enums work correctly and are recognized as possible input parameters for **_both_** numeric and enum indicators.


bkeryan:
mypy errors:
```
ni_measurementlink_service/_internal/parameter/metadata.py:127: error: Second argument of Enum() must be string, tuple, list or dict literal for mypy to determine Enum members  [misc]
ni_measurementlink_service/_internal/parameter/serializer.py:228: error: "type" has no attribute "__iter__" (not iterable)  [attr-defined]
ni_measurementlink_service/_internal/parameter/serializer.py:232: error: "type" has no attribute "__iter__" (not iterable)  [attr-defined]
ni_measurementlink_service/measurement/service.py:225: error: Incompatible default for argument "enum_type" (default has type "None", argument has type "Type[Enum]")  [assignment]
ni_measurementlink_service/measurement/service.py:[22](https://github.com/ni/measurementlink-python/actions/runs/5282558386/jobs/9558208369?pr=288#step:9:23)5: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
ni_measurementlink_service/measurement/service.py:225: note: Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
ni_measurementlink_service/measurement/service.py:279: error: Incompatible default for argument "enum_type" (default has type "None", argument has type "Type[Enum]")  [assignment]
ni_measurementlink_service/measurement/service.py:279: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
ni_measurementlink_service/measurement/service.py:279: note: Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
ni_measurementlink_service/measurement/service.py:349: error: Incompatible default for argument "enum_type" (default has type "None", argument has type "Type[Enum]")  [assignment]
ni_measurementlink_service/measurement/service.py:349: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
ni_measurementlink_service/measurement/service.py:349: note: Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
Found 6 errors in 3 files (checked 45 source files)
```

ni_measurementlink_service/_internal/parameter/metadata.py
```diff
@@ -53,17 +58,76 @@ def validate_default_value_type(parameter_metadata: ParameterMetadata) -> None:
...
         Context.get_type_default(parameter_metadata.type, parameter_metadata.repeated)
     )
-
-    if not isinstance(default_value, expected_type):
-        raise TypeError(
-            f"Unexpected type {type(default_value)} in the default value for '{display_name}'. Expected type: {expected_type}."
-        )
+    has_enum_values_annotation, enum_values_annotation = try_get_enum_values_annotation(
+        parameter_metadata
+    )
```
bkeryan, line 63: Applying the C# try-get pattern in Python seems cumbersome in this case.

I would prefer simpler helper functions like this:
```
def is_enum(parameter_metadata: ParameterMetadata) -> bool:
    return parameter_metadata.annotations.get("ni/type_specialization") == TypeSpecialization.Enum.value

def get_enum_annotations(parameter_metadata: ParameterMetadata) -> str:
    return parameter_metadata.annotations.get("ni/enum.values", "")
```

or this:

```
def get_enum_annotations(parameter_metadata: ParameterMetadata) -> str:
    if parameter_metadata.annotations.get("ni/type_specialization") == TypeSpecialization.Enum.value:
        return parameter_metadata.annotations.get("ni/enum.values", "")
    else:
        return ""
```

In Python, empty strings are "falsy" (treated as false when used in conditionals), so it's fine to write
```
enum_annotations = get_enum_annotations(...)
if enum_annotations:
   # do enum stuff
```

ni_measurementlink_service/_internal/parameter/metadata.py
```diff
@@ -53,17 +58,76 @@ def validate_default_value_type(parameter_metadata: ParameterMetadata) -> None:
...
+                f"Unexpected type {type(default_value)} in the default value for '{display_name}'. Expected type: {expected_type}."
+            )
         expected_element_type = type(Context.get_type_default(parameter_metadata.type, False))
-        for element in default_value:
-            if not isinstance(element, expected_element_type):
+        if has_enum_values_annotation:
+            for element in default_value:
+                if not _is_default_enum_match_annotations(element, enum_values_annotation):
+                    raise TypeError(
+                        f"The default value, `{element}`, for '{display_name}' is not valid. Expected values: `{enum_values_annotation}`."
```
bkeryan, line 75: IMO this buries the lede. "Invalid default value X for enum Y"?

ni_measurementlink_service/_internal/parameter/metadata.py
```diff
@@ -53,17 +58,76 @@ def validate_default_value_type(parameter_metadata: ParameterMetadata) -> None:
...
+        == TypeSpecialization.Enum.value
+        and "ni/enum.values" in parameter_metadata.annotations
+    ):
+        enum_values = parameter_metadata.annotations["ni/enum.values"]
+        if enum_values != "" and enum_values is not None:
+            return True, enum_values
+    return False, ""
+
+
+def _is_default_enum_match_annotations(default_value: Any, enum_values_annotation: str):
```
bkeryan, line 123: I find this function name hard to parse. Does `_is_valid_enum_value` describe what it's doing?

ni_measurementlink_service/_internal/parameter/metadata.py
```diff
@@ -53,17 +58,76 @@ def validate_default_value_type(parameter_metadata: ParameterMetadata) -> None:
...
+    if not isinstance(default_value, Enum):
+        return False
+    user_enum = json.loads(enum_values_annotation.replace("'", '"'))
+    user_enum_class = Enum("UserDefinedEnum", user_enum)
+    if (
+        any(member.value == default_value.value for member in user_enum_class)
+        and user_enum_class(default_value.value).name == default_value.name
+    ):
+        return True
+    return False
```
bkeryan, line 133: `json.loads()` returns a dict. Why construct a temporary enum class rather than validate against the dict? Does it handle some Python-specific enum naming special case? Or is this to look up keys by value?

ni_measurementlink_service/_internal/parameter/metadata.py
```diff
@@ -53,17 +58,76 @@ def validate_default_value_type(parameter_metadata: ParameterMetadata) -> None:
...
+        if not isinstance(default_value, expected_type):
+            raise TypeError(
+                f"Unexpected type {type(default_value)} in the default value for '{display_name}'. Expected type: {expected_type}."
+            )
         expected_element_type = type(Context.get_type_default(parameter_metadata.type, False))
-        for element in default_value:
-            if not isinstance(element, expected_element_type):
+        if has_enum_values_annotation:
+            for element in default_value:
+                if not _is_default_enum_match_annotations(element, enum_values_annotation):
```
bkeryan, line 73: This parses the JSON and constructs a temporary enum class for each element in the list.

bkeryan (changes requested):
I still haven't reviewed serializer.py or the tests.

ni_measurementlink_service/_internal/parameter/metadata.py
bkeryan: If the point is to verify that both the name and value are in the enum annotation and not some `Enum` class specific behavior, this should work:
```
user_enum = json.loads(enum_values_annotation.replace("'", '"'))
return default_value.name in user_enum and user_enum[default_value.name] == default_value.value
```

ni_measurementlink_service/_internal/parameter/metadata.py
bkeryan: Suggestion: parse the enum JSON once, outside the loop.

ni_measurementlink_service/_internal/parameter/serializer.py
```diff
@@ -180,3 +198,44 @@ def _get_missing_parameters(
...
+        has_enum_values_annotation, enum_values_annotation = try_get_enum_values_annotation(
+            parameter_metadata
+        )
+        if has_enum_values_annotation:
+            enum_type = _get_enum_type(parameter_metadata)
+            if parameter_metadata.repeated:
+                for j, member_value in enumerate(value):
+                    for enum in enum_type:
+                        if enum.value == member_value:
+                            parameter_by_id[i][j] = enum
```
bkeryan, line 227: `i` and `j` sound like indices. I think `id` and `index` would be clearer.

ni_measurementlink_service/_internal/parameter/metadata.py
WesleyTangNationalInstruments: Moved json parsing out of this helper. Also removed creating temporary enum class as part of addressing a different PR comment.

ni_measurementlink_service/_internal/parameter/metadata.py
WesleyTangNationalInstruments: Removed temporary enum class as it is unnecessary. I had not realized that `json.loads` returned an easy-to-use dict and thought it necessary to create an enum class to easily validate that the default value matched the enum information from the annotations.

ni_measurementlink_service/_internal/parameter/metadata.py
WesleyTangNationalInstruments: Renamed as suggested.

ni_measurementlink_service/_internal/parameter/metadata.py
WesleyTangNationalInstruments: Rewored as suggested.

ni_measurementlink_service/_internal/parameter/metadata.py
WesleyTangNationalInstruments: Fixed as suggested. Went with the second suggested option and updated some usages to conditionally check the result directly.

ni_measurementlink_service/measurement/service.py
```diff
@@ -243,14 +250,20 @@ def configuration(
             For custom instruments the user defined instrument type id is defined in the
             pin map file.
 
+            enum_type (Type[Enum]): Optional
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Updated to use the style `enum_type (Optional[Type[Enum]])` to be consistent with existing comments.

ni_measurementlink_service/measurement/service.py
```diff
@@ -215,7 +216,13 @@ def measure(configuration1, configuration2):
...
 
     def configuration(
-        self, display_name: str, type: DataType, default_value: Any, *, instrument_type: str = ""
+        self,
+        display_name: str,
+        type: DataType,
+        default_value: Any,
+        *,
+        instrument_type: str = "",
+        enum_type: Type[Enum] = None,
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: I [added a work item](https://ni.visualstudio.com/DevCentral/_backlogs/backlog/InstrumentStudio%20Platform/Features/?workitem=2422815) to cover support for protobuf enums and address this in a separate PR.

ni_measurementlink_service/measurement/service.py
```diff
@@ -322,7 +342,11 @@ def host_service(self) -> MeasurementService:
         return self
 
     def _get_annotations(
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Renamed as suggested.

ni_measurementlink_service/measurement/service.py
```diff
@@ -332,9 +356,18 @@ def _get_annotations(
         if type_specialization == TypeSpecialization.Pin:
             if instrument_type != "" or instrument_type is not None:
                 annotations["ni/pin.instrument_type"] = instrument_type
+        if type_specialization == TypeSpecialization.Enum:
+            if enum_type is not None:
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: No, we do not want to support a user not providing an enum type when specifying the enum type. It doesn't really make much sense when a user can just use the underlying data type instead. I updated this to throw an error if enum_type is not provided here.

ni_measurementlink_service/measurement/service.py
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Fixed as suggested.

ni_measurementlink_service/_internal/parameter/serializer.py
```diff
@@ -180,3 +198,44 @@ def _get_missing_parameters(
...
+            else:
+                for enum in enum_type:
+                    if enum.value == value:
+                        parameter_by_id[i] = enum
```
bkeryan, line 231: Iterating over all of the possible enum values is not a performant way to deserialize an enum. How about
`parameter_by_id[i] = enum_type(value)`?

```
>>> from enum import Enum
>>> class Color(Enum):
...     RED = 1
...     GREEN = 2
...     BLUE = 3
...
>>> Color(3)
<Color.BLUE: 3>
```

If the value is not in the enum, your current code leaves it unassigned. Is that really what you want? Or should the measurement service report an error?

Constructing an enum with an unsupported value raises `ValueError`. You can catch this exception and implement whatever behavior you want (leave the value unassigned, set it to None, set it to default, etc.).
```
>>> Color(4)
  File "C:\Users\Administrator\.pyenv\pyenv-win\versions\3.9.13\lib\enum.py", line 384, in __call__
    return cls.__new__(cls, value)
  File "C:\Users\Administrator\.pyenv\pyenv-win\versions\3.9.13\lib\enum.py", line 702, in __new__
    raise ve_exc
ValueError: 4 is not a valid Color
```

tests/unit/test_serializer.py
```diff
@@ -268,6 +292,26 @@ def _get_test_parameter_by_id(default_values):
...
             annotations={},
         ),
+        17: ParameterMetadata(
+            display_name="enum_data",
+            type=type_pb2.Field.TYPE_ENUM,
+            repeated=False,
+            default_value=default_values[16],
+            annotations={
+                "ni/type_specialization": TypeSpecialization.Enum.value,
+                "ni/enum.values": "{'PURPLE': 0, 'ORANGE': 1, 'TEAL': 2, 'BROWN': 3}",
```
bkeryan, line 302: Why is this using single quotes for the JSON key names? JSON strings don't support single quotes.

ni_measurementlink_service/_internal/parameter/metadata.py
```diff
@@ -53,17 +58,76 @@ def validate_default_value_type(parameter_metadata: ParameterMetadata) -> None:
...
+        enum_values = parameter_metadata.annotations["ni/enum.values"]
+        if enum_values != "" and enum_values is not None:
+            return True, enum_values
+    return False, ""
+
+
+def _is_default_enum_match_annotations(default_value: Any, enum_values_annotation: str):
+    if not isinstance(default_value, Enum):
+        return False
+    user_enum = json.loads(enum_values_annotation.replace("'", '"'))
```
bkeryan: Why is this converting single quotes to double quotes? I thought the annotation was JSON, which doesn't support single quotes for strings.

ni_measurementlink_service/_internal/parameter/serializer.py
```diff
@@ -67,6 +75,16 @@ def serialize_parameters(
...
             parameter_metadata.repeated,
         )
+        # Convert enum parameters to their underlying value.
+        if "ni/type_specialization" in parameter_metadata.annotations:
+            if (
+                parameter_metadata.annotations["ni/type_specialization"]
+                == TypeSpecialization.Enum.value
+            ):
+                if parameter_metadata.repeated:
+                    parameter = list(map(lambda x: x.value, parameter))
```
bkeryan, line 85: List comprehensions are preferred.
```suggestion
                    parameter = [x.value for x in parameter]
```

ni_measurementlink_service/_internal/parameter/serializer.py
```diff
@@ -67,6 +75,16 @@ def serialize_parameters(
             parameter_metadata.type,
             parameter_metadata.repeated,
         )
+        # Convert enum parameters to their underlying value.
+        if "ni/type_specialization" in parameter_metadata.annotations:
+            if (
+                parameter_metadata.annotations["ni/type_specialization"]
+                == TypeSpecialization.Enum.value
+            ):
```
bkeryan, line 83:
```suggestion
        if parameter_metadata.annotations.get("ni/type_specialization") == TypeSpecialization.Enum.value:
```
and fix indent

bkeryan (changes requested):
Minor details aside, this looks great. I'm really happy to have real enum support and stop using stringly typed enums.

ni_measurementlink_service/_internal/parameter/metadata.py
WesleyTangNationalInstruments: Didn't see your latest comment. I updated to take your simplified logic as suggested.

ni_measurementlink_service/_internal/parameter/metadata.py
WesleyTangNationalInstruments: Removed as suggested. This seemed to be an artifact from incorrect JSON string constants from the tests as you mentioned in the other comment.

tests/unit/test_serializer.py
WesleyTangNationalInstruments: Hmm, this appears incorrect. Switched to using `"` which coincidentally removed the need for `.replace("'", '"')` I used in `json.load`.

ni_measurementlink_service/_internal/parameter/serializer.py
WesleyTangNationalInstruments: Note that the indentation becomes a bit wonky again after running `poetry run ni-python-styleguide fix`. 

image

ni_measurementlink_service/_internal/parameter/metadata.py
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Fixed as suggested.

ni_measurementlink_service/_internal/parameter/metadata.py
```diff
@@ -53,17 +58,76 @@ def validate_default_value_type(parameter_metadata: ParameterMetadata) -> None:
...
-
-    if not isinstance(default_value, expected_type):
-        raise TypeError(
-            f"Unexpected type {type(default_value)} in the default value for '{display_name}'. Expected type: {expected_type}."
-        )
+    has_enum_values_annotation, enum_values_annotation = try_get_enum_values_annotation(
+        parameter_metadata
+    )
 
     if parameter_metadata.repeated:
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Split into two methods as suggested.

examples/sample_measurement/measurement.py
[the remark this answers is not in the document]
bkeryan: It's submitted.

You will need to edit the examples' pyproject.tomls and uncommit the dev-dependency reference to ni-measurementlink-service at path ../.. to make the example checks pass with your new feature, until we build a prerelease with this feature. Don't comment out the non-dev reference to ni-measurementlink-service.

bkeryan:
FYI, I updated the GitHub workflow names, so you may need to merge main before checks pass.

examples/nivisa_dmm_measurement/measurement.py
```diff
@@ -41,32 +41,56 @@
...
+    FourPointFive = auto()
+    FivePointFive = auto()
+    SixPointFive = auto()
+
+
+RESOLUTION_DIGITS_TO_DISPLAY_VALUE = {
+    ResolutionDigits.ThreePointFive: 0.001,
+    ResolutionDigits.FourPointFive: 0.0001,
+    ResolutionDigits.FivePointFive: 1e-5,
+    ResolutionDigits.SixPointFive: 1e-6,
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Reverted changes to resolution_digits so that it uses a numeric control again.

examples/nivisa_dmm_measurement/measurement.py
```diff
@@ -41,32 +41,56 @@
...
-    "6.5": 1e-6,
+
+class Function(Enum):
+    """Function that represents the measurement type."""
+
+    DCVolts = auto()
+    ACVolts = auto()
+
+
+FUNCTION_TO_DISPLAY_VALUE = {
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Renamed to "FUNCTIONS_TO_VALUE".

ni_measurementlink_service/measurement/service.py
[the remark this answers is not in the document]
WesleyTangNationalInstruments: I had some confusion with addressing some of the `mypy` issues that led me to add it as the default value, but I realize I was interpreting it wrong. I've updated this to `enum_type: Optional[Type[Enum]] = None` a ssuggested.

examples/nivisa_dmm_measurement/measurement.py
```diff
@@ -41,32 +41,56 @@
...
-RESOLUTION_DIGITS_TO_VALUE = {
-    "3.5": 0.001,
-    "4.5": 0.0001,
-    "5.5": 1e-5,
-    "6.5": 1e-6,
+
+class Function(Enum):
+    """Function that represents the measurement type."""
+
+    DCVolts = auto()
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Fixed as suggested.

ni_measurementlink_service/_internal/parameter/metadata.py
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Split off enums from basic types and refactored to avoid duplication. Updated terminology to use 'scalar' vs '`basic' better.

ni_measurementlink_service/_internal/parameter/serializer.py
WesleyTangNationalInstruments: Updated as suggested and to throw TypeError exceptions for when the value is not a valid value within the enum. I thought an error would be more appropriate than leaving it was unassigned, None, or default as if a user is giving us bad values here, we should be letting the user know.

ni_measurementlink_service/_internal/parameter/metadata.py
```diff
@@ -50,51 +50,97 @@ def validate_default_value_type(parameter_metadata: ParameterMetadata) -> None:
...
+        _validate_default_value_type_for_basic_type(
+            default_value, expected_type, enum_values_annotation, display_name
+        )
+    return None
+
+
+def _validate_default_value_type_for_basic_type(
+    default_value: Any, expected_type: type, enum_values_annotation: str, display_name: str
+) -> None:
+    """Validate and raise exception if the default value does not match the type info.
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Shortened docstrings for private methods in this file to remove argus/raises as I agree they are unnecessary. There are existing private helpers in serializer that have full docstrings, but I left those alone. Additionally, I added one private helper, `_deserialize_enum_parameters`, in serializer which seems more valuable than the ones pointed out here so I left that in as well.

examples/sample_measurement/measurement.py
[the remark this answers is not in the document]
WesleyTangNationalInstruments: I un-commented the dev dependencies for ni-measurementlink-service as suggested and added a TODO to re-comment them after we build a prerelease with this feature for the examples that were modified in this PR.

examples/nivisa_dmm_measurement/measurement.py
```diff
@@ -109,8 +133,8 @@ def measure(
             if not session_info.session_exists:
                 reset_instrument(session)
 
-            function_enum = str_to_enum(FUNCTION_TO_ENUM, measurement_type)
-            resolution_value = str_to_enum(RESOLUTION_DIGITS_TO_VALUE, str(resolution_digits))
+            function_enum = FUNCTION_TO_DISPLAY_VALUE[measurement_type]
+            resolution_value = RESOLUTION_DIGITS_TO_DISPLAY_VALUE[resolution_digits]
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: I reverted to using str_to_enum for resolution_digits in this case so I believe this should no longer be an issue.

It is difficult but not impossible to pass in an invalid enum. If the measurement panel is set up to use an enum control, we do not allow the user to pass in invalid values. However, if the enum configuration can still be bound to a numeric control, then it would be possible. Additionally, if the enum control in the measui file does not match the enum in the service, then it may also be possible to pass in an invalid value. This should all get handled in the validation we do during deserialization though so I don't think it's an issue. Let me know if I'm misinterpreting your concern.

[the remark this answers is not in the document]
bkeryan: This is the only remaining call to `str_to_enum` so I think it's appropriate to remove it from `_helpers.py` and change this example to do the lookup directly.

examples/nifgen_standard_function/measurement.py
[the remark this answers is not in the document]
WesleyTangNationalInstruments: I created a [separate work item](https://ni.visualstudio.com/DevCentral/_backlogs/backlog/InstrumentStudio%20Platform/Features/?workitem=2426140) for this and will put those changes up in a separate PR. On retrospect, I think it would've been better to have all of the example changes in a separate PR and it's sort of weird to have the changes to examples split between this PR and another one, but I think in the spirit of pursuing task-sized PRs, that'll be better.

nstokes-nati (approved):

tests/unit/test_metadata.py
```diff
@@ -1,48 +1,109 @@
...
+from enum import Enum
+
 import pytest
 
 from ni_measurementlink_service._internal.parameter import metadata
-from ni_measurementlink_service.measurement.info import DataType
+from ni_measurementlink_service.measurement.info import DataType, TypeSpecialization
+
+
+class Color(Enum):
```
dixonjoel, line 10: @synesthetize I vaguely remember something about enums having to have a 0 value. Did we have a restriction like this or associated validation?

dixonjoel (approved):
Just a few minor comments.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -113,13 +113,15 @@ def vector_decoder(field_index, key):
 UIntEncoder = _scalar_encoder(encoder.UInt32Encoder)
 BoolEncoder = _scalar_encoder(encoder.BoolEncoder)
 StringEncoder = _scalar_encoder(encoder.StringEncoder)
+EnumEncoder = _scalar_encoder(encoder.Int32Encoder)
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: I agree that it's unnecessary. I thought it was slightly easier to read, but on retrospect, I don't think it's necessary. I've removed this and switched to just using int encoders and decoders.

tests/unit/test_metadata.py
WesleyTangNationalInstruments: Ah, I overlooked this requirement. I've added a check to validate enums have a zero value and added a test for this. Also fixed up existing tests.

ni_measurementlink_service/_internal/parameter/metadata.py
```diff
@@ -50,51 +50,95 @@ def validate_default_value_type(parameter_metadata: ParameterMetadata) -> None:
...
+            )
+    else:
+        for element in default_value:
+            _validate_default_value_type_for_basic_type(
+                element, expected_element_type, display_name
+            )
+
+
+def _validate_default_value_type_for_basic_type(
+    default_value: Any,
```
bkeryan, line 119: FYI: https://mypy.readthedocs.io/en/stable/dynamic_typing.html#any-vs-object

ni_measurementlink_service/_internal/parameter/serializer.py
```diff
@@ -216,22 +215,25 @@ def _deserialize_enum_parameters(
...
-                    for enum in enum_type:
-                        if enum.value == member_value:
-                            parameter_by_id[i][j] = enum
+                for index, member_value in enumerate(value):
+                    try:
+                        parameter_by_id[id][index] = enum_type(member_value)
+                    except ValueError:
+                        raise TypeError(
+                            f"Unexpected value {member_value} is not valid for enum type '{enum_type.__name__}'."
+                        )
```
bkeryan, line 229: https://docs.python.org/3/library/exceptions.html#TypeError

> Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

Seems like "unexpected value X" should be a ValueError. 

Do you really need to customize the error message? The default is something like `ValueError: 1234 is not a valid UsageTypeAI`.

tests/unit/test_service.py
```diff
@@ -260,6 +278,26 @@ def test___service_config___create_measurement_service___service_info_matches_se
...
 
 
+@pytest.mark.parametrize(
+    "display_name,type,default_value,enum_type",
+    [
+        ("EnumConfiguration", DataType.Enum, Color.GREEN, None),
+        ("EnumConfiguration", DataType.Enum, ColorWithoutZeroValue.GREEN, ColorWithoutZeroValue),
+    ],
+)
+def test___measurement_service___add_configuration_with_invalid_enum_type__raises_type_error(
```
bkeryan, line 288: 3 underscores
```suggestion
def test___measurement_service___add_configuration_with_invalid_enum_type___raises_type_error(
```

examples/nivisa_dmm_measurement/_helpers.py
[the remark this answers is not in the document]
WesleyTangNationalInstruments: I standardized `_helpers.py` across all examples. Note that this file for `nivisa_dmm_measurement` was out of date and did not contain some changes to `get_grpc_device_channel` that the other `_helpers.py` had so I removed `str_to_enum` from one of the others and then propogated that `_helpers.py` across all examples so now all copies should reflect the latest.

ni_measurementlink_service/_internal/parameter/metadata.py
```diff
@@ -50,51 +50,95 @@ def validate_default_value_type(parameter_metadata: ParameterMetadata) -> None:
...
     return None
 
 
-def try_get_enum_values_annotation(parameter_metadata: ParameterMetadata) -> Tuple[bool, str]:
+def _validate_default_value_type_for_scalar_type(
+    default_value: Any, expected_type: type, enum_values_annotation: str, display_name: str
+) -> None:
+    """Validate and raise exception if the default value does not match the type info."""
+    if enum_values_annotation:
+        user_enum_dict = json.loads(enum_values_annotation.replace("'", '"'))
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Whoops. I removed it in one place, but missed the other. Also fixed a related issue where we were generating enum values annotations that required this.

ni_measurementlink_service/_internal/parameter/metadata.py
WesleyTangNationalInstruments: Switched to using `object` for these private validation helpers instead of `Any`.

ni_measurementlink_service/_internal/parameter/serializer.py
WesleyTangNationalInstruments: Removed custom error message. I agree that default should be fine.

examples/nivisa_dmm_measurement/_helpers.py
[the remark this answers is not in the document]
bkeryan: Thanks. @subash-suresh is adding `_helpers.py` to the ni-measurementlink-generator template, so I suggested that we also add a validation check to make sure each example has an up-to-date `_helpers.py`: https://github.com/ni/measurementlink-python/pull/293/files#r1238997517

bkeryan (approved):
Looks good, ship it!

PR #294, opened 2023-06-22
Add annotations to generator

gpachecoNI:

TODO: Check the above box with an 'x' indicating you've read and followed [CONTRIBUTING.md](https://github.com/ni/measurement-services-python/blob/main/CONTRIBUTING.md).

Includes the service annotation support in the generator. 
- Add Description, Collection, and Tags to the  .serviceconfig
- Provide optional command line options for the user. If not assigned just use "" (Empty)

[AB#2408665](https://dev.azure.com/ni/DevCentral/_workitems/edit/2408665)

 - Add test with all new command lines.
 - Keep test where annotations were not being inputted to ensure they are optional.


ni_measurementlink_generator/tests/test_assets/example_renders/measurementWithAnnotations.py
```diff
@@ -0,0 +1,47 @@
+"""A default measurement with an array in and out."""
+import logging
+import pathlib
+
+import click
+import ni_measurementlink_service as nims
+
+service_directory = pathlib.Path(__file__).resolve().parent
+measurement_service = nims.MeasurementService(
+    service_config_path=service_directory / "SampleMeasurementWithAnnotations.serviceconfig",
```
gpachecoNI, line 10: This is the only difference between the measurement.py file and this.

gpachecoNI:
I will change the target branch to Main once my other PR gets in.

bkeryan:
FYI, I created a releases/1.1 branch and updated main to version 1.2.0-dev0 in order to unblock these PRs.

ni_measurementlink_generator/tests/test_assets/example_renders/SampleMeasurementWithAnnotations.serviceconfig
```diff
@@ -0,0 +1,16 @@
...
+    {
+      "displayName": "Sample Measurement With Annotations",
+      "serviceClass": "SampleMeasurement_Python",
+      "descriptionUrl": "https://www.example.com/SampleMeasurement.html",
+      "providedInterfaces": [ "ni.measurementlink.measurement.v1.MeasurementService" ],
+      "path": "start.bat",
+      "annotations": {
+        "ni/service.description": "Measurement description",
+        "ni/service.collection": "Measurement collection",
+        "ni/service.tags": ['M1', 'M2', 'M3']
```
cumitche, line 12: JSON strings must be double quotes, not single.

gpachecoNI: Fixed

ni_measurementlink_generator/ni_measurementlink_generator/template.py
```diff
@@ -104,6 +105,22 @@ def _resolve_service_class(service_class: str, display_name: str) -> str:
...
 )
+@click.option(
+    "-a",
+    "--annotations-description",
+    help="Short description for the Measurement",
+)
+@click.option(
+    "-c",
+    "--collection",
+    help="The collection that this measurement belongs to.",
```
dixonjoel, line 116: Can we add something like this to clarify (text from the .proto file)?
//   - Expected format: "." delimited namespace/hierarchy case-insensitive string
//   - Example: "CurrentTests.Inrush"

ni_measurementlink_generator/ni_measurementlink_generator/template.py
```diff
@@ -146,6 +166,12 @@ def create_measurement(
 
     directory_out_path.mkdir(exist_ok=True, parents=True)
 
+    if annotations_description is None:
+        annotations_description = ""
+    if collection is None:
+        collection = ""
+    annotation_tags = json.dumps(tags)
```
dixonjoel, line 173: Why aren't we checking for None on the tags like the other two? Are there tests that we handle None for these three?

ni_measurementlink_generator/tests/test_assets/example_renders/SampleMeasurement.serviceconfig
```diff
@@ -5,7 +5,12 @@
       "serviceClass": "SampleMeasurement_Python",
       "descriptionUrl": "https://www.example.com/SampleMeasurement.html",
       "providedInterfaces": [ "ni.measurementlink.measurement.v1.MeasurementService" ],
-      "path": "start.bat"
+      "path": "start.bat",
+      "annotations": {
+        "ni/service.description": "",
+        "ni/service.collection": "",
+        "ni/service.tags": []
```
dixonjoel, line 12: I'm confused as to why the tags are typed as Optional[str] just like the other two, but we treat them as an array here.

dixonjoel (changes requested):

ni_measurementlink_generator/ni_measurementlink_generator/template.py
gpachecoNI: json.dumps takes care of none... so I remove it.  
Yes there is one test.

ni_measurementlink_generator/tests/test_assets/example_renders/SampleMeasurement.serviceconfig
gpachecoNI: It is because on the @ClickOption, I set it as multiple but each is entered as a string.

ni_measurementlink_generator/ni_measurementlink_generator/template.py
dixonjoel: Yes, I was thinking to add it to this help.
help="The collection that this measurement belongs to."

ni_measurementlink_generator/tests/test_assets/example_renders/SampleMeasurement.serviceconfig
dixonjoel: I see that Multiple=True now.

ni_measurementlink_generator/ni_measurementlink_generator/template.py
```diff
@@ -104,6 +105,22 @@ def _resolve_service_class(service_class: str, display_name: str) -> str:
     "--directory-out",
     help="Output directory for measurement files. Default: '<current_directory>/<display_name>'",
 )
+@click.option(
+    "-a",
+    "--annotations-description",
```
bkeryan, line 110: "annotations" is an implementation detail. As far as users are concerned, this is "the description".

I recommend renaming this to `--description` and moving it above `--description-url` in both the options and all parameter lists. 

Unfortunately, we wasted `-d` on `--description-url`, so we can't use it for `--description`. You could use `-D`, though.

BTW, if we were leaving `--annotations-description` as it is, I would prefer not using a short option rather than wasting `-a` on something that no one will remember is connected to the letter "A".

ni_measurementlink_generator/ni_measurementlink_generator/template.py
bkeryan, line 173: I think conversion to JSON belongs in the template, not the command line parser.

ni_measurementlink_generator/tests/test_assets/example_renders/SampleMeasurement.serviceconfig
bkeryan: I see how this works. See my comment on the serviceconfig template.

ni_measurementlink_generator/ni_measurementlink_generator/templates/measurement.serviceconfig.mako
```diff
@@ -7,7 +7,12 @@
       "serviceClass": "${service_class}",
       "descriptionUrl": "${description_url}",
       "providedInterfaces": [ "ni.measurementlink.measurement.v1.MeasurementService" ],
-      "path": "start.bat"
+      "path": "start.bat",
+      "annotations": {
+        "ni/service.description": "${annotations_description}",
+        "ni/service.collection": "${collection}",
+        "ni/service.tags": ${tags}
```
bkeryan, line 14: This highlights the fact that we are generating JSON with a mako template. This is a bad practice because there are a lot of cases it doesn't handle correctly. For example, strings that contain quotes or whitespace should use JSON escape sequences.

It would be better to build a nested dict and use `json.dumps` to convert it to JSON:

```
<% 
    service_config = {
        "services": [
            {
                "displayName": display_name,
                "serviceClass": service_class,
                ...
            }
        ]
    }
%>\
${json.dumps(service_config, indent=2)}
```

Dict preserves key order in recent Python versions, so I don't think we need OrderedDict to have a stable key order.

You could add `separators=` if needed but I don't know if we care.

ni_measurementlink_generator/tests/acceptance/test_generator.py
```diff
@@ -37,6 +37,54 @@ def test___command_line_args___create_measurement___render_without_exception(
...
+    filenames = [
+        "measurementWithAnnotations.py",
+        "SampleMeasurementWithAnnotations.serviceconfig",
+        "start.bat",
+    ]
+    for filename in filenames:
+        golden_file = golden_path / filename
+        # We always generate only measurement.py
+        if filename == "measurementWithAnnotations.py":
+            filename = "measurement.py"
```
bkeryan, line 83: I would prefer to put each test case's render in a separate subdirectory rather than special-case the file comparisons.

bkeryan (changes requested):

ni_measurementlink_generator/ni_measurementlink_generator/templates/measurement.serviceconfig.mako
gpachecoNI: I don't think we have any problem with the whitespace, since the description already has them. Tbh I don't think we are expecting any complex string input from the customer on these specific annotations. If this is important for the customer we can add this.

ni_measurementlink_generator/tests/acceptance/test_generator.py
gpachecoNI: Done

ni_measurementlink_generator/ni_measurementlink_generator/template.py
gpachecoNI: Fixed

ni_measurementlink_generator/ni_measurementlink_generator/template.py
gpachecoNI: Is this the command line parser? I am just manipulating data before cre_file the same way other stuff is being built on top.

bkeryan: Pass the list into the template. Let the template handle presentation in JSON or whatever format it wants.

ni_measurementlink_generator/ni_measurementlink_generator/templates/measurement.serviceconfig.mako
bkeryan: I don't want to maintain a template that generates JSON in this way.

I posted a separate PR to fix this: https://github.com/ni/measurementlink-python/pull/314

bkeryan (changes requested):

ni_measurementlink_generator/ni_measurementlink_generator/templates/measurement.serviceconfig.mako
gpachecoNI: hey Brad,  sorry I missed you posted the whole solution here.

ni_measurementlink_generator/ni_measurementlink_generator/template.py
```diff
@@ -104,6 +111,23 @@ def _resolve_service_class(service_class: str, display_name: str) -> str:
     "--directory-out",
     help="Output directory for measurement files. Default: '<current_directory>/<display_name>'",
 )
+@click.option(
+    "-c",
+    "--collection",
+    default="",
+    help='\b\nThe collection that this measurement belongs to.'
+    + '\n-Expected format: "." delimited namespace/hierarchy case-insensitive string'
+    + '\n-Example: "CurrentTests.Inrush"',
```
bkeryan, line 120: What is `\b\n` supposed to do? It doesn't seem to do anything.

Is the dash in `\n-Expected` intended to be a bullet? It doesn't look like one. 

The help for the other options uses `Default: x` without a line break or bullet, so I think you should remove the newlines and dashes. 

The help for the other options displays single quotes, not double quotes.

Example to add 2 tags has an odd number of double quotes.

```
Options:
  --measurement-version TEXT  Version number in the form 'x.y.z.q'. Default:
                              '1.0.0.0'
  -u, --ui-file TEXT          Name of the UI File. Default:
                              '<display_name>.measui'
  -s, --service-class TEXT    Service Class that the measurement belongs to.
                              Default: '<display_name>_Python'
  -D, --description TEXT      Short description of the measurement
  -d, --description-url TEXT  Description URL that links to a web page
                              containing information about the measurement.
  -o, --directory-out TEXT    Output directory for measurement files. Default:
                              '<current_directory>/<display_name>'
  -c, --collection TEXT       The collection that this measurement belongs to.
                              -Expected format: "." delimited namespace/hierarchy case-insensitive string
                              -Example: "CurrentTests.Inrush"
  -t, --tags TEXT             Add one or multiple tags for the measurement adding one by one.
                              -Expected format: Enter one tag per -t option.
                              -Example to add 2 tags: "-t "test" -t "Internal"
  -v, --verbose               Enable verbose logging. Repeat to increase
                              verbosity.
  --help                      Show this message and exit.
```

Suggestion: 
- `help="The collection that this measurement belongs to. Collection names are specified using period-delimited namespace hierarchy and are case-insensitive. Example: 'CurrentTests.Inrush'"`
- `help="Tags describing the measurement. This option may be repeated to specify multiple tags. Example: '-t test -t internal'"`

ni_measurementlink_generator/tests/test_assets/example_renders/SampleMeasurement.serviceconfig
bkeryan, line 1: Move the old example render into a subdirectory too. You can call it `measurement` and call your new directory `measurement_with_annotations` so that they match the test case names.

ni_measurementlink_generator/tests/acceptance/test_generator.py
```diff
@@ -37,6 +37,47 @@ def test___command_line_args___create_measurement___render_without_exception(
...
+                "--service-class",
+                "SampleMeasurement_Python",
+                "-D",
+                "Measurement description",
+                "--description-url",
+                "https://www.example.com/SampleMeasurement.html",
+                "--directory-out",
+                temp_directory,
+                "--collection",
+                "Measurement collection",
```
bkeryan, line 62: I heard this is supposed to be a period-delimited namespace hierarchy.

bkeryan (changes requested):

ni_measurementlink_generator/ni_measurementlink_generator/template.py
gpachecoNI: The \b\n allows  the -Expected and -Example on a different line. Facilitates readability. 

https://stackoverflow.com/questions/55389049/why-are-newlines-added-to-the-help-under-python-click/55389152#55389152

bkeryan: Thanks, I didn't know that.

ni_measurementlink_generator/ni_measurementlink_generator/template.py
```diff
@@ -104,6 +110,22 @@ def _resolve_service_class(service_class: str, display_name: str) -> str:
     "--directory-out",
     help="Output directory for measurement files. Default: '<current_directory>/<display_name>'",
 )
+@click.option(
+    "-c",
+    "--collection",
+    default="",
+    help="\b\nThe collection that this measurement belongs to. Collection names are specified"\
```
bkeryan, line 117: Black doesn't use trailing backslashes.

```
PS D:\dev\measurementlink-python\ni_measurementlink_generator> poetry run ni-python-styleguide lint
.\ni_measurementlink_generator\template.py:117:95: BLK100 Black would make changes.
```

Why didn't "Check nimg" catch this?

ni_measurementlink_generator/ni_measurementlink_generator/template.py
```diff
@@ -104,6 +110,22 @@ def _resolve_service_class(service_class: str, display_name: str) -> str:
...
+        "using a period-delimited namespace hierarchy and are case-insensitive."\
+        "\nExample: 'CurrentTests.Inrush'",
+)
+@click.option(
+    "-t",
+    "--tags",
+    default=[],
+    multiple=True,
+    help="\b\nTags describing the measurement. This option may be repeated to specify multiple tags and are case-insensitive."\
```
bkeryan, line 126: subject/verb agreement: "option" is singular and "are" is plural

```suggestion
    help="\b\nTags describing the measurement. This option may be repeated to specify multiple tags. Tags are case-insensitive."\
```

ni_measurementlink_generator/ni_measurementlink_generator/template.py
```diff
@@ -117,6 +139,9 @@ def create_measurement(
     service_class: str,
     description_url: str,
     directory_out: Optional[str],
+    description: str,
+    collection: str,
+    tags: List[str],
```
bkeryan, line 144: This isn't a list after all.
```suggestion
    tags: Tuple[str,...],
```

ni_measurementlink_generator/ni_measurementlink_generator/template.py
```diff
@@ -146,6 +171,10 @@ def create_measurement(
 
     directory_out_path.mkdir(exist_ok=True, parents=True)
 
+    annotation_tag = []
+    for i in tags:
+        annotation_tag.append(i)
+
```
bkeryan, line 177:
```suggestion
```

ni_measurementlink_generator/ni_measurementlink_generator/template.py
```diff
@@ -166,6 +195,9 @@ def create_measurement(
         service_class=service_class,
         description_url=description_url,
         ui_file_type=ui_file_type,
+        description=description,
+        collection=collection,
+        tags=annotation_tag,
```
bkeryan, line 200:
```suggestion
        tags=list(tags),
```

bkeryan (changes requested):
This should be the last change

ni_measurementlink_generator/ni_measurementlink_generator/template.py
gpachecoNI: I guess you are right we are not expecting any change.

bkeryan: I expected Click to pass us a list when we use `default=[], multiple=True`, but apparently it uses a tuple.

https://click.palletsprojects.com/en/8.1.x/options/#multiple-options says the default can be a list or tuple but it doesn't really say that it converts from list to tuple.

bkeryan: `Tuple[str]` means a tuple with 1 element which is a `str`.

`Tuple[str,...]` means a tuple with any number of elements, which are `str`s.

dixonjoel (approved):

PR #297, opened 2023-06-23
Fixing issues with example enums

WesleyTangNationalInstruments:

- Fixed an issue where we were using the wrong enum type in some examples.

Addressing PR comments from #295.

- Manually validated that the examples services starts successfully.
- Manually validated that we can connect to the service in InstrumentStudio.

### DIFF
--- a/examples/nidaqmx_analog_input/_helpers.py
+++ b/examples/nidaqmx_analog_input/_helpers.py
@@ -294,7 +294,7 @@ def get_sessions_and_channels_for_pins(
     return session_and_channel_info
 
 
-def generate_wrapper_enum(enum_type: Type[Enum]) -> Enum:
+def generate_wrapper_enum(enum_type: Type[Enum]) -> Type[Enum]:
     """Wraps an enum that does not have a zero value with one that does."""
     wrapper_enum_dict = {k: v.value for k, v in enum_type.__members__.items()}
     if 0 not in wrapper_enum_dict.values():
--- a/examples/nidaqmx_analog_input/_helpers.py
+++ b/examples/nidaqmx_analog_input/_helpers.py
@@ -4,16 +4,7 @@ import logging
 import pathlib
 import types
 from enum import Enum
-from typing import (
-    Any,
-    Callable,
-    List,
-    NamedTuple,
-    Optional,
-    Tuple,
-    TypeVar,
-    Union,
-)
+from typing import Any, Callable, List, NamedTuple, Optional, Tuple, TypeVar, Union
 
 import click
 import grpc
@@ -303,10 +294,9 @@ def get_sessions_and_channels_for_pins(
     return session_and_channel_info
 
 
-def generate_wrapper_enum(
-    enum_type: Enum
-) -> Enum:
-    wrapper_enum_dict = {k:v.value for k,v in enum_type.__members__.items()}
+def generate_wrapper_enum(enum_type: Enum) -> Enum:
+    """Wraps an enum that does not have a zero value with one that does."""
+    wrapper_enum_dict = {k: v.value for k, v in enum_type.__members__.items()}
     if 0 not in wrapper_enum_dict.values():
         wrapper_enum_dict["NONE"] = 0
     return Enum(enum_type.__name__ + "Wrapper", wrapper_enum_dict)
--- a/examples/nidaqmx_analog_input/_helpers.py
+++ b/examples/nidaqmx_analog_input/_helpers.py
@@ -3,6 +3,7 @@
 import logging
 import pathlib
 import types
+from enum import Enum
 from typing import (
     Any,
     Callable,
@@ -300,3 +301,12 @@ def get_sessions_and_channels_for_pins(
         raise KeyError(f"Pin(s) {pins} and site {site} not found")
 
     return session_and_channel_info
+
+
+def generate_wrapper_enum(
+    enum_type: Enum
+) -> Enum:
+    wrapper_enum_dict = {k:v.value for k,v in enum_type.__members__.items()}
+    if 0 not in wrapper_enum_dict.values():
+        wrapper_enum_dict["NONE"] = 0
+    return Enum(enum_type.__name__ + "Wrapper", wrapper_enum_dict)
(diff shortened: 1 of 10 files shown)


examples/nidaqmx_analog_input/_helpers.py
```diff
@@ -300,3 +301,12 @@ def get_sessions_and_channels_for_pins(
         raise KeyError(f"Pin(s) {pins} and site {site} not found")
 
     return session_and_channel_info
+
+
+def generate_wrapper_enum(
+    enum_type: Enum
+) -> Enum:
+    wrapper_enum_dict = {k:v.value for k,v in enum_type.__members__.items()}
```
bkeryan: >  `_helpers.py:309: error: "Enum" has no attribute "__members__"  [attr-defined]`

https://github.com/python/typeshed/blob/main/stdlib/enum.pyi has had `EnumMeta.__members__` since 2/22/2022.

I think this is failing because `enum_type: Enum` declares it to be an instance of `Enum`, not a type derived from `Enum`. Try `enum_type: Type[Enum]` and `-> Type[Enum]`

examples/nidmm_measurement/measurement.py
```diff
@@ -31,27 +32,7 @@
...
-    PERIOD = nidmm.Function.PERIOD.value
-    TEMPERATURE = nidmm.Function.TEMPERATURE.value
-    AC_VOLTS_DC_COUPLED = nidmm.Function.AC_VOLTS_DC_COUPLED.value
-    DIODE = nidmm.Function.DIODE.value
-    WAVEFORM_VOLTAGE = nidmm.Function.WAVEFORM_VOLTAGE.value
-    WAVEFORM_CURRENT = nidmm.Function.WAVEFORM_CURRENT.value
-    CAPACITANCE = nidmm.Function.CAPACITANCE.value
-    INDUCTANCE = nidmm.Function.INDUCTANCE.value
-
+FunctionWrapper = generate_wrapper_enum(nidmm.Function)
```
bkeryan, line 34: FYI, this is a tradeoff: dynamically generated enums don't have type annotations according to https://stackoverflow.com/a/74335189 so mypy can't check whether `FunctionWrapper.SDKFJDKFJ` is valid and VS Code can't auto-complete `FunctionWrapper.` into `FunctionWrapper.DC_VOLTS`.

I'm not sure whether this is more or less error-prone than listing out all of the enum values like in your previous PR. Neither seems ideal.

It would be more convenient if users could specify an enum that doesn't a 0 value and nims would automatically translate 0 -> default_value.

bkeryan (changes requested):

examples/nifgen_standard_function/measurement.py
```diff
@@ -76,7 +64,7 @@ def measure(
     logging.info(
         "Starting generation: pin_name=%s waveform_type=%s frequency=%g amplitude=%g",
         pin_name,
-        waveform_type if waveform_type != Waveform.NONE else Waveform.SINE,
+        nifgen.Waveform(waveform_type.value) if waveform_type.value != 0 else nifgen.Waveform.SINE,
```
bkeryan, line 67: Translating enums when logging seems unnecessary

examples/nidaqmx_analog_input/_helpers.py
bkeryan: And yeah, mypy can't handle enums with dynamic members.
https://mypy-play.net/?mypy=latest&python=3.11&gist=63649673382c32dfd9d77b912040997a
image

examples/nidmm_measurement/measurement.py
bkeryan: And yeah, mypy can't handle enums with dynamic members.
https://mypy-play.net/?mypy=latest&python=3.11&gist=63649673382c32dfd9d77b912040997a
image

examples/nifgen_standard_function/measurement.py
WesleyTangNationalInstruments: Removed as suggested.

examples/nidmm_measurement/measurement.py
bkeryan: Hold on, I have another idea I want to try out

bkeryan: https://docs.python.org/3/howto/enum.html#restricted-enum-subclassing explains why the obvious solution is not supported:
```
class FunctionWrapper(nidmm.Function):
    NONE = 0
```

It would be nice if we could do this:
```
class FunctionWrapper(nidmm.Function):
    @classmethod
    def _missing_(cls, value: object) -> Any:
        if value == 0:
            return cls.RED
        return super()._missing_(value)
```

However, subclassing an enum that has values is not supported, even if the subclass doesn't define any values.

examples/nidaqmx_analog_input/_helpers.py
bkeryan: > `
_helpers.py:302: error: Incompatible return value type (got "Enum", expected "type[Enum]")  [return-value]`

Believe it or not, splitting this into `enum_type = Enum(...)` and `return enum_type` seems to prevent mypy from thinking `Enum()` returns an instance of `Enum`. But I think you will run into other errors instead.

examples/nidmm_measurement/measurement.py
WesleyTangNationalInstruments:
```
class FunctionWrapper(nidmm.Function):
    NONE = 0
```

Yeah, I was slightly annoyed that python doesn't support enum subclassing like this since it feels so natural.

examples/nidaqmx_analog_input/_helpers.py
WesleyTangNationalInstruments: Yeah, I didn't have much luck in getting this to work either. I've reverted the change for dynamically creating enums.

dixonjoel (approved):

examples/nidaqmx_analog_input/_helpers.py
dixonjoel: @WesleyTangNationalInstruments please update the PR description not to mention the dynamic helper if it's gone.

bkeryan (approved):

PR #322, opened 2023-06-30
[releases/1.1] examples: Depend on released dependencies

bkeryan:

Update dependency versions and remove `allow-prereleases`.

Update version number to 1.1.0 so it will be correct in the source zip file.

Ensure that the examples in the released zip file do not automatically download pre-release versions of `ni-measurementlink-service`, `nidaqmx`, etc. from PyPI.

This won't pass the examples check because the dependencies aren't actually released yet.


bkeryan:
FYI @dixonjoel this probably should be the last PR we commit before making the final release of 1.1.0.

BTW, stuff like this is why examples have a separate check :)

PR #256, opened 2023-04-29
RFC: examples: Refactor TestStand code module and add multiple_measurements example

bkeryan:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/measurement-services-python/blob/main/CONTRIBUTING.md).

### What does this Pull Request accomplish?

Refactor TestStand code module to minimize impact of name conflicts between multiple examples:
- Move pin map support to `teststand_pin_map.py`
- Move modular instrument session support to `teststand_${driver}.py` (e.g. `teststand_nidcpower.py`)

Add a `multiple_measurements` example that demonstrates how to call multiple measurement services from TestStand. This example includes only the TestStand sequence and the relevant TestStand code modules.

### Why should this Pull Request be merged?

Make it easier to get started using multiple measurement services in TestStand.

### What testing has been done?

Manually tested.

PR #323, opened 2023-07-07
Add measurement example involving NI-DCPower and NI-VISA DMM instruments

samchris007:

This measurement example explains handling two different instruments namely an NI instrument(NI-DCPower) and a third-party instrument(NI-VISA DMM), within a single measurement service.
-  This example also defines creating sessions(reserving and registering sessions) for two completely different instruments within a single measurement service.

### Implementation
Add measurement example which involves sourcing NI-DCPower to supply an input pin and measuring the output pin using the NI-VISA DMM instrument and an NI Instrument Simulator v2.0..

- Implemented measure function by replicating sourcing and measuring logic from existing examples [nidcpower_source_dc_voltage](https://github.com/ni/measurementlink-python/tree/main/examples/nidcpower_source_dc_voltage) and [nivisa_dmm_measurement](https://github.com/ni/measurementlink-python/tree/main/examples/nivisa_dmm_measurement).
- For reserving and creating sessions,
  - Both input and output pins are reserved using `session_management_client.reserve_sessions()`
  - Session objects for both NI-DCPower and NI-VISA DMM instruments are created within a single `with` statement, 

Manually tested with both simulation and hardware.


examples/nidcpower_nivisa_dmm_measurement/measurement.py
```diff
@@ -0,0 +1,242 @@
...
+            measured_value = float(response)
+
+            source_session = None  # Don't abort after this point
+
+    logging.info("Completed measurement: measured_value=%g", measured_value)
+    return (measured_value,)
+
+
+def _get_session_info_from_pin(session_info, pin_name):
+    session_index = get_session_and_channel_for_pin(session_info, pin_name)[0]
```
keer1111, line 180: will this list be empty? should be checked?

examples/nidcpower_nivisa_dmm_measurement/_helpers.py
```diff
@@ -0,0 +1,306 @@
...
+    """Return created session management client."""
+    return nims.session_management.Client(
+        grpc_channel=measurement_service.get_channel(
+            provided_interface=nims.session_management.GRPC_SERVICE_INTERFACE_NAME,
+            service_class=nims.session_management.GRPC_SERVICE_CLASS,
+        )
+    )
+
+
+def get_session_and_channel_for_pin(
```
keer1111, line 271: public methods can have doc string?

bkeryan:
I really think we should revisit https://github.com/ni/measurementlink-python/pull/256

examples/nidcpower_nivisa_dmm_measurement/NIDCPowerNIVisaDmmMeasurement.instudioproj
bkeryan, line 1: This is missing TestStand support.

When you add it, please don't combine the NI-DCPower and VISA code into a single, monolithic `teststand_fixture.py` module. Use separate modules for pin map, NI-DCPower, and VISA so that we can "vendor" identical copies of these modules in multiple examples. I previously started to do this in #256, but I put it on hold due to the other example refactoring.

examples/nidcpower_nivisa_dmm_measurement/pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[tool.poetry]
+name = "nidcpower_nivisa_dmm_measurement"
```
bkeryan, line 2: I don't like this example name. It sounds like it's using both NI-DCPower and NI-VISA to do a DMM measurement, but NI-DCPower is not involved in the DMM measurement.

What is it really doing? 

It's using NI-DCPower to source a DC voltage for the DUT's input pin, and it's using VISA and a DMM to measure the DUT's output pin.

Is there a higher level way to describe this? 

"Stimulus response" comes to mind, but I think that might imply time domain signals, whereas this is working with DC signals.

What sort of DUT would you test with this measurement?

bkeryan (changes requested):

examples/nidcpower_nivisa_dmm_measurement/measurement.py
```diff
@@ -0,0 +1,242 @@
...
+
+RESERVATION_TIMEOUT_IN_SECONDS = 60.0
+"""
+If another measurement is using the session, the reserve function will wait
+for it to complete. Specify a reservation timeout to aid in debugging missed
+unreserve calls. Long measurements may require a longer timeout.
+"""
+
+
+class Function(Enum):
```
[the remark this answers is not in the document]
samchris007: I was looking into one of the LabVIEW examples for Agilent 34401 and I noticed that it has the same property name of `Function` for configuring the measurement. So, I think it's ok to use `Function` here in reference to the LabVIEW example.

image

examples/nidcpower_nivisa_dmm_measurement/_helpers.py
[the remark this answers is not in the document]
samchris007: Ok. Updated the helpers to the original examples.
Changes made in helpers
- Renamed APIs `create_visa_resource_manager`, `create_visa_session` to `create_resource_manager`, `create_session` in **_visa_helpers.py**.
- Move `USE_SIMULATION` to measurement.py so as to normalize the simulation between two instruments in both the instrument helpers.

examples/nidcpower_nivisa_dmm_measurement/_helpers.py
```diff
@@ -0,0 +1,306 @@
...
+)
+from ni_measurementlink_service.measurement.service import (
+    GrpcChannelPool,
+    MeasurementService,
+)
+
+USE_SIMULATION = True
+"""
+To use physical instruments, set this to False or specify
+--no-use-simulation on the command line.
```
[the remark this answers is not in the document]
samchris007: Sure. Updated the session creation APIs to use `service_options.use_simulation` as suggested.

examples/nidcpower_nivisa_dmm_measurement/measurement.py
```diff
@@ -0,0 +1,242 @@
...
+        )
+        source_session_info = _get_session_info_from_pin(reservation.session_info, input_pin)
+        measure_session_info = _get_session_info_from_pin(reservation.session_info, output_pin)
+        resource_manager = create_visa_resource_manager(service_options.use_simulation)
+        # Creates NI-DCPower and NI-VISA DMM sessions.
+        with create_session(
+            source_session_info, grpc_device_channel
+        ) as source_session, create_visa_session(
+            resource_manager, measure_session_info.resource_name
+        ) as measure_session:
```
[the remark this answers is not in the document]
samchris007: Oh sure. 
- Renamed APIs `create_visa_resource_manager` and `create_visa_session` to `create_resource_manager` and `create_session` in **_visa_helpers.py**.
- Updated the example using module scoping.

examples/nidcpower_nivisa_dmm_measurement/pyproject.toml
samchris007: As discussed in our PO Connect, I've renamed the measurement to `Output Voltage Measurement`.

Also, I will be creating a separate PR for moving this example into the **multi_instrument** sub-directory with the necessary workflow changes.

examples/nidcpower_nivisa_dmm_measurement/measurement.py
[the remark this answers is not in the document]
samchris007: As discussed in our PO Connect, I have updated the order of measurement operations,
1. Instead of using the trigger, configured both instruments at the start.
2. Initiated the SMU and waited for the source complete.
3. Then, Read the DMM

examples/nidcpower_nivisa_dmm_measurement/measurement.py
```diff
@@ -0,0 +1,242 @@
...
+)
+@measurement_service.configuration("range", nims.DataType.Double, 1.0)
+@measurement_service.configuration("resolution_digits", nims.DataType.Double, 3.5)
+@measurement_service.configuration(
+    "output_pin",
+    nims.DataType.Pin,
+    "OutPin",
+    instrument_type=INSTRUMENT_TYPE_DMM_SIMULATOR,
+)
+@measurement_service.output("measured_value", nims.DataType.Double)
```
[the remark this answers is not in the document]
samchris007: We are aligned that there is no or less advantage in triggering the DMM as both are different instruments and are not connected through the same PXI. 
> If you are trying to minimize test time, then you should configure both instruments up front, then initiate the SMU, wait for source complete, and read the DMM. This way, the SMU's [source delay property](https://www.ni.com/docs/en-US/bundle/ni-dcpower-properties/page/dcpowerpropref/pnidcpower_sourcedelay.html) more accurately controls the time between the source and measurement operations.

As suggested, I have updated the order of the measurement operations by configuring both instruments at the start, initiating the SMU and waiting for the source complete, and then reading the DMM.

So please share your thoughts on whether we still need to return the SMU's InPin measurements.

examples/nidcpower_nivisa_dmm_measurement/measurement.py
samchris007: `session_info` will only be empty if the existing code gets edited.
Hence, Not handling empty lists should be okay in this context.

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
```diff
@@ -7,21 +7,16 @@
...
-USE_SIMULATION = True
-"""
-To use a physical NI SMU instrument, set this to False or specify
---no-use-simulation on the command line.
-"""
-
 
 def create_session(
     session_info: nims.session_management.SessionInformation,
+    use_simulation: bool,
```
bkeryan, line 13: `teststand_fixture.py` is still using the old signature.

https://github.com/ni/measurementlink-python/blob/main/examples/nidcpower_source_dc_voltage/teststand_fixture.py#L63
```
                _ = create_session(
                    session_info,
                    grpc_device_channel,
                    initialization_behavior=nidcpower.SessionInitializationBehavior.INITIALIZE_SERVER_SESSION,
                )
```

examples/nivisa_dmm_measurement/teststand_fixture.py
```diff
@@ -37,13 +36,16 @@ def update_pin_map(pin_map_path: str, sequence_context: Any) -> None:
     teststand_support.set_active_pin_map_id(pin_map_id)
 
 
-def create_nivisa_dmm_sessions(sequence_context: Any) -> None:
+def create_nivisa_dmm_sessions(sequence_context: Any, use_simulation: bool) -> None:
```
bkeryan, line 39: If we are going to move `use_simulation` to the TestStand file globals:
- Do it for all examples in a separate PR
- Update README.md

bkeryan, line 39: @rfriedma I think you originally suggested the `USE_SIMULATION` constant. Is moving it to the TestStand file globals okay with you?

examples/output_voltage_measurement/NIInstrumentSimulatorV2_0.yaml
bkeryan, line 1: The path we discussed in our conference call on Thursday was `examples/multi_instrument/output_voltage_measurement/`

examples/output_voltage_measurement/README.md
```diff
@@ -0,0 +1,44 @@
+## Output Voltage Measurement
+
+This is a MeasurementLink example that sources DC voltage as input with an NI SMU and measures output using NI-VISA DMM.
```
bkeryan, line 3: as input to what?
measures output of what?

Use the word "DUT".

examples/output_voltage_measurement/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Features
+
+- Uses the `nidcpower` and `PyVISA` packages to access NI-DCPower and PyVISA from Python
```
bkeryan, line 7: don't combine these

examples/output_voltage_measurement/README.md
```diff
@@ -0,0 +1,44 @@
...
+- Uses the open-source `PyVISA-sim` package to simulate instruments in software
+- Pin-aware, supporting one session and pin per instrument
+  - Sources the DC voltage level on the selected pin/site
+  - Measures output voltage in a single output pin/site
```
bkeryan, line 11:
```suggestion
- Pin-aware, supporting one session and pin per instrument and a single site
  - Sources the DC voltage level of the DUT input pin
  - Measures the voltage of the DUT output pin
```

examples/output_voltage_measurement/README.md
bkeryan, line 3: "NI-VISA DMM" may be useful shorthand but it's imprecise at best and misleading at worst. That's why the other readme doesn't use this term.

examples/output_voltage_measurement/README.md
```diff
@@ -0,0 +1,44 @@
...
+Note: there is no Python instrument driver for the NI Instrument Simulator, so
+this example directly performs low-level, device-specific commands and queries.
+
+### Required Hardware
+
+By default, this example does not require hardware; it uses a simulated instrument and PyVISA-sim to simulate NI-DCPower and NI-VISA DMM instruments in software. [`NIInstrumentSimulatorV2_0.yaml`](./NIInstrumentSimulatorV2_0.yaml) defines the behavior of the simulated NI-VISA DMM instrument.
+
+This example requires an NI SMU that is supported by NI-DCPower (e.g. PXIe-4141).
+
+To use physical instruments, edit `measurement.py` to specify `USE_SIMULATION = False`
```
bkeryan, line 31: but in TestStand...

examples/nivisa_dmm_measurement/measurement.py
```diff
@@ -92,10 +97,8 @@ def measure(
         instrument_type_id=INSTRUMENT_TYPE_DMM_SIMULATOR,
         timeout=RESERVATION_TIMEOUT_IN_SECONDS,
     ) as reservation:
-        resource_manager = create_visa_resource_manager(service_options.use_simulation)
-        with create_visa_session(
-            resource_manager, reservation.session_info.resource_name
-        ) as session:
+        resource_manager = create_resource_manager(service_options.use_simulation)
+        with create_session(resource_manager, reservation.session_info.resource_name) as session:
```
bkeryan, line 101: FYI, we could remove another call by making `create_session`'s `resource_manager` parameter optional. However, I guess we would have to pass `use_simulation` instead.

examples/output_voltage_measurement/teststand_nivisadmm.py
bkeryan, line 1: `teststand_nivisa_dmm.py`

examples/output_voltage_measurement/README.md
bkeryan, line 3: I thought we were going to add content about how this example also works with a real DMM, like a 34401A

examples/output_voltage_measurement/pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[tool.poetry]
+name = "output_voltage_measurement"
+version = "0.5.0"
+description = ""
+authors = ["National Instruments"]
+
+[tool.poetry.dependencies]
+python = "^3.8"
+ni-measurementlink-service = {version = "^1.1.0-dev2", allow-prereleases = true}
```
bkeryan, line 9:
```suggestion
ni-measurementlink-service = {version = "^1.1.0"}
```

examples/output_voltage_measurement/teststand_pin_map.py
bkeryan, line 1: Are you going to do this for all the examples? in a separate PR?

bkeryan (changes requested):

examples/output_voltage_measurement/NIInstrumentSimulatorV2_0.yaml
samchris007: Yes. That's right. But we must also need to update the GitHub workflows(**check_examples.yml** and **install_examples.py**). Hence, the moving of `output_voltage_measurement` into the `multi_instrument` sub-directory will be raised as a separate PR.

examples/nivisa_dmm_measurement/teststand_fixture.py
rfriedma: I lean yes however, if a user sees "use_simulation" as a global on the sequence... I think it'd be expected that it would simulate everything. Would that be the case?

rfriedma: I agree with that approach for all examples that support simulation. Add a new File Global as sibling to the MeasurementLink container (not within it).

examples/nidcpower_source_dc_voltage/_nidcpower_helpers.py
Jayaseelan-James: I have reverted the changes specific to `nidcpower` and `nivisa_dmm` measurement. Will take this up as part of work item 2473562.

examples/nivisa_dmm_measurement/teststand_fixture.py
Jayaseelan-James: Resolving the conversation now, but the change will be taken up as part of work item 2473562.

examples/output_voltage_measurement/README.md
Jayaseelan-James: Updated to `... measures the DUT output with DMM using NI-VISA.`.

examples/output_voltage_measurement/README.md
Jayaseelan-James: Added individual points explaining the edit flow for both MeasurementLink UI Editor / InstrumentStudio and TestStand.

examples/nivisa_dmm_measurement/measurement.py
Jayaseelan-James: Updated this change for `output_voltage_measurement` , whereas updates to the `nivisa_dmm_measurement` will be taken up as part of work item 2473562.

examples/output_voltage_measurement/README.md
Jayaseelan-James: Added the below points to emphasize the use of real DMM:

To use a DMM like Keysight 34401A:
  - Update the [`OutputVoltageMeasurement.pinmap`](./OutputVoltageMeasurement.pinmap) to reflect the connected instrument's type and resource name.
  - Additionally, verify driver specific commands in `_visa_helpers.py` and update them when required.

examples/nivisa_dmm_measurement/teststand_fixture.py
Jayaseelan-James: @rfriedma, @bkeryan Now, I have added a new `_constants.py` file under `output_voltage_measurement` directory which holds the `USE_SIMULATION` boolean and is imported in `measurement.py`, `teststand_nidcpower.py` and `teststand_nivisa_dmm.py` and passed to the `create_session` method. 

Pros of this approach:
1. The existing user flow of updating the USE_SIMULATION bool in a .py is retained, just that now the user has to update the instrument mode in `_constants.py` instead of `_<instrument>_helpers.py`.
2. A `UseSimulation` boolean in teststand sequence is not required as it is handled using the same boolean in `_constants.py`. So, the user will update only one file to switch between hardware and simulation instead of updating in .py and in TestStand.

Cons:
1. A new `_constants.py` file is added.

examples/output_voltage_measurement/README.md
```diff
@@ -0,0 +1,58 @@
+## Output Voltage Measurement
+
+This is a MeasurementLink example that sources DC voltage as input to the DUT with an NI SMU and measures the DUT output with DMM using NI-VISA.
+
+### Features
+
+- Uses the `nidcpower` package to access NI-DCPower from Python
+- Uses the open-source `PyVISA` package to access PyVISA from Python
```
bkeryan, line 8: This does not match https://github.com/ni/measurementlink-python/blob/main/examples/nivisa_dmm_measurement/README.md
```suggestion
- Uses the open-source `PyVISA` package to access NI-VISA from Python
```

examples/output_voltage_measurement/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Required Hardware
+
+By default, this example does not require hardware; it uses a simulated instrument and PyVISA-sim to simulate NI-DCPower and NI-VISA DMM instruments in software. [`NIInstrumentSimulatorV2_0.yaml`](./NIInstrumentSimulatorV2_0.yaml) defines the behavior of the simulated NI-VISA DMM instrument.
+
+This example requires an NI SMU that is supported by NI-DCPower (e.g. PXIe-4141).
+
+To use physical instruments: 
+- Edit `measurement.py` to specify `USE_SIMULATION = False` when working with MeasurementLink UI Editor / InstrumentStudio.
+- Edit `OutputVoltageMeasurement.seq` to specify `FileGlobals.UseSimulation` value to `False` when working with TestStand.
```
bkeryan, line 41: This is out of date

examples/output_voltage_measurement/README.md
```diff
@@ -0,0 +1,58 @@
...
+  `GPIB0::3::INSTR`, which matches the NI Instrument Simulator's factory default
+  settings when connected via GPIB.
+  - If this doesn't match your configuration, edit
+    [`OutputVoltageMeasurement.pinmap`](./OutputVoltageMeasurement.pinmap) and replace
+    `GPIB0::3::INSTR` with the desired resource name (e.g. `ASRL1::INSTR`).
+  - To modify the NI Instrument Simulator configuration (e.g. GPIB address,
+    serial configuration), use the `Instrument Simulator Wizard` included with
+    the NI Instrument Simulator software.
+
+To use a DMM like Keysight 34401A:
```
bkeryan, line 56:
```suggestion
To use a 3rd party DMM, such as a Keysight 34401A:
```

examples/output_voltage_measurement/README.md
```diff
@@ -0,0 +1,58 @@
...
+  - Update the [`OutputVoltageMeasurement.pinmap`](./OutputVoltageMeasurement.pinmap) to reflect the connected instrument's type and resource name.
+  - Additionally, verify driver specific commands in `_visa_helpers.py` and update them when required.
```
bkeryan, line 58:
```suggestion
  - Additionally, verify driver specific commands in `_visa_helpers.py` and update them if required.
```

examples/output_voltage_measurement/README.md
bkeryan, line 56: Add this paragraph to the other VISA example's readme.

examples/output_voltage_measurement/README.md
```diff
@@ -0,0 +1,58 @@
...
+  settings when connected via GPIB.
+  - If this doesn't match your configuration, edit
+    [`OutputVoltageMeasurement.pinmap`](./OutputVoltageMeasurement.pinmap) and replace
+    `GPIB0::3::INSTR` with the desired resource name (e.g. `ASRL1::INSTR`).
+  - To modify the NI Instrument Simulator configuration (e.g. GPIB address,
+    serial configuration), use the `Instrument Simulator Wizard` included with
+    the NI Instrument Simulator software.
+
+To use a DMM like Keysight 34401A:
+  - Update the [`OutputVoltageMeasurement.pinmap`](./OutputVoltageMeasurement.pinmap) to reflect the connected instrument's type and resource name.
```
bkeryan, line 57: Changing the instrument type in the pinmap requires changing the Python code and the .sfp file, so let's not recommend that here.

examples/output_voltage_measurement/_constants.py
```diff
@@ -0,0 +1,6 @@
+"""Constants."""
+USE_SIMULATION = True
```
bkeryan, line 2: It's weird that ni-python-styleguide didn't make you leave a blank line here.

examples/output_voltage_measurement/measurement.py
```diff
@@ -0,0 +1,237 @@
...
+        context=measurement_service.context.pin_map_context,
+        pin_or_relay_names=[input_pin, output_pin],
+        timeout=RESERVATION_TIMEOUT_IN_SECONDS,
+    ) as reservation:
+        grpc_device_channel = get_grpc_device_channel(
+            measurement_service, nidcpower, service_options
+        )
+        source_session_info = _get_session_info_for_pin(reservation.session_info, input_pin)
+        measure_session_info = _get_session_info_for_pin(reservation.session_info, output_pin)
+        # Creates NI-DCPower and NI-VISA DMM sessions.
```
bkeryan, line 125: This should be obvious from the code.

examples/output_voltage_measurement/measurement.py
```diff
@@ -0,0 +1,237 @@
...
+    session_index = get_session_and_channel_for_pin(session_info, pin_name)[0]
+    return session_info[session_index]
+
+
+def _add_cancel_callback(session, measurement_service, pending_cancellation):
+    def cancel_callback():
+        session_to_abort = session
+        if session_to_abort is not None:
+            nonlocal pending_cancellation
+            pending_cancellation = True
```
bkeryan, line 184: This doesn't work due to Python scoping rules. Unfortunately, this minor refactor seems to have opened another "can of worms" that may lead to changing all of the examples.

The nested function `cancel_callback` can modify the `pending_cancellation` variable captured from the `_add_cancel_callback` function. However, this has no effect on the `measure` function because `bool` in Python behaves as if it is passed by value.

Really, it's more complicated: there are two singleton instances of `bool`: `True` and `False`, and you pass around references to them. However, these instances are immutable, so assigning to a `bool` variable works by rebinding it to a different reference. `_add_cancel_callback` cannot modify `measure`'s `pending_cancellation` variable.

For more details and explanation, read:
- https://realpython.com/python-scope-legb-rule/
- https://stackoverflow.com/questions/1494696/are-pythons-bools-passed-by-value

An alternative is to use `threading.Event` like in https://github.com/ni/measurementlink-python/blob/main/examples/ui_progress_updates/measurement.py . Event objects are mutable, so you can have a callback set the event without rebinding any references. As a plus, you don't even need to define a callback function, because you can use the event's `set` method as the callback.

In addition, calling `abort()` on the NI-DCPower session doesn't make as much sense as we originally thought. It doesn't unblock waiting API calls like it does in NI-DAQmx. Also, raising an exception in the `with session.initiate()` block automatically calls `abort()`, so we don't really need to do it again.

examples/output_voltage_measurement/pyproject.toml
```diff
@@ -0,0 +1,32 @@
...
+PyVISA-sim = "^0.5.1"
+nidcpower = { version = ">=1.4.4", extras = ["grpc"] }
+click = ">=7.1.2"
+grpcio = "*"
+
+[tool.poetry.group.dev.dependencies]
+mypy = ">=1.0"
+grpc-stubs = "^1.53"
+# Uncomment to use prerelease dependencies.
+# nidcpower = { git = "https://github.com/ni/nimi-python.git", subdirectory = "generated/nidcpower"}
```
bkeryan, line 20: sorry, this is wrong in all the examples

```suggestion
# nidcpower = { git = "https://github.com/ni/nimi-python.git", subdirectory = "generated/nidcpower", extras = ["grpc"]}
```

bkeryan (changes requested):

examples/output_voltage_measurement/_constants.py
Jayaseelan-James: Included a blank line now.

examples/output_voltage_measurement/pyproject.toml
Jayaseelan-James: Sure, I updated this example's `pyproject.toml` now. Will update the other examples as a part of work item 2484653.

examples/output_voltage_measurement/README.md
Jayaseelan-James: Will update this as part of work item 2473562.

examples/output_voltage_measurement/measurement.py
Jayaseelan-James: I have updated the `measurement.py` to use `threading.Event` for "cancellation_callback".

examples/output_voltage_measurement/README.md
Jayaseelan-James: Updated.

examples/output_voltage_measurement/measurement.py
```diff
@@ -176,25 +176,14 @@ def _get_session_info_for_pin(session_info, pin_name):
...
-def _wait_for_source_complete_event(measurement_service, channels, pending_cancellation):
+def _wait_for_source_complete_event(measurement_service, channels, cancellation_event):
     deadline = time.time() + measurement_service.context.time_remaining
     while True:
         if time.time() > deadline:
             measurement_service.context.abort(
                 grpc.StatusCode.DEADLINE_EXCEEDED, "deadline exceeded"
             )
-        if pending_cancellation:
+        if cancellation_event.wait(0.1):
```
bkeryan, line 186: The purpose of this function is to wait for the source complete event. Ideally, we would call `channels.wait_for_event()` once with a long timeout, but there is no way to cancel this long wait, so we break it up into shorter 100 ms waits and check for cancellation after each short wait.

Adding a second wait to the loop hurts performance because it will keep waiting for up to 100 ms after the source complete event has occurred.

`channels.wait_for_event()` performs a synchronous wait and doesn't support multiplexed I/O like `select()` or `WaitForMultipleObjects()`, so the best solution is to call `channels._wait_for_event()` with a 100 ms timeout and use the cancellation event as an atomic flag.

```suggestion
        if cancellation_event.is_set():
```

bkeryan (approved):
Approved with suggestions:
- Use `event.is_set()` instead of `event.wait(timeout=0.1)`

examples/output_voltage_measurement/measurement.py
Jayaseelan-James: Thanks for the suggestion. Updated to invoke `is_set()`.

Jayaseelan-James:
@dixonjoel, @csjall Please help review this PR.

dixonjoel (approved):

dixonjoel:
> @dixonjoel, @csjall Please help review this PR.

Done. I don't think you need to wait for @csjall

PR #328, opened 2023-07-12
[US2422815] Protobuf Enum Support

WesleyTangNationalInstruments:

- Adds support for using protobuf enums with MeasurementLink.
- Updated sample_measurement to include an example of protobuf enums

[AB#2422815](https://ni.visualstudio.com/DevCentral/_backlogs/backlog/InstrumentStudio%20Platform/Features/?workitem=2422815)

- Manually verified sample_measurement using protobuf enums runs in InstrumentStudio.
- Manually verified sample_measurement using protobuf enums are linked to the correct controls and indicators in MeasurementUIEditor.
- Verified tests pass.

image
image


examples/sample_measurement/assets/color_pb2.pyi
bkeryan, line 1: Did you generate this by running protoc manually? Please update `scripts/generate_grpc_stubs.py`

ni_measurementlink_service/_internal/parameter/serializer.py
```diff
@@ -224,11 +227,18 @@ def _deserialize_enum_parameters(
         parameter_metadata = parameter_metadata_dict[id]
         if get_enum_values_annotation(parameter_metadata):
             enum_type = _get_enum_type(parameter_metadata)
+            is_protobuf_enum = enum_type is int
```
bkeryan, line 230: Please test `IntEnum` as well. An `IntEnum` is an instance of both `Enum` and `int`.

ni_measurementlink_service/_internal/parameter/serializer.py
```diff
@@ -75,15 +76,17 @@ def serialize_parameters(
...
             == TypeSpecialization.Enum.value
         ):
             if parameter_metadata.repeated:
-                parameter = [x.value for x in parameter]
+                if len(parameter) > 0 and isinstance(parameter[0], Enum):
+                    parameter = [x.value for x in parameter]
             else:
-                parameter = parameter.value
+                if isinstance(parameter, Enum):
+                    parameter = parameter.value
```
bkeryan, line 89: It would be more readable to call a `_get_enum_value` function that does the `isinstance` check because then you don't have to think about what it means to skip assigning to `parameter`.

ni_measurementlink_service/measurement/service.py
```diff
@@ -7,9 +7,10 @@
 from os import path
 from pathlib import Path
 from threading import Lock
-from typing import Any, Callable, Dict, List, Optional, Type, TypeVar
+from typing import Any, Callable, Dict, List, Optional, Type, TypeVar, Union
 
 import grpc
+from google.protobuf.internal.enum_type_wrapper import EnumTypeWrapper
```
bkeryan, line 13: This is internal, so I'm not sure we're allowed to reference it.

If we must import an internal type, we can reduce the risk of breakage by doing the import only when type checking, not at run time.

```
...
import typing
...

if typing.TYPE_CHECKING:
   from google.protobuf.internal.enum_type_wrapper import EnumTypeWrapper
```

ni_measurementlink_service/measurement/service.py
```diff
@@ -370,12 +375,20 @@ def _make_annotations_dict(
 
         return annotations
 
-    def _enum_to_annotations_value(self, enum_type: Type[Enum]) -> str:
-        if not any(member.value == 0 for member in enum_type):
-            raise ValueError("The enum does not have a value for 0.")
+    def _enum_to_annotations_value(
+        self, enum_type: Union[Type[Enum], Type[EnumTypeWrapper]]
```
bkeryan, line 379: Consider defining a type alias instead of duplicating this type.

ni_measurementlink_service/measurement/service.py
```diff
@@ -370,12 +375,20 @@ def _make_annotations_dict(
...
+    ) -> str:
         enum_values = {}
-        for member in enum_type:
-            enum_values[member.name] = member.value
+        if isinstance(enum_type, EnumTypeWrapper):
```
bkeryan, line 382: Should this be `issubclass`? Or is an instance of `EnumTypeWrapper` itself a type?

bkeryan (changes requested):

examples/sample_measurement/assets/color_pb2.pyi
WesleyTangNationalInstruments: Updated `scripts/generate_grpc_stubs.py` to automatically generate these files.

ni_measurementlink_service/_internal/parameter/serializer.py
WesleyTangNationalInstruments: Added tests for IntEnum.

ni_measurementlink_service/_internal/parameter/serializer.py
WesleyTangNationalInstruments: Added `_get_enum_value` helper as suggested.

ni_measurementlink_service/measurement/service.py
WesleyTangNationalInstruments: Added `SupportedEnumType` alias.

ni_measurementlink_service/measurement/service.py
WesleyTangNationalInstruments: The enum_type that gets passed in for protobuf enums appears to be an instance of EnumTypeWrapper in addition to being a type itself which is why we used `isinstance` here. I'm not entirely sure why the protobuf enum types are instances of EnumTypeWrapper rather than just being a subtype of it, but that seems to be the way it behaves. 

In our decorator, the intent was for users to pass in that type directly.

`@sample_measurement_service.configuration("Protobuf Enum In", nims.DataType.Enum, ProtobufColor.BLACK, enum_type=ProtobufColor)`

If instead we wanted to use `issubclass(enum_type, EnumTypeWrapper)` here, we would need the users to pass in `enum_type=type(ProtobufColor))` instead. 

I updated this code to check `issubclass(type(enum_type), EnumTypeWrapper)` and added a comment to make it clearer, but note that `type(enum_type)` is required since `issubclass` throws an error if the parameter is an instance.

bkeryan: I'm not sure whether the enum type wrapper instance is a type or not.

The `.pyi` file says it subclasses `builtins.type`:
https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/_stubs/session_pb2.pyi

But at run time, `isinstance` says it's not a type.
```
>>> from nidaqmx._stubs import nidaqmx_pb2
>>> session_pb2.SessionInitializationBehavior
<google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper object at 0x000001D87531E8E0>
>>> session_pb2.SessionInitializationBehavior.keys()
['SESSION_INITIALIZATION_BEHAVIOR_UNSPECIFIED', 'SESSION_INITIALIZATION_BEHAVIOR_INITIALIZE_NEW', 'SESSION_INITIALIZATION_BEHAVIOR_ATTACH_TO_EXISTING']
>>> type(session_pb2.SessionInitializationBehavior)
<class 'google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper'>
>>> isinstance(session_pb2.SessionInitializationBehavior, type)
False
```

I wonder if `SupportedEnumType = Union[Type[Enum], EnumTypeWrapper]` would also work.

WesleyTangNationalInstruments: Yeah, I saw this too and I'm not entirely sure why it behaves this way. Despite the `type()` of it being a class, issubclass does not believe it to be so. 

image

Note that `SupportedEnumType = Union[Type[Enum], EnumTypeWrapper]` does also work. Let me know if you prefer this to `Union[Type[Enum], Type[EnumTypeWrapper]]` and I can make the change, but I've left it as-is for now since I'm not sure it really makes it any clearer.

examples/sample_measurement/measurement.py
```diff
@@ -4,6 +4,7 @@
 from enum import Enum
 
 import click
+from examples.sample_measurement.assets.color_pb2 import ProtobufColor
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Hmm, I'm not sure how to best resolve this without trying to separate the acceptance tests from using sample_measurement service. 

I thought we could include the assets in the examples.zip and then reference `assets.color_pb2` but I don't think that works if we don't make assets an module because the tests break. Perhaps we can create a module that contains just `colors.proto` and associated files that sample_measurement and tests can use, but that seems a bit excessive. We could remove the protobuf enum from sample_measurement and implement test coverage at a lower level as an alternative.

Note that importing `examples.sample_measurement.assets.color_pb2` does work when running the example directly and with running tests on my machine, but I suppose that's only because I'm running it out of the repo where the assets files can be found in the examples directory.

ni_measurementlink_service/measurement/service.py
WesleyTangNationalInstruments: Let me know if I'm misinterpreting this situation, but only importing during type checking doesn't seem to work as we use `EnumTypeWrapper` for more than just `SupportedEnumType`. We reference `EnumTypeWrapper` as part of generating the annotations for the enum (in `MeasurementService._enum_to_annotations_value`) when we resolve the configuration and output decorators. I tried to find a public alternative to `EnumTypeWrapper` that we could use instead to effectively make the same check but couldn't find anything that panned out. I've left this as-is for now.

examples/sample_measurement/measurement.py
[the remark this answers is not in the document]
bkeryan: Let's do a conditional import for now and consider separating the examples from the tests later.

```
try:
    from .assets import color_pb2
except ImportError:
    from examples.sample_measurement.assets import color_pb2
```

ni_measurementlink_service/measurement/service.py
```diff
@@ -377,12 +384,21 @@ def _make_annotations_dict(
...
-        for member in enum_type:
-            enum_values[member.name] = member.value
+        # Note that the type of protobuf enums are an instance of EnumTypeWrapper
+        # in addition to being a type itself. Additionally, issubclass excludes
+        # instances as valid parameters so we use isinstance here.
+        if isinstance(enum_type, EnumTypeWrapper):
+            if not any(value.number == 0 for value in enum_type.DESCRIPTOR.values):
+                raise ValueError("The enum does not have a value for 0.")
+            for value in enum_type.DESCRIPTOR.values:
+                enum_values[value.name] = value.number
```
bkeryan, line 396: `EnumTypeWrapper` provides `dict`-style methods that are more concise than accessing the `EnumDescriptor` directly.

```
            if 0 not in enum_type.values():
                raise ValueError("The enum does not have a value for 0.")
            for name, value in enum_type.items():
                enum_values[name] = value
```

ni_measurementlink_service/measurement/service.py
bkeryan: Yes, I prefer treating it as an instance of `EnumTypeWrapper` because that is true for both run time and type checking.

Also, please remove "in addition to being a type itself" from the comment because this is only true for type checking, not run time.

If you needed to change the order of the if-statement cases, I think `if isinstance(enum_type, type) and issubclass(enum_type, Enum)` would prevent it from throwing.

tests/unit/test_metadata.py
```diff
@@ -25,6 +25,15 @@ class DifferentColor(Enum):
     BROWN = 3
 
 
+class Countries(IntEnum):
+    """Countries enum used for testing enum-typed config and output."""
```
bkeryan, line 29: Please call out that this tests `IntEnum`

examples/sample_measurement/pyproject.toml
```diff
@@ -14,6 +14,7 @@ mypy = ">=1.0"
 grpc-stubs = "^1.53"
 # Uncomment to use prerelease dependencies.
 # ni-measurementlink-service = {path = "../..", develop = true}
+types-protobuf = "^4.21"
```
bkeryan, line 17: [style] Let's keep the type stub modules together

bkeryan (changes requested):

examples/sample_measurement/pyproject.toml
WesleyTangNationalInstruments: Moved above the commented out dependencies as suggested.

tests/unit/test_metadata.py
WesleyTangNationalInstruments: Updated comment as suggested.

ni_measurementlink_service/_internal/parameter/metadata.py
```diff
@@ -161,5 +161,8 @@ def get_enum_values_annotation(parameter_metadata: ParameterMetadata) -> str:
 
 def _is_valid_enum_value(enum_value: object, user_enum: Dict[str, int]) -> bool:
     if not isinstance(enum_value, Enum):
-        return False
+        if isinstance(enum_value, int):
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Simplified as suggested.

ni_measurementlink_service/measurement/service.py
WesleyTangNationalInstruments: Refactored as suggested. Note that I also had to add a cast here to resolve mypy issue complaining about `Union[Type[Enum], EnumTypeWrapper]` not having `values` or `items` because we swapped the `isinstance` check to not reference `EnumTypeWrapper` directly anymore.

ni_measurementlink_service/measurement/service.py
WesleyTangNationalInstruments: Updated to import EnumTypeWrapper only on type checking and swapped the runtime `isinstance` usage via the EnumDescriptor instead. Thanks for the suggestion here.

ni_measurementlink_service/measurement/service.py
WesleyTangNationalInstruments: Switched to `Union[Type[Enum], EnumTypeWrapper]` as suggested and fixed comment. Fixed condition as suggested to fix mypy error as suggested.

examples/sample_measurement/measurement.py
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Switched to using a conditional import as suggested.

examples/sample_measurement/assets/color.proto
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Added `__init__.py` and renamed directory as suggested.

examples/sample_measurement/measurement.py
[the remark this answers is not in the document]
WesleyTangNationalInstruments: @bkeryan I'm hitting some mypy issues when it runs on this PR and I'm not sure how best to resolve them. 

These do not appear when I run locally via `poetry run mypy -p examples`. I tried directly specifying `sample_measurement` via `poetry run mypy .\examples\sample_measurement\` (which is how it looks like the PR checks are running this check). I do get a "Source file found twice under different module names", but I get none of the same errors from this PR's example check. I'm not sure mypy that's being run on this PR likes the conditional import workaround or respecting the `TYPE_CHECK` condition to import `EnumTypeWrapper`. 

Doing some investigating on stackoverflow, I suspect that the PR's mypy is configured differently than what I have locally, but I'm unsure fighting with mypy into ignoring these errors is the right approach. I tried making sample_measurement's a module by adding a `__init__.py` for that directory as well and changing the imports to reference differently without much success. Please let me know if you have any ideas on how to get the examples check passing.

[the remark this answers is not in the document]
bkeryan: These problems are why the GitHub "check examples" works the way it does: https://github.com/ni/measurementlink-python/blob/main/.github/workflows/check_examples.yml

You need to run `poetry install` and `poetry run mypy .` separately for each example.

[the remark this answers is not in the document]
bkeryan: https://github.com/ni/measurementlink-python/actions/runs/5557275444/jobs/10150844869?pr=328

```
  measurement.py:11: error: Cannot find implementation or library stub for module named "examples.sample_measurement._stubs"  [import]
  measurement.py:11: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
  measurement.py:11: error: Name "color_pb2" already defined (by an import)  [no-redef]
```
The first error can be avoided by setting `[[tool.mypy.overrides]]` appropriately: https://github.com/ni/measurementlink-python/blob/main/examples/nidmm_measurement/pyproject.toml#L25

But the second error probably needs `# type: ignore`. 

I think I saw the no-redef error here too but never got around to posting a PR for the fix: https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/__init__.py

```
  measurement.py:[50](https://github.com/ni/measurementlink-python/actions/runs/5557275444/jobs/10150844869?pr=328#step:7:59): error: Argument "enum_type" to "configuration" of "MeasurementService" has incompatible type "type[ProtobufColor]"; expected "Optional[type[Enum]]"  [arg-type]
  measurement.py:61: error: Argument "enum_type" to "output" of "MeasurementService" has incompatible type "type[ProtobufColor]"; expected "Optional[type[Enum]]"  [arg-type]
  Found 4 errors in 1 file (checked 4 source files)
```

This is because the example is using the released `ni-measurementlink-service` 1.1.0 which does not include your changes. You need to uncomment the ../.. stuff again.

ni_measurementlink_service/measurement/service.py
bkeryan: You can remove the `if TYPE_CHECKING` around the cast by passing the type name as a string :)

image

Typeshed confirms that `cast` has an overload that takes `typ: str`.
https://github.com/python/typeshed/blob/main/stdlib/typing.pyi#L825

WesleyTangNationalInstruments: Swapped to casting by name rather than type as suggested.

examples/sample_measurement/measurement.py
bkeryan, line 1: In isolation, this PR looks good to go. However, https://github.com/ni/measurementlink-python/pull/329 is decoupling the tests from the examples, so I think you should undo your example changes and reapply them to the new `fake_measurement_service` test utility.

bkeryan (changes requested):

bkeryan (approved):

examples/sample_measurement/measurement.py
WesleyTangNationalInstruments: @bkeryan I think having the protobuf enum in the sample_measurement example brings value to the end user that it is worth it to keep it in. We wanted to ensure that we have examples that covered all of the enum use cases anyways so protobuf enum in the examples would have been something we wanted in anyways. I'd be in favor of checking in this PR now and then either me or @vigkre can update `fake_measurement_service` afterwards if you are fine with that.

bkeryan: Ok, then I approve of merging it as-is. Should any other reviewers approve? @dixonjoel @synesthetize

examples/sample_measurement/pyproject.toml
```diff
@@ -12,9 +12,16 @@ click = ">=7.1.2, !=8.1.4" # mypy fails with click 8.1.4: https://github.com/pal
 [tool.poetry.group.dev.dependencies]
 mypy = ">=1.0"
 grpc-stubs = "^1.53"
+types-protobuf = "^4.21"
 # Uncomment to use prerelease dependencies.
-# ni-measurementlink-service = {path = "../..", develop = true}
+ni-measurementlink-service = {path = "../..", develop = true}
```
dixonjoel, line 17: This should be commented out when you check it in.

examples/sample_measurement/pyproject.toml
```diff
@@ -12,9 +12,16 @@ click = ">=7.1.2, !=8.1.4" # mypy fails with click 8.1.4: https://github.com/pal
...
 
 [build-system]
 requires = ["poetry-core>=1.2.0"]
 build-backend = "poetry.core.masonry.api"
+
+[[tool.mypy.overrides]]
+module = [
+    "ni-measurementlink-service.*",
+]
+ignore_missing_imports = true
```
dixonjoel, line 27: Why was this necessary?

dixonjoel: The ignore_missing_imports, that is.

dixonjoel (approved):

examples/sample_measurement/pyproject.toml
WesleyTangNationalInstruments: Fixed as suggested.

examples/sample_measurement/pyproject.toml
WesleyTangNationalInstruments: I had copied this from the other examples' `pyproject.toml` and thought it is intended to solve the missing imports mypy issue, but everything still seems to work after I got rid of it so I've removed it.

bkeryan: `ni-measurementlink-service` should not be missing.

examples/sample_measurement/measurement.py
```diff
@@ -38,6 +43,12 @@ class Color(Enum):
 @sample_measurement_service.configuration(
     "Enum In", nims.DataType.Enum, Color.BLUE, enum_type=Color
 )
+@sample_measurement_service.configuration(
+    "Protobuf Enum In",
+    nims.DataType.Enum,
+    color_pb2.ProtobufColor.BLACK,
+    enum_type=color_pb2.ProtobufColor,  # type: ignore
```
bkeryan, line 50: `# type: ignore` should not be necessary here unless you are testing with the previous release of ni-measurementlink-service, in which case protobuf enums don't work at all.

bkeryan (changes requested):

examples/sample_measurement/pyproject.toml
bkeryan: I previously asked Wesley to uncomment this so the example will use his update code. Otherwise, it uses ni-measurementlink-service 1.1.0, which doesn't support protobuf enums.

examples/sample_measurement/pyproject.toml
bkeryan, line 1: Congrats, this example now requires version 1.2

```
ni-measurementlink-service = {version = "^1.2.0-dev0", allow-prereleases = true}
...
ni-measurementlink-service = {path = "../..", develop = true}
```

bkeryan (changes requested):

examples/sample_measurement/pyproject.toml
dixonjoel: Sorry

WesleyTangNationalInstruments: Uncommented again so that the checked in code uses prerelease dependencies. I wasn't entirely sure if this was something we did locally or if we actually wanted this change in the repo.

examples/sample_measurement/pyproject.toml
WesleyTangNationalInstruments: Updated to use 1.2.

bkeryan (approved):

examples/sample_measurement/pyproject.toml
bkeryan: I want this change in the repo until we build a prerelease that has protobuf enum support, then we can comment it back out.

We need to figure out a better workflow for this. :(

examples/sample_measurement/pyproject.toml
```diff
@@ -6,15 +6,21 @@ authors = ["National Instruments"]
...
+ni-measurementlink-service = {path = "../..", develop = true}
 
 [build-system]
 requires = ["poetry-core>=1.2.0"]
 build-backend = "poetry.core.masonry.api"
+
+[[tool.mypy.overrides]]
+module = [
+    "ni-measurementlink-service.*",
+]
```
bkeryan, line 26: This has no configuration keys so delete it. The module key identifies which module the other keys apply to.
https://mypy.readthedocs.io/en/stable/config_file.html#using-a-pyproject-toml-file

examples/sample_measurement/measurement.py
bkeryan: Ok, I see what's going on.

The runtime type is `google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper`, but the type used in `color_pb2.pyi` is `google.protobuf.internal.enum_type_wrapper._EnumTypeWrapper[_V]`:
```
class _ProtobufColor:
    ValueType = typing.NewType("ValueType", builtins.int)
    V: typing_extensions.TypeAlias = ValueType

class _ProtobufColorEnumTypeWrapper(google.protobuf.internal.enum_type_wrapper._EnumTypeWrapper[_ProtobufColor.ValueType], builtins.type):
    DESCRIPTOR: google.protobuf.descriptor.EnumDescriptor
    NONE: _ProtobufColor.ValueType  # 0
    PINK: _ProtobufColor.ValueType  # 1
    WHITE: _ProtobufColor.ValueType  # 2
    BLACK: _ProtobufColor.ValueType  # 3

class ProtobufColor(_ProtobufColor, metaclass=_ProtobufColorEnumTypeWrapper):
    """---------------------------------------------------------------------
    ---------------------------------------------------------------------
    """
```

In the type stubs, `EnumTypeWrapper` inherits from `_EnumTypeWrapper[int]`. `_EnumTypeWrapper[_V]` does not inherit from `EnumTypeWrapper`.
https://github.com/python/typeshed/blob/main/stubs/protobuf/google/protobuf/internal/enum_type_wrapper.pyi
```
_V = TypeVar("_V", bound=int)

# Expose a generic version so that those using mypy-protobuf
# can get autogenerated NewType wrapper around the int values
class _EnumTypeWrapper(Generic[_V]):
    DESCRIPTOR: EnumDescriptor
    def __init__(self, enum_type: EnumDescriptor) -> None: ...
    def Name(self, number: _V) -> str: ...
    def Value(self, name: str | bytes) -> _V: ...
    def keys(self) -> list[str]: ...
    def values(self) -> list[_V]: ...
    def items(self) -> list[tuple[str, _V]]: ...

class EnumTypeWrapper(_EnumTypeWrapper[int]): ...
```

This seems to work:
```
diff --git a/ni_measurementlink_service/measurement/service.py b/ni_measurementlink_service/measurement/service.py
index 3e70dc4..7ba372a 100644
--- a/ni_measurementlink_service/measurement/service.py
+++ b/ni_measurementlink_service/measurement/service.py
@@ -40,9 +40,9 @@ from ni_measurementlink_service.measurement.info import (
 from ni_measurementlink_service.session_management import PinMapContext

 if TYPE_CHECKING:
-    from google.protobuf.internal.enum_type_wrapper import EnumTypeWrapper
+    from google.protobuf.internal.enum_type_wrapper import EnumTypeWrapper, _EnumTypeWrapper, _V

-    SupportedEnumType = Union[Type[Enum], EnumTypeWrapper]
+    SupportedEnumType = Union[Type[Enum], EnumTypeWrapper, _EnumTypeWrapper[_V]]

```

bkeryan: Correction: I think this is sufficient. I don't think we need `EnumTypeWrapper` and `_V`.
```
SupportedEnumType = Union[Type[Enum], _EnumTypeWrapper]
```

examples/sample_measurement/pyproject.toml
WesleyTangNationalInstruments: Removed as suggested.

examples/sample_measurement/measurement.py
WesleyTangNationalInstruments: Switched to `SupportedEnumType = Union[Type[Enum], _EnumTypeWrapper]` as suggested. Thanks for all the help on these issues.

examples/sample_measurement/pyproject.toml
bkeryan: It should have ignored missing imports for `examples.*`

PR #338, opened 2023-07-19
Fix double quotes on Annotations

gpachecoNI:

- Removes the double quotes (") from the annotations whenever they are a string. 

The problem happened because - our Annotations are not all strings we allowed arrays for  Tags and wanted to allow customers to add dictionaries too.  I am using "json.dumps" to help manage that, but the problem is that it also adds quotes to the strings. I tried multiple fixes but this was the simplest I can come up. 

We do not want to display the description or the Title with double quotes. 

Fix and reran tests. 
Manually look at the example: 

image
image

### DIFF
--- a/ni_measurementlink_service/measurement/service.py
+++ b/ni_measurementlink_service/measurement/service.py
@@ -181,8 +181,13 @@ class MeasurementService:
             ui_file_paths=ui_file_paths,
         )
 
+        def removeDoubleQuote(string: str):
+            if string and string[0] == string[-1] == '"':
+                    string = string[1:-1]
+            return string
+
         service_annotations_string = {
-            key: json.dumps(value, separators=(",", ":"))
+            key: removeDoubleQuote(json.dumps(value, separators=(",", ":")))
             for key, value in service.get("annotations", {}).items()
         }
 
--- a/ni_measurementlink_service/measurement/service.py
+++ b/ni_measurementlink_service/measurement/service.py
@@ -181,13 +181,14 @@ class MeasurementService:
             ui_file_paths=ui_file_paths,
         )
 
-        def process_annotations(value: object):
+        def convert_value_to_str(value: object) -> str:
             if isinstance(value, str):
                 return value
             return json.dumps(value, separators=(",", ":"))
 
         service_annotations_string = {
-            key: process_annotations(value) for key, value in service.get("annotations", {}).items()
+            key: convert_value_to_str(value)
+            for key, value in service.get("annotations", {}).items()
         }
 
         self.service_info = ServiceInfo(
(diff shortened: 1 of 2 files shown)


ni_measurementlink_service/measurement/service.py
bkeryan, line 1: Double quotes are not the only consequence of encoding a string as JSON when you shouldn't. Try encoding some accented characters or emoji; they will be double-escaped.

I think the correct solution is to be more selective about which values to encode as JSON. Can we choose based on the value datatype? Or perhaps the key name?

bkeryan (changes requested):

ni_measurementlink_service/measurement/service.py
gpachecoNI: Yes, I agree. I initially tried removing the encoding but could not make it work for some of my tests. I guess I did not try based on the data type, based on Key name is not possible because the customer is able to add any name as the key.  Let me try based on datatype.

gpachecoNI: Fixed

tests/unit/test_service.py
```diff
@@ -261,8 +261,8 @@ def _fake_measurement_function():
                 "ni.measurementlink.measurement.v2.MeasurementService",
             ],
             {
-                "ni/service.description": '"Measure inrush current with a shorted load and validate results against configured limits."',
-                "ni/service.collection": '"CurrentTests.Inrush"',
+                "ni/service.description": "Measure inrush current with a shorted load and validate results against configured limits.",
+                "ni/service.collection": "CurrentTests.Inrush",
                 "ni/service.tags": '["powerup","current"]',
```
cumitche, line 266: It would be good to have a test where the tags were an actual array of strings rather than an encoded JSON string.  I expect most users will populate their tags this way instead of escaping double-quotes in their serviceconfig file.

cumitche (approved):

ni_measurementlink_service/measurement/service.py
```diff
@@ -181,9 +181,13 @@ def __init__(
             ui_file_paths=ui_file_paths,
         )
 
+        def process_annotations(value: object):
```
bkeryan, line 184: This function doesn't process annotations. It processes the value part of one annotation.

"convert" or "escape" would be more descriptive than "process".

Suggestion: "convert_value"

bkeryan (approved):
Approved with suggestions

ni_measurementlink_service/measurement/service.py
bkeryan, line 184: type hint
```suggestion
        def process_annotations(value: object) -> str:
```

PR #342, opened 2023-07-25
Add xydata.proto and generate stubs

dixonjoel:

Generate stubs and add the new xydata.proto from its original source.

We'll be adding the ability to source XY data in the Python measurement service soon, so we need stubs to interact with it.

None. Depending on the PR checks.


ni_measurementlink_service/_internal/stubs/ni/protobuf/__init__.py
bkeryan, line 1: Adding more internal stubs is ok on a temporary basis, but if we expect clients to use these types, we should distribute them in a separate, shared package.

bkeryan (approved):

ni_measurementlink_service/_internal/stubs/ni/protobuf/__init__.py
dixonjoel: Ok, I'm going to submit this for now. I'm not sure whether we will expose the grpc types directly to clients, but if we end up doing so and the measurement.py, for example, needs to access and construct these, we can create a separate package at that time.

PR #343, opened 2023-07-27
discovery_client: When a python measurement starts discovery service, hide discovery service logs

vigkre:

- This PR redirects the logs to /dev/null (hide logs) when a python measurement starts discovery service.
- Before:

image
- After:

image

- Handle `FileNotFoundError` when wrong discovery service .exe path is specified for subprocess.popen function.
- Add tests to register service function to handle `FileNotFoundError` exception.

- Implements [Task 2462367](https://dev.azure.com/ni/DevCentral/_workitems/edit/2462367): When a python measurement starts discovery service, it should not show logs from discovery service to the user.

- Verified discovery service logs are not displayed.
- While the discovery service logs are hidden, it might hide other errors that might arise while starting discovery service through python measurement. To test this case:
        -  Specified wrong discovery service .exe path to `subprocess.popen` function and noticed the log error is captured in the console: 
        
image

### DIFF
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -162,6 +162,7 @@ class DiscoveryClient:
                 _logger.info("Successfully unregistered with discovery service.")
             else:
                 _logger.info("Not registered with discovery service.")
+                return False
         except grpc.RpcError as e:
             if e.code() == grpc.StatusCode.UNAVAILABLE:
                 _logger.error(
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -274,7 +274,7 @@ def _start_service(exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path)
                 raise TimeoutError("Timed out waiting for discovery service to start")
             time.sleep(_START_SERVICE_POLLING_INTERVAL)
     except FileNotFoundError:
-       raise Exception
+        raise Exception
 
 
 def _service_already_running(key_file_path: pathlib.Path) -> bool:
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -135,6 +135,7 @@ class DiscoveryClient:
             _logger.error(
                 "Unable to register with discovery service. Possible reason: discovery service not running."
             )
+            return False
         except Exception:
             _logger.exception("Error in registering with discovery service.")
             return False
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -254,19 +254,27 @@ def _key_file_exists(key_file_path: pathlib.Path) -> bool:
 
 def _start_service(exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path) -> None:
     """Starts the service at the specified path and wait for the service to get up and running."""
-    subprocess.Popen([exe_file_path], cwd=exe_file_path.parent)
-    # After the execution of process, check for key file existence in the path
-    # stop checking after 30 seconds have elapsed and throw error
-    timeout_time = time.time() + _START_SERVICE_TIMEOUT
-    while True:
-        try:
-            with _open_key_file(str(key_file_path)) as _:
-                return
-        except IOError:
-            pass
-        if time.time() >= timeout_time:
-            raise TimeoutError("Timed out waiting for discovery service to start")
-        time.sleep(_START_SERVICE_POLLING_INTERVAL)
+    try:
+        subprocess.Popen(
+            [exe_file_path],
+            cwd=exe_file_path.parent,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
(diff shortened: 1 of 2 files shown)


ni_measurementlink_service/_internal/discovery_client.py
```diff
@@ -254,19 +254,27 @@ def _key_file_exists(key_file_path: pathlib.Path) -> bool:
...
+        while True:
+            try:
+                with _open_key_file(str(key_file_path)) as _:
+                    return
+            except IOError:
+                pass
+            if time.time() >= timeout_time:
+                raise TimeoutError("Timed out waiting for discovery service to start")
+            time.sleep(_START_SERVICE_POLLING_INTERVAL)
+    except FileNotFoundError:
```
mshafer-NI, line 276: why is the try/except needed?

vigkre: https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/_internal/discovery_client.py#L87

In the above function, exception is handled in such a way that when `FileNotFoundError` happens, `True` is returned which denotes that measurement service is registered successfully.

image

But the real scenario is that, when there is `FileNotFoundError` happens from `subprocess.popen`, measurement service is not registered successfully.

bkeryan: You cannot solve every problem by writing more code. Please think about why the `register_measurement_service` method catches `FileNotFoundError` in the first place and update the existing code.

Also, replacing a descriptive exception with an empty `Exception` seems like a bad practice.

bkeryan (changes requested):

tests/unit/test_discovery_client.py
```diff
@@ -158,7 +164,35 @@ def test___start_discovery_service___key_file_exist_after_poll___service_start_s
...
+        return_value=temp_registration_json_file_path,
+    )
+    mocker.patch("subprocess.Popen", side_effect=FileNotFoundError)
+
+    discovery_client = DiscoveryClient()
+    registration_success_flag = discovery_client.register_measurement_service(
+        _TEST_SERVICE_PORT, _TEST_SERVICE_INFO, _TEST_MEASUREMENT_INFO
+    )
+
+    assert ~registration_success_flag
```
bkeryan, line 195: `~` is the bitwise ones-complement operator. Applying it to Booleans does not do what you expect it to.
```
>>> x,y=True,False
>>> ~x,~y
(-2, -1)
```
Please also fix the other test cases that misuse this operator.

bkeryan (changes requested):

ni_measurementlink_service/_internal/discovery_client.py
vigkre: Sure Brad. I have updated the code to handle exception in `register_measurement_service`

tests/unit/test_discovery_client.py
vigkre: Remo the usage of `~` operator

tests/unit/test_discovery_client.py
```diff
@@ -76,7 +77,7 @@ def test___discovery_service_available___unregister_non_registered_service___unr
 ):
     unregistration_success_flag = discovery_client.unregister_service()
 
-    assert ~unregistration_success_flag  # False
+    assert unregistration_success_flag
```
bkeryan, line 80: Is this supposed to fail or not? The test name says failure but the assertion says success.

vigkre: This test case scenario is:
   1. When measurement service is not registered with disc service.
   2. We cannot unregister a service which is not register at all.
   3. So, this logs and info that `Not registered with discovery service.`

And yes assertion can be updated to `unregistration_failure_flag`

bkeryan: `test___discovery_service_available___unregister_non_registered_service___unregistration_failure` means "Given that the discovery service is available, when I unregister a non-registered service, then I get an unregistration failure".

True means success and False means failure. The assertion expects True, which means success. However, the test name said you should get failure, not success. Which is it?

It makes no sense to rename the variable to mean the opposite of what it actually means.

vigkre: Okay, now I have updated the `unregister_service` method to return `False` if we are trying to unregister a non-registered service. And in test, I am expecting the assertion to be `False`.

tests/unit/test_discovery_client.py
```diff
@@ -76,7 +77,7 @@ def test___discovery_service_available___unregister_non_registered_service___unr
 ):
     unregistration_success_flag = discovery_client.unregister_service()
 
-    assert ~unregistration_success_flag  # False
+    assert not unregistration_success_flag
 
 
 def test___discovery_service_unavailable___register_service_registration_failure(
```
bkeryan, line 83: This test name is wrong and the code is now identical to `test___discovery_service_available___unregister_registered_service___unregistration_success`. 

I think your new tests cover the case where the discovery service is unavailable, so delete this test case.

bkeryan (changes requested):

tests/unit/test_discovery_client.py
vigkre: Deleted

bkeryan (approved):

samchris007 (approved):

Tharun-Sundar (approved):

Issue #331, opened 2023-07-12
Auto-launching discovery service produces a ResourceWarning


bkeryan:
# Bug Report

Auto-launching the discovery service leaks a `subprocess.Popen`, which produces a `ResourceWarning`. 

## Repro or Code Sample

Use Task Manager to kill the discovery service.
Manually launch an example service with `poetry run python measurement.py -v`

## Expected Behavior

Does not generate `ResourceWarning`.

## Current Behavior

Generates `ResourceWarning`.
```
PS C:\dev\measurementlink-python-examples-1.1.0\nidaqmx_analog_input> poetry run python .\measurement.py -v
2023-07-12 18:02:03,673 INFO: Measurement service hosted on port: 63699
C:\Users\Administrator\.pyenv\pyenv-win\versions\3.9.13\lib\subprocess.py:1052: ResourceWarning: subprocess 16572 is still running
  _warn("subprocess %s is still running" % self.pid,
ResourceWarning: Enable tracemalloc to get the object allocation traceback
[18:02:04 INF] Now listening on: http://localhost:42000
[18:02:04 INF] Application started. Press Ctrl+C to shut down.
[18:02:04 INF] Hosting environment: Production
[18:02:04 INF] Content root path: C:\Program Files\National Instruments\Shared\MeasurementLink\Discovery\
[18:02:04 ERR] Failed to parse static registration file: C:\Program Files\National Instruments\Shared\MeasurementLink\MeasurementLinkServices.json.
No valid services were found in the service configuration.
[18:02:04 INF] HTTP POST /ni.measurementlink.discovery.v1.DiscoveryService/RegisterService responded 2023-07-12 18:02:04,863 INFO: Successfully registered with discovery service.
Press enter to close the measurement service.
200 in 171.3586 ms
```

## Your Environment

* OS & Device: Windows 10
* `ni-measurementlink-service` version: 1.1.0
* MeasurementLink version: 23.5.0f778
* Python version: 3.9.13

[AB#2454176](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2454176)

vigkre:
@bkeryan Does this happen only for `nidaqmx_analog_input` example? 

I tested this issue and I can only reproduce for nidaqmx_analog_input example.

bkeryan:
[...]

PR #344, opened 2023-07-28
discovery_client: Fix auto-launching discovery service producing a ResourceWarning

vigkre:

- Save the reference of the `subprocess.popen` object to a global variable.
- This prevents leakage of `subprocess` which in turn denotes that the subprocess is still running and produces a `ResourceWarning`.

- Fixes -- Auto-launching the discovery service leaks a `subprocess.Popen`, which produces a `ResourceWarning`. 
- Issue #331 

1. Ran example (nidaqmx_analog_input) service and make sure it is registered with the discovery service.
2. Used Task Manager to kill the discovery service.
3. Manually launched the example service with `poetry run python measurement.py -v`.
4. Verified `ResourceWarning` doesn't happen anymore.
5. Repeated step 1 to 4 for other examples as well.
6. For other examples, in step 3 used `poetry run python -X dev measurement.py` to reproduce the `ResourceWarning`.

### DIFF
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -27,6 +27,9 @@ if sys.platform == "win32":
     import winerror
 
 _logger = logging.getLogger(__name__)
+# store sub-process that is running; it's only to avoid
+# "ResourceWarning: subprocess N is still running" from subprocess library
+disc_service_subprocess = None
 
 _START_SERVICE_TIMEOUT = 30.0
 _START_SERVICE_POLLING_INTERVAL = 100e-3
@@ -254,7 +257,8 @@ def _key_file_exists(key_file_path: pathlib.Path) -> bool:
 
 def _start_service(exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path) -> None:
     """Starts the service at the specified path and wait for the service to get up and running."""
-    subprocess.Popen([exe_file_path], cwd=exe_file_path.parent)
+    global disc_service_subprocess # only to avoid resource warnings about subprocess still running
+    disc_service_subprocess = subprocess.Popen([exe_file_path], cwd=exe_file_path.parent)
     # After the execution of process, check for key file existence in the path
     # stop checking after 30 seconds have elapsed and throw error
     timeout_time = time.time() + _START_SERVICE_TIMEOUT
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -257,7 +257,7 @@ def _key_file_exists(key_file_path: pathlib.Path) -> bool:
 
 def _start_service(exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path) -> None:
     """Starts the service at the specified path and wait for the service to get up and running."""
-    global disc_service_subprocess # only to avoid resource warnings about subprocess still running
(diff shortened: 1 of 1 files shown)


ni_measurementlink_service/_internal/discovery_client.py
```diff
@@ -27,6 +27,9 @@
     import winerror
 
 _logger = logging.getLogger(__name__)
+# store sub-process that is running; it's only to avoid
+# "ResourceWarning: subprocess N is still running" from subprocess library
+disc_service_subprocess = None
```
bkeryan, line 32: Naming:
- Don't abbreviate "discovery"
- Prefix with `_` to indicate that this variable is private

bkeryan, line 32: Type hint: `Optional[subprocess.Popen]`

ni_measurementlink_service/_internal/discovery_client.py
bkeryan, line 1: > Manually launched the example service with poetry run python measurement.py -v.

Did you test with `-W default` or `-X dev` as previously discussed? `ResourceWarning` is disabled by default and the only reason we saw it is because the `nidaqmx` module enables it for the entire process.

bkeryan (changes requested):

ni_measurementlink_service/_internal/discovery_client.py
```diff
@@ -27,6 +27,9 @@
     import winerror
 
 _logger = logging.getLogger(__name__)
+# store sub-process that is running; it's only to avoid
+# "ResourceWarning: subprocess N is still running" from subprocess library
```
bkeryan, line 31: This is wordier than it has to be.

```suggestion
# Save Popen object to avoid "ResourceWarning: subprocess N is still running"
```

ni_measurementlink_service/_internal/discovery_client.py
```diff
@@ -254,7 +257,8 @@ def _key_file_exists(key_file_path: pathlib.Path) -> bool:
 
 def _start_service(exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path) -> None:
     """Starts the service at the specified path and wait for the service to get up and running."""
-    subprocess.Popen([exe_file_path], cwd=exe_file_path.parent)
+    global disc_service_subprocess  # only to avoid resource warnings about subprocess still running
```
bkeryan, line 260:
```suggestion
    global disc_service_subprocess  # save Popen object to avoid ResourceWarning
```

ni_measurementlink_service/_internal/discovery_client.py
vigkre: Yes Brad, forgot to mention explictly in description. Updated now.

ni_measurementlink_service/_internal/discovery_client.py
vigkre: Updated

vigkre: Added type hint

ni_measurementlink_service/_internal/discovery_client.py
vigkre: updated

ni_measurementlink_service/_internal/discovery_client.py
vigkre: updated

bkeryan (approved):

Tharun-Sundar (approved):

PR #345, opened 2023-08-07
Added Creation flag CREATE_BREAKAWAY_FROM_JOB to stop Discovery service terminating in Debug mode

MounikaBattu17:

When a measurement and Discovery Service are started in `Debug mode`, the Discovery service from task manager is killed/terminated as soon as the measurement is stopped. So to stop Discovery Service from terminating the following changes are made.
1) Added  a `creationflags` parameter to `subprocess.Popen` in `discovery_client.py`.
2) Set the `creationflags` to `subprocess.CREATE_BREAKAWAY_FROM_JOB`.
3) Updated tests in `test_discovery_client.py` with `creationflags` parameter.

This PR fixes [Bug 2474096](https://dev.azure.com/ni/DevCentral/_workitems/edit/2474096): Discovery Service is terminated when measurement service is stopped in debug mode

Tested these scenarios
image
Ran `poetry run pytest`. All tests passed.
Ran the tests in debug mode and observed Discovery Service is not terminated after test run ends.

### DIFF
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -8,7 +8,7 @@ import subprocess
 import sys
 import time
 import typing
-from typing import Optional
+from typing import Any, Dict, Optional
 
 import grpc
 
@@ -260,15 +260,16 @@ def _key_file_exists(key_file_path: pathlib.Path) -> bool:
 def _start_service(exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path) -> None:
     """Starts the service at the specified path and wait for the service to get up and running."""
     global _discovery_service_subprocess  # save Popen object to avoid ResourceWarning
-
-    # Added hexadecimal equivalent of subprocess.CREATE_BREAKAWAY_FROM_JOB to avoid
-    # Mypy error 'Module subprocess has no attribute CREATE_BREAKAWAY_FROM_JOB'.
+    kwargs: Dict[str, Any] = {}
+    if sys.platform == "win32":
+        # Terminating the measurement service should not terminate the discovery service.
+        kwargs["creationflags"] = subprocess.CREATE_BREAKAWAY_FROM_JOB
     _discovery_service_subprocess = subprocess.Popen(
         [exe_file_path],
         cwd=exe_file_path.parent,
         stdout=subprocess.DEVNULL,
         stderr=subprocess.DEVNULL,
-        creationflags=0x1000000,
+        **kwargs,
     )
     # After the execution of process, check for key file existence in the path
     # stop checking after 30 seconds have elapsed and throw error
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -8,7 +8,6 @@ import subprocess
 import sys
 import time
 import typing
-from subprocess import CREATE_BREAKAWAY_FROM_JOB
 from typing import Optional
 
 import grpc
@@ -18,6 +17,7 @@ from ni_measurementlink_service._internal.stubs.ni.measurementlink.discovery.v1 
     discovery_service_pb2_grpc,
 )
 from ni_measurementlink_service.measurement.info import MeasurementInfo, ServiceInfo
+from subprocess import CREATE_BREAKAWAY_FROM_JOB
 
 if sys.platform == "win32":
     import errno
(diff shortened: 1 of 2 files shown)


ni_measurementlink_service/_internal/discovery_client.py
```diff
@@ -260,11 +260,13 @@ def _key_file_exists(key_file_path: pathlib.Path) -> bool:
 def _start_service(exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path) -> None:
     """Starts the service at the specified path and wait for the service to get up and running."""
     global _discovery_service_subprocess  # save Popen object to avoid ResourceWarning
+    creationflags = subprocess.CREATE_BREAKAWAY_FROM_JOB
```
dixonjoel, line 263: Why do we need a variable for one usage?

dixonjoel (approved):

ni_measurementlink_service/_internal/discovery_client.py
MounikaBattu17: I have reverted back. For test purpose I made the change since `Mypy` was throwing error on the `CREATE_BREAKAWAY_FROM_JOB`. Now I have used the hexadecimal equivalent `0x1000000` to resolve the `Mypy` issue

MounikaBattu17:
`Mypy` was throwing error on `subprocess.CREATE_BREAKAWAY_FROM_JOB/_winapi.CREATE_BREAKAWAY_FROM_JOB`. 

`ni_measurementlink_service/_internal/discovery_client.py:263: error: Module has no attribute "CREATE_BREAKAWAY_FROM_JOB"  [attr-defined]`

So I have used the hexadecimal value for the `0x1000000`.

ni_measurementlink_service/_internal/discovery_client.py
```diff
@@ -260,11 +260,15 @@ def _key_file_exists(key_file_path: pathlib.Path) -> bool:
 def _start_service(exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path) -> None:
     """Starts the service at the specified path and wait for the service to get up and running."""
     global _discovery_service_subprocess  # save Popen object to avoid ResourceWarning
+
+    # Added hexadecimal equivalent of subprocess.CREATE_BREAKAWAY_FROM_JOB to avoid
+    # Mypy error 'Module subprocess has no attribute CREATE_BREAKAWAY_FROM_JOB'.
```
bkeryan, line 265: The reason for this error is that `CREATE_BREAKAWAY_FROM_JOB` is Windows-specific:
- https://github.com/python/cpython/blob/main/Lib/subprocess.py#L80
- https://github.com/python/typeshed/blob/main/stdlib/subprocess.pyi#L2585

In addition, the `creationflags` parameter is only supported on Windows and will cause an error on Linux.
- https://github.com/python/cpython/blob/main/Lib/subprocess.py#L854

Our build workflow runs mypy on Linux, so Windows-specific constants are not available. The solution is to check `sys.platform` before calling platform-specific code.

```
kwargs: Dict[str, Any] = {}
if sys.platform == "win32":
    # Terminating the measurement service should not terminate the discovery service.
    kwargs["creationflags"] = subprocess.CREATE_BREAKAWAY_FROM_JOB

_discovery_service_subprocess = subprocess.Popen(
    ...
    stderr=subprocess.DEVNULL,
    **kwargs
)
```

tests/unit/test_discovery_client.py
```diff
@@ -109,6 +109,7 @@ def test___get_discovery_service_address___start_service_jit___returns_expected_
         cwd=exe_file_path.parent,
         stdout=subprocess.DEVNULL,
         stderr=subprocess.DEVNULL,
+        creationflags=0x1000000,
```
bkeryan, line 112: Add `sys.platform` checks and use the constant.

bkeryan (changes requested):

ni_measurementlink_service/_internal/discovery_client.py
MounikaBattu17: Thanks for the inputs Brad! I have updated the `creationflags` as plaform specific. The same change I have done in tests as well.

tests/unit/test_discovery_client.py
MounikaBattu17: I used this constant because `mypy` was throwing error for `CREATE_BREAK_AWAY_FROM_JOB`. But now we have made it as platform specific, error is no longer thrown. So I kept the attribute name itself instead of constant.

tests/unit/test_discovery_client.py
```diff
@@ -183,6 +190,14 @@ def test___discovery_service_exe_unavailable___register_service___registration_f
     assert not registration_success_flag
 
 
+@pytest.fixture(scope="module")
+def subprocess_popen_creationflags() -> Dict[str, Any]:
```
bkeryan, line 194: Rename to `subprocess_popen_kwargs`. `creationflags` is only one element of the kwargs and we will likely add a different one for Linux.

bkeryan (approved):
Approved with suggestions.

PR #346, opened 2023-08-08
Killing the Discovery Service started by the Test Environment

MounikaBattu17:

Acceptance tests (`test_measurement_service.py, test_streaming_data_measurement.py, test_yield_vs_return.py`) starts the discovery service for running their tests but doesn't terminate/kill it. This causes a `Resource Warning` while running all tests.
Approach is to start the Discovery service independently and scope it to whole test session.
1) Created a Context Manager  `DiscoveryServiceProcess` in `tests/utilities/discovery_service_process.py`.
      - `__init__` is responsible for starting the discovery service and store the instance in `self._proc`.
      - `__exit__` is responsible for killing the active Discovery service instance.
2) Added a pytest fixture `discovery_service_process` that is responsible for 
     - skipping tests for non-windows and measurement link registration file not found.
     -  yielding the `proc` and scoped it to `session` for windows platform.
3) Called the `discovery_service_process` fixture in `measurement_service` fixture in each acceptance test files.

1) Starts and Closes Discovery Service instances.
2) Fixes Resource Warning while running tests.

1) Ran `poetry run pytest`, Observed there is no `Resource Warning`.
2) Observed Discovery Service is started once and closed at the end of the test session.
3) Verified acceptance tests are skipped if platform is non windows or measurement link registration file not available.

### DIFF
--- a/ni_measurementlink_service/_internal/utilities/globaltestingstate.py
+++ b/ni_measurementlink_service/_internal/utilities/globaltestingstate.py
@@ -1,8 +1,0 @@
-"""Contains indicators to identify the Production/Test environment.
-"""
-
-
-class GlobalTestingState:
-    """Indicator for measurement running in testing environment."""
-
-    IsInTestState: bool = False
--- a/ni_measurementlink_service/_internal/utilities/globaltestingstate.py
+++ b/ni_measurementlink_service/_internal/utilities/globaltestingstate.py
@@ -1,0 +1,8 @@
+"""Contains indicators to identify the Production/Test environment.
+"""
+
+
+class GlobalTestingState:
+    """Indicator for measurement running in testing environment."""
+
+    IsInTestState: bool = False
--- a/tests/acceptance/test_measurement_service.py
+++ b/tests/acceptance/test_measurement_service.py
@@ -205,8 +205,9 @@ def test___measurement_service_v2___measure_with_large_array___returns_output(
 
 
 @pytest.fixture(scope="module")
-def measurement_service() -> Generator[MeasurementService, None, None]:
+def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
+    discovery_service_process
     with loopback_measurement.measurement_service.host_service() as service:
         yield service
 
--- a/tests/acceptance/test_measurement_service.py
+++ b/tests/acceptance/test_measurement_service.py
@@ -205,9 +205,8 @@ def test___measurement_service_v2___measure_with_large_array___returns_output(
 
 
 @pytest.fixture(scope="module")
-def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
+def measurement_service() -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
-    discovery_service_process
     with loopback_measurement.measurement_service.host_service() as service:
         yield service
 
--- a/tests/acceptance/test_measurement_service.py
+++ b/tests/acceptance/test_measurement_service.py
@@ -207,6 +207,7 @@ def test___measurement_service_v2___measure_with_large_array___returns_output(
 @pytest.fixture(scope="module")
 def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
+    discovery_service_process
     with loopback_measurement.measurement_service.host_service() as service:
         yield service
 
--- a/tests/acceptance/test_measurement_service.py
+++ b/tests/acceptance/test_measurement_service.py
@@ -207,7 +207,7 @@ def test___measurement_service_v2___measure_with_large_array___returns_output(
 @pytest.fixture(scope="module")
 def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
-    _ = discovery_service_process
+    discovery_service_process
     with loopback_measurement.measurement_service.host_service() as service:
         yield service
 
--- a/tests/acceptance/test_measurement_service.py
+++ b/tests/acceptance/test_measurement_service.py
@@ -17,7 +17,6 @@ from ni_measurementlink_service._internal.stubs.ni.measurementlink.measurement.v
     measurement_service_pb2 as v2_measurement_service_pb2,
     measurement_service_pb2_grpc as v2_measurement_service_pb2_grpc,
 )
-from ni_measurementlink_service._internal.utilities.globaltestingstate import GlobalTestingState
 from ni_measurementlink_service.measurement.service import MeasurementService
 from tests.assets import sample_measurement_test_pb2
 from tests.assets.sample_measurement_test_pb2 import ProtobufColor
@@ -206,9 +205,9 @@ def test___measurement_service_v2___measure_with_large_array___returns_output(
 
 
 @pytest.fixture(scope="module")
-def measurement_service() -> Generator[MeasurementService, None, None]:
+def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
-    GlobalTestingState.IsInTestState = True
+    _ = discovery_service_process
     with loopback_measurement.measurement_service.host_service() as service:
         yield service
 
--- a/tests/acceptance/test_measurement_service.py
+++ b/tests/acceptance/test_measurement_service.py
@@ -205,7 +205,7 @@ def test___measurement_service_v2___measure_with_large_array___returns_output(
 
 
 @pytest.fixture(scope="module")
-def measurement_service() -> Generator[MeasurementService, None, None]:
+def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
     with loopback_measurement.measurement_service.host_service() as service:
         yield service
--- a/tests/acceptance/test_measurement_service.py
+++ b/tests/acceptance/test_measurement_service.py
@@ -207,7 +207,6 @@ def test___measurement_service_v2___measure_with_large_array___returns_output(
 @pytest.fixture(scope="module")
 def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
-    discovery_service_process
     with loopback_measurement.measurement_service.host_service() as service:
         yield service
 
--- a/tests/acceptance/test_measurement_service.py
+++ b/tests/acceptance/test_measurement_service.py
@@ -17,6 +17,7 @@ from ni_measurementlink_service._internal.stubs.ni.measurementlink.measurement.v
     measurement_service_pb2 as v2_measurement_service_pb2,
     measurement_service_pb2_grpc as v2_measurement_service_pb2_grpc,
 )
+from ni_measurementlink_service._internal.utilities.globaltestingstate import GlobalTestingState
 from ni_measurementlink_service.measurement.service import MeasurementService
 from tests.assets import sample_measurement_test_pb2
 from tests.assets.sample_measurement_test_pb2 import ProtobufColor
@@ -207,6 +208,7 @@ def test___measurement_service_v2___measure_with_large_array___returns_output(
 @pytest.fixture(scope="module")
 def measurement_service() -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
+    GlobalTestingState.IsInTestState = True
     with loopback_measurement.measurement_service.host_service() as service:
         yield service
 
--- a/tests/acceptance/test_streaming_data_measurement.py
+++ b/tests/acceptance/test_streaming_data_measurement.py
@@ -9,6 +9,7 @@ from ni_measurementlink_service._internal.stubs.ni.measurementlink.measurement.v
     measurement_service_pb2 as v2_measurement_service_pb2,
     measurement_service_pb2_grpc as v2_measurement_service_pb2_grpc,
 )
+from ni_measurementlink_service._internal.utilities.globaltestingstate import GlobalTestingState
 from ni_measurementlink_service.measurement.service import MeasurementService
 from tests.assets import sample_streaming_measurement_test_pb2
 from tests.utilities import streaming_data_measurement
@@ -139,5 +140,6 @@ def _get_serialized_measurement_outputs(
 @pytest.fixture(scope="module")
 def measurement_service() -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
+    GlobalTestingState.IsInTestState = True
     with streaming_data_measurement.measurement_service.host_service() as service:
         yield service
--- a/tests/acceptance/test_streaming_data_measurement.py
+++ b/tests/acceptance/test_streaming_data_measurement.py
@@ -139,6 +139,6 @@ def _get_serialized_measurement_outputs(
 @pytest.fixture(scope="module")
 def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
-    _ = discovery_service_process
+    discovery_service_process
     with streaming_data_measurement.measurement_service.host_service() as service:
         yield service
--- a/tests/acceptance/test_streaming_data_measurement.py
+++ b/tests/acceptance/test_streaming_data_measurement.py
@@ -9,7 +9,6 @@ from ni_measurementlink_service._internal.stubs.ni.measurementlink.measurement.v
     measurement_service_pb2 as v2_measurement_service_pb2,
     measurement_service_pb2_grpc as v2_measurement_service_pb2_grpc,
 )
-from ni_measurementlink_service._internal.utilities.globaltestingstate import GlobalTestingState
 from ni_measurementlink_service.measurement.service import MeasurementService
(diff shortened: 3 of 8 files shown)


vigkre (approved):

ni_measurementlink_service/_internal/utilities/globaltestingstate.py
bkeryan, line 1: I think you should delete this file and use a different approach. 

However, I want to share my feedback on this file anyway:
- I do not like `_internal`. Going forward, I would prefer to prefix module names with `_` to indicate that they are internal.
- I think `utilities` is too general and does not make a meaningful distinction.
- The `GlobalTestingState` class does not need to exist. Python allows global variables at file/module scope.
- `IsInTestState` isn't following https://pep8.org/#method-names-and-instance-variables
- The wording of `IsInTestState` doesn't really make sense to me. The corresponding C# property is `IsInTestCode`, not `IsInTestCode`.

bkeryan (changes requested):

tests/utilities/discovery_service_process.py
```diff
@@ -0,0 +1,38 @@
+"""Generator class to create and terminate Discovery Service instance."""
```
bkeryan, line 1: This is a context manager, not a generator.

```suggestion
"""Class to create and terminate Discovery Service instance."""
```

tests/utilities/discovery_service_process.py
```diff
@@ -0,0 +1,38 @@
...
+    _get_key_file_path,
+    _service_already_running,
+    _start_service,
+)
+
+
+class DiscoveryServiceProcess:
+    """Maintains the processes involved in creating and terminating discovery service."""
+
+    def __init__(self):
```
bkeryan, line 14:
```suggestion
    def __init__(self) -> None:
```

tests/utilities/discovery_service_process.py
```diff
@@ -0,0 +1,38 @@
...
+            self._proc = None
+            key_file_path = _get_key_file_path()
+            if _service_already_running(key_file_path):
+                return
+            self._proc = _start_service(_get_discovery_service_location(), key_file_path)
+        except Exception:
+            self._close_discovery_service()
+            raise
+
+    def __enter__(self):
```
bkeryan, line 26: Add one of these (preferably the 1st, if it works):
- `from typing_extensions import Self`
- `Self = TypeVar("Self", bound=object)`

```suggestion
    def __enter__(self: Self) -> Self:
```

tests/utilities/discovery_service_process.py
```diff
@@ -0,0 +1,38 @@
...
+        """Returns the DiscoveryServiceProcess instance."""
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
bkeryan, line 30:
```suggestion
    def __exit__(self, exc_type: Optional[Type[BaseException]], exc_val: Optional[BaseException], exc_tb: Optional[TracebackType]) -> None:
```

tests/utilities/discovery_service_process.py
```diff
@@ -0,0 +1,38 @@
...
+        """Closes the DiscoveryServiceProcess instance."""
+        self._close_discovery_service()
+
+    def _close_discovery_service(self):
```
bkeryan, line 34:
```suggestion
    def _close_discovery_service(self) -> None:
```

tests/acceptance/test_yield_vs_return.py
```diff
@@ -53,7 +53,8 @@ def _get_serialized_measurement_configuration_parameters(
 
 
 @pytest.fixture(scope="module")
-def measurement_service() -> Generator[MeasurementService, None, None]:
+def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
+    _ = discovery_service_process
```
bkeryan, line 58: Did ni-python-styleguide warn about the unused parameter?

bkeryan (changes requested):

ni_measurementlink_service/_internal/utilities/globaltestingstate.py
bkeryan: The new fixture looks good.

tests/acceptance/test_yield_vs_return.py
MounikaBattu17: I ran `poetry run ni-python-styleguide lint`, no errors.

mshafer-NI: a: assigining it to "_" (throwing it away) is referencing it, 
b: unused parens are allowed.

Remove this assignment.

vigkre (approved):

dixonjoel:
Please update the PR description to match the implementation since you're not using the `IsInTestState` global now.

tests/utilities/discovery_service_process.py
```diff
@@ -0,0 +1,48 @@
...
+        """Creates a DiscoveryServiceProcess instance."""
+        try:
+            self._proc = None
+            key_file_path = _get_key_file_path()
+            if _service_already_running(key_file_path):
+                return
+            self._proc = _start_service(_get_discovery_service_location(), key_file_path)
+        except Exception:
+            self._close_discovery_service()
+            raise
```
MounikaBattu17, line 29: @bkeryan The pipeline(Run Tests for all platforms, all acceptance tests) errors if I raise the exception here. There are 2 errors that occurs in `discovery_client.py`
1) Non windows platform - `NotImplementedError("Platform not supported")` error from `_get_key_file_directory()`
2) windows platform - FileNotFound error from `_get_discovery_service_location()`

`discovery_client.py` handles these exceptions by logging them.

Due to these errors, Discovery Service is not even started and the acceptance tests pass showing that these tests are independent of Discovery Service running or not

So, my question is 
1) Whether we should validate the loggers for appropriate error messages in acceptance tests? or
2) Since the tests are independent on Discovery Service, can we fake the discovery service stub?

tests/acceptance/test_yield_vs_return.py
bkeryan: @MounikaBattu17 Why are you still explicitly referencing the unused parameter?

tests/utilities/discovery_service_process.py
bkeryan: Another option is to have the fixture call pytest.skip if the discovery service is not installed, like nidaqmx-python does. This is pretty easy to do and would unblock this PR, but we still need to figure out our strategy for testing Python code that uses real MeasurementLink core services.

tests/acceptance/test_yield_vs_return.py
MounikaBattu17: This file change is outdated. I have removed the unused params(`_`). Please refer to the new change.

tests/utilities/discovery_service_process.py
MounikaBattu17: @bkeryan we want some clarifications on the Acceptance tests. So as we know acceptance tests starts and tries to register the measurements to the discovery service, 
If Discovery Service  is successfully started, it is not terminated
If Discovery Service  doesn't start successfully, exceptions are logged and tests pass

In the second case, the tests passing even though the measurement is not even registered to the discovery service is conveying that the overall workflow is not tested in the "Acceptance Tests" and I assume that Acceptance tests are end to end testing that tests the complete workflow of running a measurement. Please correct me If I'm wrong here.

So, we want to make the discovery service up and running for windows (which is only supported) such that acceptance tests works with the discovery service. To achieve this we are proposing the following approaches:

**1) Make Acceptance tests run only for windows:**
Mark all acceptance tests as windows only and start and terminate the discovery service using the `discovery_service_process` fixture. Doing this we will be testing the complete workflow needed to run a measurement.
(or)
**2) Use Fake discovery client stub for non windows platforms:**
Run acceptance tests for all platforms and use a fake discovery service if it is a non windows platform and use `discovery_service_process` for windows.
(or)
**3) Add new acceptance tests to verify discovery logs:**
Add new acceptance tests to verify the logs of Discovery service, thus we can test the error logs for non windows platform and success logs for windows platform

Please share your views on the approaches and direct on what approach to take or any other different direction will be helpful.

Ps: If we are going for approach 1, there is an `FileNotFound` error in pipeline, we are unsure of why that's occurring. please help us in debugging it.

MounikaBattu17:
> @MounikaBattu17 Did #344 not fix the ResourceWarning?

Yes, It fixed the resource warning. Now the ResourceWarning doesn't come if I run a measurement.

bkeryan:
> This causes a `Resource Warning` while running all tests.

If the resource warning was already fixed, why does the description of this PR talk about the resource warning?

tests/utilities/discovery_service_process.py
bkeryan: > 1) Make Acceptance tests run only for windows:
Mark all acceptance tests as windows only and start and terminate the discovery service using the discovery_service_process fixture. Doing this we will be testing the complete workflow needed to run a measurement.

> Ps: If we are going for approach 1, there is an FileNotFound error in pipeline, we are unsure of why that's occurring. please help us in debugging it.

```
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Program Files\\National Instruments\\Shared\\MeasurementLink\\MeasurementLinkServices.json'
```

The GitHub-hosted runner does not have MeasurementLink installed, so the JSON registration file does not exist. Note that this is separate from the key file. The registration file is created by the MeasurementLink installer, and the key file is created by the discovery service itself.
 
For this PR, I think it is reasonable for the discovery service test fixture to call pytest.skip in these situations:
- sys.platform != "win32"
- JSON registration file does not exist

I think skipping in the test fixture is better than adding `@pytest.skipif(sys.platform != "win32", reason="...")` to a bunch of tests.

> 2) Use Fake discovery client stub for non windows platforms:
Run acceptance tests for all platforms and use a fake discovery service for a non windows platform and use discovery_service_process for windows.

This adds more non-Windows test coverage but it's more effort/complication than skipping the tests on non-Windows platforms.

> 3) Add new acceptance tests to verify discovery logs:
Add new acceptance tests to verify the logs of Discovery service, thus we can test the error logs for non windows platform and success logs for windows platform

I don't think this is worth it. I would rather that we improve the error handling so that it's not necessary to check the logs for errors.

> Please share your views on the approaches and direct on what approach to take or any other different direction will be helpful.

I like this approach best:
- Now: have discovery fixture skip on non-Windows
- Now: have discovery fixture skip when MeasurementLink is not installed (registration file not found)
- Separately: improve error handling to exit when registration fails
- Separately: add self-hosted GitHub runner w/MeasurementLink installed (note that we might have a simpler way to do this than what we did for nidaqmx-python)

@dixonjoel @pbirkhol-ni

Tharun-Sundar (approved):

tests/conftest.py
```diff
@@ -48,5 +51,16 @@ def stub_v2(grpc_channel: grpc.Channel) -> v2_measurement_service_pb2_grpc.Measu
...
 def discovery_service_process() -> Generator[DiscoveryServiceProcess, None, None]:
     """Test fixture that creates discovery service process."""
-    with DiscoveryServiceProcess() as proc:
-        yield proc
+    if sys.platform != "win32":
+        pytest.skip(
+            f"Platform {sys.platform} is not supported for starting discovery service."
+            "Could not proceed to run the tests."
+        )
+    elif os.path.exists(_get_registration_json_file_path()):
```
vigkre: @MounikaBattu17  Can we use `pathlib.Path.exists` function here?

vigkre (approved):

tests/conftest.py
```diff
@@ -42,3 +45,21 @@ def stub_v1(grpc_channel: grpc.Channel) -> v1_measurement_service_pb2_grpc.Measu
...
+
+@pytest.fixture(scope="session")
+def discovery_service_process() -> Generator[DiscoveryServiceProcess, None, None]:
+    """Test fixture that creates discovery service process."""
+    if sys.platform != "win32":
+        pytest.skip(
+            f"Platform {sys.platform} is not supported for starting discovery service."
+            "Could not proceed to run the tests."
+        )
+    elif not pathlib.Path.exists(_get_registration_json_file_path()):
```
mshafer-NI, line 58:
```suggestion
    elif not pathlib.Path(_get_registration_json_file_path()).exists():
```

This is if _get_registration_json_file_path() returns a str, if it returns a Path object, just do

```suggestion
    elif not _get_registration_json_file_path().exists():
```

tests/utilities/discovery_service_process.py
MounikaBattu17:
```
Now: have discovery fixture skip on non-Windows
Now: have discovery fixture skip when MeasurementLink is not installed (registration file not found)
```
Thanks @bkeryan for the approach. I have added if conditions in `discovery_service_process` fixture in `conftest.py` to skip tests for  non-Windows and when MeasurementLink is not installed (registration file not found) with appropriate messages.

As per the other items:
`Separately: improve error handling to exit when registration fails` - This will be addressed in the upcoming PR.

`Separately: add self-hosted GitHub runner w/MeasurementLink installed (note that we might have a simpler way to do this than what we did for nidaqmx-python)` - I have created a **[Technical Debt 2486477](https://dev.azure.com/ni/DevCentral/_workitems/edit/2486477): Add Self hosted GitHub runner to run tests in measurementlink-python pipeline** to track.

tests/conftest.py
MounikaBattu17: The return type is a `Path` object, So I have used the second suggested change.

MounikaBattu17:
> > This causes a `Resource Warning` while running all tests.
> 
> If the resource warning was already fixed, why does the description of this PR talk about the resource warning?

The resource warning which I'm referring is from the tests. Earlier/ without this PR changes, there is a resource warning after the test run.

tests/utilities/discovery_service_process.py
dixonjoel: We have 'spawningpool' that we use for running the LabVIEW tests: https://github.com/ni/measurementlink-labview/settings/actions/runners
It's already set up as a self-hosted runner. I wonder if it's feasible to use it on two repos.

dixonjoel: In the meantime, it seems like we should be running the skipped tests locally with MeasurementLink installed before submitting PRs.

dixonjoel (approved):
All tests pass locally for me and don't get skipped when I have the DS installed.

tests/acceptance/test_yield_vs_return.py
bkeryan: Why is this still in the code?

```
@pytest.fixture(scope="module")
def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
    """Test fixture that creates and hosts a measurement service."""
    discovery_service_process  # <- THIS
    with yield_vs_return_measurement.measurement_service.host_service() as service:
        yield service
```

tests/acceptance/test_measurement_service.py
```diff
@@ -205,8 +205,9 @@ def test___measurement_service_v2___measure_with_large_array___returns_output(
 
 
 @pytest.fixture(scope="module")
-def measurement_service() -> Generator[MeasurementService, None, None]:
+def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
+    discovery_service_process
```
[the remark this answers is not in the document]
bkeryan: See this discussion: https://github.com/ni/measurementlink-python/pull/346#discussion_r1289046069

tests/acceptance/test_yield_vs_return.py
```diff
@@ -53,7 +53,8 @@ def _get_serialized_measurement_configuration_parameters(
 
 
 @pytest.fixture(scope="module")
-def measurement_service() -> Generator[MeasurementService, None, None]:
+def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
+    discovery_service_process
```
bkeryan:
```suggestion
```

tests/acceptance/test_streaming_data_measurement.py
```diff
@@ -137,7 +137,8 @@ def _get_serialized_measurement_outputs(
 
 
 @pytest.fixture(scope="module")
-def measurement_service() -> Generator[MeasurementService, None, None]:
+def measurement_service(discovery_service_process) -> Generator[MeasurementService, None, None]:
     """Test fixture that creates and hosts a measurement service."""
+    discovery_service_process
```
bkeryan:
```suggestion
```

tests/acceptance/test_measurement_service.py
bkeryan:
```suggestion
```

tests/utilities/discovery_service_process.py
bkeryan: Skipping tests is sufficient for this PR.

Adding a self-hosted runner is a matter for another PR, and we probably need to create a work item for it too.

bkeryan (changes requested):

tests/acceptance/test_yield_vs_return.py
MounikaBattu17: Sorry for the confusion. I misinterpreted wrongly and removed only the left hand side assignment. Now I have corrected.

tests/acceptance/test_measurement_service.py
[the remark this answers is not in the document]
MounikaBattu17: Sorry for the confusion. I removed the unused call since it is dependency injected.

MounikaBattu17:
This PR has got necessary approval. Hence merging.

Issue #349, opened 2023-08-10
Service registration errors are not reported to the calling process via exit code


bkeryan:
# Bug Report

`DiscoveryClient.register_measurement_service` reports service registration errors by [logging to the console and returning `False`](https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/_internal/discovery_client.py#L128). The calling code [ignores the return value](https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/_internal/service_manager.py#L104C56-L104C56). This means that the measurement service continues running, but no other services can access it.

When the discovery service activates a measurement service, it waits 30 seconds for the measurement service to register with the discovery service. If the measurement service reported the error normally, it would be reported sooner.

I think the bigger problem with this is that the acceptance tests rely on this behavior to run on a GitHub runner without the MeasurementLink discovery service installed. It seems like the production code is ignoring an error condition for the benefit of the tests.

## Repro or Code Sample

https://github.com/ni/measurementlink-python/pull/346

## Expected Behavior

Acceptance tests either don't start the discovery service or they are skipped when running on a GitHub-hosted runner.

## Current Behavior

Acceptance tests try to start the discovery service, but if it fails, they ignore the error. If it works, the discovery service keeps running when the tests are done.

## Possible Solution

Add a pytest mark to skip tests that require the discovery service.

-or-

Update acceptance tests to use a fake discovery client.

## Context

https://github.com/ni/measurementlink-python/pull/346

## Your Environment

PR build

[AB#2485046](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2485046)

PR #353, opened 2023-08-16
Improve error handling in measurement register with discovery service

MounikaBattu17:

**This PR stops measurement service from starting if the measurement is not successfully registered with discovery service.**

Currently, On a unsuccessful measurement register with discovery service, the exceptions are just logged and service continues to run. So to improve the error handling(not starting the measurement service on unsuccessful register), made the following changes:
1) Raised exceptions instead of returning False in `register_measurement_service` and `unregister_service` from `discovery_client.py`.
2) Propagated the raised exceptions to callers `service_manager.py`, `service.py`.

**Tests Updates:**
1) Updated `unit/test_service.py` to cover the exception scenario of `host_service` method.
2) Updated `integration/test_service_manager.py` to cover the exception scenario of `start` method.
   - Added a new fixture `expect_discovery_service_error_stub` return a boolean to choose from `FakeDiscoveryServiceStub` or `FakeDiscoveryServiceStubError`. By default, Return value `False` such that choosing `FakeDiscoveryServiceStub` for all test cases
   - Overridden the value to `True` to pick `FakeDiscoveryServiceStubError` for the one test cases testing exception scenario.
3) Updated test cases assert  in `unit/test_discovery_client.py`. 

This PR stops measurement service from starting if the measurement is not successfully registered with discovery service by raising an exception.

1) Added new tests to cover the exception scenarios and verified existing and new tests pass by running `poetry run pytest`.
2) Ran sample measurement example and verified the below scenarios,
  - Sample measurement successfully registered with discovery service -> Able to run the measurement without any error
  - Sample measurement not successfully registered with discovery service (removed measurementlink json file) -> measurement is not started and throwed an error
```
  File "measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 125, in register_measurement_service
    register_response = self.stub.RegisterService(request)
  File "measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 84, in stub
    address = _get_discovery_service_address()
  File "measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 220, in _get_discovery_service_address
    _ensure_discovery_service_started(key_file_path)
  File "measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 232, in _ensure_discovery_service_started
    exe_file_path = _get_discovery_service_location()
  File "measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 240, in _get_discovery_service_location
    registration_json_obj = json.loads(registration_json_path.read_text())
  File "Python\Python38\lib\pathlib.py", line 1236, in read_text        
    with self.open(mode='r', encoding=encoding, errors=errors) as f:
  File "Python\Python38\lib\pathlib.py", line 1222, in open
    return io.open(self, mode, buffering, encoding, errors, newline,
  File "Python\Python38\lib\pathlib.py", line 1078, in _opener
    return self._accessor.open(self, flags, mode)
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Program Files\\National Instruments\\Shared\\MeasurementLink\\MeasurementLinkService.json'

The above exception was the direct cause of the following exception:

  File "/measurementlink-python/examples/sample_measurement/measurement.py", line 113, in <module>
    main()
  File "measurementlink-python\examples\sample_measurement\.venv\lib\site-packages\click\core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "measurementlink-python\examples\sample_measurement\.venv\lib\site-packages\click\core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "measurementlink-python\examples\sample_measurement\.venv\lib\site-packages\click\core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "measurementlink-python\examples\sample_measurement\.venv\lib\site-packages\click\core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/measurementlink-python/examples/sample_measurement/measurement.py", line 108, in main
    with sample_measurement_service.host_service():
  File "measurementlink-python\ni_measurementlink_service\measurement\service.py", line 388, in host_service
    raise e
  File "measurementlink-python\ni_measurementlink_service\measurement\service.py", line 379, in host_service
    self.grpc_service.start(
  File "measurementlink-python\ni_measurementlink_service\_internal\service_manager.py", line 111, in start
    raise e
  File "measurementlink-python\ni_measurementlink_service\_internal\service_manager.py", line 105, in start
    self.discovery_client.register_measurement_service(port, service_info, measurement_info)
  File "measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 140, in register_measurement_service
    raise FileNotFoundError(e.strerror, e.filename) from e
FileNotFoundError: [Errno No such file or directory] C:\Program Files\National Instruments\Shared\MeasurementLink\MeasurementLinkService.json
(sample-measurement-py3.8) PS measurementlink-python> 
```

### DIFF
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -80,9 +80,10 @@ class DiscoveryClient:
     @property
     def stub(self) -> discovery_service_pb2_grpc.DiscoveryServiceStub:
         """Get the gRPC stub used to interact with the discovery service."""
-        address = _get_discovery_service_address()
-        channel = grpc.insecure_channel(address)
-        self._stub = discovery_service_pb2_grpc.DiscoveryServiceStub(channel)
+        if self._stub is None:
+            address = _get_discovery_service_address()
+            channel = grpc.insecure_channel(address)
+            self._stub = discovery_service_pb2_grpc.DiscoveryServiceStub(channel)
         return self._stub
 
     def register_measurement_service(
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -16,7 +16,6 @@ from ni_measurementlink_service._internal.stubs.ni.measurementlink.discovery.v1 
     discovery_service_pb2,
     discovery_service_pb2_grpc,
 )
-from ni_measurementlink_service._internal.utilities.globaltestingstate import GlobalTestingState
 from ni_measurementlink_service.measurement.info import MeasurementInfo, ServiceInfo
 
 if sys.platform == "win32":
@@ -182,13 +181,6 @@ class DiscoveryClient:
         except Exception:
             _logger.exception("Error in unregistering with discovery service.")
             return False
-        finally:
-            global _discovery_service_subprocess
-            # Killing the Discovery Service started by the Test environment.
-            if GlobalTestingState.IsInTestState and _discovery_service_subprocess is not None:
-                _discovery_service_subprocess.kill()
-                _discovery_service_subprocess.communicate()
-
         return True
 
     def resolve_service(self, provided_interface: str, service_class: str = "") -> ServiceLocation:
@@ -238,7 +230,8 @@ def _ensure_discovery_service_started(key_file_path: pathlib.Path) -> None:
         return
 
     exe_file_path = _get_discovery_service_location()
-    _start_service(exe_file_path, key_file_path)
+    global _discovery_service_subprocess  # save Popen object to avoid ResourceWarning
+    _discovery_service_subprocess = _start_service(exe_file_path, key_file_path)
 
 
 def _get_discovery_service_location() -> pathlib.PurePath:
@@ -265,9 +258,10 @@ def _key_file_exists(key_file_path: pathlib.Path) -> bool:
     return key_file_path.is_file() and key_file_path.stat().st_size > 0
 
 
-def _start_service(exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path) -> None:
+def _start_service(
+    exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path
+) -> subprocess.Popen:
     """Starts the service at the specified path and wait for the service to get up and running."""
-    global _discovery_service_subprocess  # save Popen object to avoid ResourceWarning
     _discovery_service_subprocess = subprocess.Popen(
         [exe_file_path],
         cwd=exe_file_path.parent,
@@ -280,7 +274,7 @@ def _start_service(exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path)
     while True:
         try:
             with _open_key_file(str(key_file_path)) as _:
-                return
+                return _discovery_service_subprocess
         except IOError:
             pass
         if time.time() >= timeout_time:
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -132,15 +132,15 @@ class DiscoveryClient:
                 )
             else:
                 _logger.exception("Error in registering with discovery service.")
-            return False
-        except FileNotFoundError:
+            raise grpc.RpcError(type(e).__name__, e.args[0]) from e
+        except FileNotFoundError as e:
             _logger.error(
                 "Unable to register with discovery service. Possible reason: discovery service not running."
             )
-            return False
-        except Exception:
+            raise FileNotFoundError(e.strerror, e.filename) from e
+        except Exception as e:
             _logger.exception("Error in registering with discovery service.")
-            return False
+            raise Exception(type(e).__name__, e.args[0]) from e
         return True
 
     def unregister_service(self) -> bool:
@@ -172,15 +172,15 @@ class DiscoveryClient:
                 )
             else:
                 _logger.exception("Error in unregistering with discovery service.")
-            return False
-        except FileNotFoundError:
+            raise Exception(type(e).__name__, e.args[0]) from e
+        except FileNotFoundError as e:
             _logger.error(
                 "Unable to unregister with discovery service. Possible reason: discovery service not running."
             )
-            return False
-        except Exception:
+            raise FileNotFoundError(e.strerror, e.filename) from e
+        except Exception as e:
             _logger.exception("Error in unregistering with discovery service.")
-            return False
+            raise Exception(type(e).__name__, e.args[0]) from e
         return True
 
     def resolve_service(self, provided_interface: str, service_class: str = "") -> ServiceLocation:
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -258,7 +258,6 @@ def _key_file_exists(key_file_path: pathlib.Path) -> bool:
     return key_file_path.is_file() and key_file_path.stat().st_size > 0
 
 
-
 def _start_service(
     exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path
 ) -> subprocess.Popen:
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -258,6 +258,7 @@ def _key_file_exists(key_file_path: pathlib.Path) -> bool:
     return key_file_path.is_file() and key_file_path.stat().st_size > 0
 
 
+
 def _start_service(
     exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path
 ) -> subprocess.Popen:
@@ -266,7 +267,7 @@ def _start_service(
     if sys.platform == "win32":
         # Terminating the measurement service should not terminate the discovery service.
         kwargs["creationflags"] = subprocess.CREATE_BREAKAWAY_FROM_JOB
-    _discovery_service_subprocess = subprocess.Popen(
+    discovery_service_subprocess = subprocess.Popen(
         [exe_file_path],
         cwd=exe_file_path.parent,
         stdout=subprocess.DEVNULL,
@@ -279,7 +280,7 @@ def _start_service(
     while True:
         try:
             with _open_key_file(str(key_file_path)) as _:
-                return _discovery_service_subprocess
+                return discovery_service_subprocess
         except IOError:
             pass
         if time.time() >= timeout_time:
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -16,6 +16,7 @@ from ni_measurementlink_service._internal.stubs.ni.measurementlink.discovery.v1 
     discovery_service_pb2,
     discovery_service_pb2_grpc,
 )
+from ni_measurementlink_service._internal.utilities.globaltestingstate import GlobalTestingState
 from ni_measurementlink_service.measurement.info import MeasurementInfo, ServiceInfo
 
 if sys.platform == "win32":
@@ -181,6 +182,13 @@ class DiscoveryClient:
         except Exception:
             _logger.exception("Error in unregistering with discovery service.")
             return False
+        finally:
+            global _discovery_service_subprocess
+            # Killing the Discovery Service started by the Test environment.
+            if GlobalTestingState.IsInTestState and _discovery_service_subprocess is not None:
+                _discovery_service_subprocess.kill()
+                _discovery_service_subprocess.communicate()
+
         return True
 
     def resolve_service(self, provided_interface: str, service_class: str = "") -> ServiceLocation:
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -80,10 +80,9 @@ class DiscoveryClient:
     @property
     def stub(self) -> discovery_service_pb2_grpc.DiscoveryServiceStub:
         """Get the gRPC stub used to interact with the discovery service."""
-        if self._stub is None:
-            address = _get_discovery_service_address()
-            channel = grpc.insecure_channel(address)
-            self._stub = discovery_service_pb2_grpc.DiscoveryServiceStub(channel)
+        address = _get_discovery_service_address()
+        channel = grpc.insecure_channel(address)
+        self._stub = discovery_service_pb2_grpc.DiscoveryServiceStub(channel)
         return self._stub
 
     def register_measurement_service(
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -8,7 +8,7 @@ import subprocess
 import sys
 import time
 import typing
-from typing import Optional
+from typing import Any, Dict, Optional
 
 import grpc
 
@@ -262,11 +262,16 @@ def _start_service(
     exe_file_path: pathlib.PurePath, key_file_path: pathlib.Path
 ) -> subprocess.Popen:
     """Starts the service at the specified path and wait for the service to get up and running."""
+    kwargs: Dict[str, Any] = {}
+    if sys.platform == "win32":
+        # Terminating the measurement service should not terminate the discovery service.
+        kwargs["creationflags"] = subprocess.CREATE_BREAKAWAY_FROM_JOB
     _discovery_service_subprocess = subprocess.Popen(
         [exe_file_path],
         cwd=exe_file_path.parent,
         stdout=subprocess.DEVNULL,
         stderr=subprocess.DEVNULL,
+        **kwargs,
     )
     # After the execution of process, check for key file existence in the path
     # stop checking after 30 seconds have elapsed and throw error
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -132,15 +132,15 @@ class DiscoveryClient:
                 )
             else:
                 _logger.exception("Error in registering with discovery service.")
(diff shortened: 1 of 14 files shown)


dixonjoel:
@MounikaBattu17 @bkeryan @pbirkhol-ni 
Is this the behavior we want? I guess the scenario would be pretty uncommon in that the DS is not installed or not installed correctly on the machine?

I've always thought the service could still have at least some value running on its own independent of the DS, so I would have thought the logging would be enough, especially since we have the behavior of auto-starting the DS now. For example, if it's got a hard-coded port or reports the port it's on some way, a client could still connect without the DS.

What does everything think? Maybe it's been discussed in a PR or I missed some discussion.

bkeryan:
> @MounikaBattu17 @bkeryan @pbirkhol-ni Is this the behavior we want? 

I haven't reviewed this PR yet, but I think it is intended to fix #349 which I filed.

MeasurementLink C# services do not ignore registration errors.

MeasurementLink LabVIEW services used to keep running after a registration error, but we fixed it: https://github.com/ni/measurementlink-labview/issues/194

> I guess the scenario would be pretty uncommon in that the DS is not installed or not installed correctly on the machine?

Registration errors should be uncommon, but I think we should surface them to the user to the extent possible.

> I've always thought the service could still have at least some value running on its own independent of the DS, so I would have thought the logging would be enough, especially since we have the behavior of auto-starting the DS now. For example, if it's got a hard-coded port or reports the port it's on some way, a client could still connect without the DS.

This Python package doesn't support configuring a hard-coded port.

Without the discovery service, the MeasurementLink clients in InstrumentStudio and TestStand cannot connect to the service.

We would have to revisit this to support hosting services in containers, but it's not supported today.

dixonjoel (approved):

ni_measurementlink_service/_internal/service_manager.py
```diff
@@ -53,7 +53,7 @@ def start(
         configuration_parameter_list: List[ParameterMetadata],
         output_parameter_list: List[ParameterMetadata],
         measure_function: Callable,
-    ) -> str:
+    ):
```
vigkre, line 56:
```suggestion
    ) -> None:
```

vigkre, line 56:
```suggestion
    ) -> Union[str, None]:
```
or 

```suggestion
    ) -> str | None:
```

ni_measurementlink_service/measurement/service.py
```diff
@@ -375,14 +375,16 @@ def host_service(self) -> MeasurementService:
         """
         if self.measure_function is None:
             raise Exception("Error, must register measurement method.")
-        self.grpc_service.start(
+        is_service_started = self.grpc_service.start(
```
vigkre, line 378: `start` function returns `str` or `None`. 

`is_service_started` doesn't seem to fit in here.. Can we rename to `port_number` or `port` itself

tests/integration/test_service_manager.py
```diff
@@ -71,8 +92,13 @@ def discovery_client(discovery_service_stub: FakeDiscoveryServiceStub) -> Discov
 
 
 @pytest.fixture
-def discovery_service_stub() -> FakeDiscoveryServiceStub:
-    """Create a FakeDiscoveryServiceStub."""
+def discovery_service_stub(request) -> FakeDiscoveryServiceStub:
```
vigkre, line 95: type hint can be updated accordingly.

ni_measurementlink_service/_internal/service_manager.py
MounikaBattu17: Since `str | None` is supported only in >=3.10 python versions. I have used `Union[str, None]`.

mshafer-NI: Why not `Optional[str]`?

vigkre (approved):
Approved with suggestions

ni_measurementlink_service/_internal/service_manager.py
MounikaBattu17: Yeah! Seems like `Optional` is more used than `Union`. I have updated to `Optional`.

bkeryan: FYI, you can use this technique in older versions if you add `from __future__ import annotations` or you put the type hint in quotes like `) -> "str | None":`. However, `Optional[str]` is fine and doesn't require bending the rules. :)

ni_measurementlink_service/measurement/service.py
```diff
@@ -375,14 +375,16 @@ def host_service(self) -> MeasurementService:
...
             self.measurement_info,
             self.service_info,
             self.configuration_parameter_list,
             self.output_parameter_list,
             self.measure_function,
         )
-        return self
+        if port:
+            return self
+        raise Exception("Error in starting the measurement service.")
```
bkeryan, line 387: This doesn't solve the root problem, which is that DiscoveryClient does not report the error to the caller. As a result, this exception contains a generic error rather than the specific error encountered by the DiscoveryClient. It is also not linked to the original exception.

Note that I also want to pull DiscoveryClient out into a separate package, which may require fixing other issues with its interface:
- register_measurement_service uses measurement-specific types
- register_measurement_service stores the registration_id as a member variable, which assumes it is only used to register one service
- register_measurement_service function name is not symmetrical with unregister_service
  - we could deprecate register_measurement_service and add a register_service method

I don't think you should address all of that in this PR. Consider keeping the existing interface and function signatures, but change the behavior to either return True or raise an exception.

In cases where we log specific messages for specific errors, perhaps we should throw an exception with the specific message and use `raise Exception("...") from e` to link this exception to the original exception.

pyproject.toml
```diff
@@ -70,6 +70,10 @@ build-backend = "poetry.core.masonry.api"
 addopts = "--doctest-modules --strict-markers"
 filterwarnings = ["always::ImportWarning", "always::ResourceWarning"]
 testpaths = ["tests"]
+markers = [
+  # Defines custom markers used by measurement link python tests. Prevents PytestUnknownMarkWarning.  
+  "expect_discovery_service_stub_error: boolean specifying that the fake discovery service stub is valid or errored."
```
bkeryan, line 75: I had trouble understanding this marker based on the name and description. It is not clear that this marker causes the fixture to return errors.

tests/integration/test_service_manager.py
```diff
@@ -43,6 +46,24 @@ def test___grpc_service_without_discovery_service___start_service___service_host
     _validate_if_service_running_by_making_rpc(port_number)
 
 
+@pytest.mark.expect_discovery_service_stub_error(True)
```
bkeryan, line 49: Is it necessary to use a marker for this?

Can direct test parameterization work for this test case? https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#override-a-fixture-with-direct-test-parametrization

bkeryan, line 49: If we keep this marker, we should use its presence or absence as a flag instead of specifying a bool parameter.

tests/integration/test_service_manager.py
```diff
@@ -71,9 +92,16 @@ def discovery_client(discovery_service_stub: FakeDiscoveryServiceStub) -> Discov
 
 
 @pytest.fixture
-def discovery_service_stub() -> FakeDiscoveryServiceStub:
-    """Create a FakeDiscoveryServiceStub."""
-    return FakeDiscoveryServiceStub()
+def discovery_service_stub(request) -> FakeDiscoveryServiceStub:
```
bkeryan, line 95: In the context of gRPC, `def discovery_service_stub(request):` looks like you are passing a gRPC request to the stub constructor, which doesn't make any sense. 

Please add the correct type hint, `pytest.FixtureRequest`. This makes it clear what "request" is referring to, and also enables type-checking and autocomplete.

ni_measurementlink_service/_internal/service_manager.py
```diff
@@ -53,7 +53,7 @@ def start(
         configuration_parameter_list: List[ParameterMetadata],
         output_parameter_list: List[ParameterMetadata],
         measure_function: Callable,
-    ) -> str:
+    ) -> Optional[str]:
```
bkeryan, line 56: This is a step in the wrong direction. If start() fails, it should throw an exception.

bkeryan (changes requested):

tests/integration/test_service_manager.py
```diff
@@ -84,3 +112,13 @@ def _validate_if_service_running_by_making_rpc(port_number):
     with grpc.insecure_channel("localhost:" + port_number) as channel:
         stub = measurement_service_pb2_grpc.MeasurementServiceStub(channel)
         stub.GetMetadata(measurement_service_pb2.GetMetadataRequest())  # RPC call
+
+
+def _get_marker_value(request, marker_name, default=None):
```
bkeryan, line 117: type hints

tests/integration/test_service_manager.py
MounikaBattu17: Thanks for the approach! Direct fixture parameterization seems to be easy approach. I have added a new fixture `expect_discovery_service_error_stub` which returns a boolean `False` as default, thus choosing `FakeDiscoveryServiceStub` as default. For that one test case which needs `FakeDiscoveryServiceStubError`, overridden the default value with `True`.

ni_measurementlink_service/measurement/service.py
MounikaBattu17: @bkeryan  I have made the following changes
1) Removed except block's return False and replaced with raise Exceptions with type and message.
2) Updated to callers (`service.py`, `service_manager.py`) to catch the exception using `try except` block.
3) Updates `test_service.py`, `test_service_manager.py`, `test_discovery_client.py`.

pyproject.toml
MounikaBattu17: Removed this marker since I have changed to direct test parameterization approach.

tests/integration/test_service_manager.py
MounikaBattu17: Removed this marker since I have changed to direct test parameterization approach.

ni_measurementlink_service/_internal/service_manager.py
MounikaBattu17: I have changed back to `str` because instead of returning `None`, I've updated to `raise` the caught exception from called methods.

tests/integration/test_service_manager.py
MounikaBattu17: Removed this method since it is no longer needed.

tests/integration/test_service_manager.py
MounikaBattu17: Removed this marker since I have changed to direct test parameterization approach.

vigkre (approved):

ni_measurementlink_service/_internal/discovery_client.py
```diff
@@ -132,15 +132,15 @@ def register_measurement_service(
...
             _logger.error(
                 "Unable to register with discovery service. Possible reason: discovery service not running."
             )
-            return False
-        except Exception:
+            raise FileNotFoundError(e.strerror, e.filename) from e
+        except Exception as e:
             _logger.exception("Error in registering with discovery service.")
-            return False
+            raise Exception(type(e).__name__, e.args[0]) from e
```
bkeryan, line 143: To re-raise the current exception, write `raise` with no arguments.

Creating a new exception instance like this throws away information (derived type and backtrace), makes the error messages harder to understand, and is generally more complicated than it needs to be.

ni_measurementlink_service/_internal/service_manager.py
```diff
@@ -73,41 +73,48 @@ def start(
...
+            port = str(self.server.add_insecure_port("[::]:0"))
+            self.server.start()
+            _logger.info("Measurement service hosted on port: %s", port)
+            self.discovery_client.register_measurement_service(port, service_info, measurement_info)
+
+            self.port = port
+
+            return port
+        except Exception as e:
+            raise e
```
bkeryan, line 111: I don't understand what the try/except blocks in this file are intended to accomplish. You don't need to catch the exception and re-raise it.

tests/integration/test_service_manager.py
```diff
@@ -43,6 +46,20 @@ def test___grpc_service_without_discovery_service___start_service___service_host
     _validate_if_service_running_by_making_rpc(port_number)
 
 
+@pytest.mark.parametrize("expect_discovery_service_error_stub", [True])
+def test___grpc_service___start_service_error_registering_measurement___raises_error(
+    grpc_service: GrpcService,
+):
+    with pytest.raises(Exception):
```
bkeryan, line 53: The fake stub's use of `Exception` hides the fact that your code throws away the exception's derived type.

Please add this to `fake_discovery_service.py`, update the fake stub to use it, and update this test to expect the new exception type. This will demonstrate some of the problems with your proposed changes.

```
class FakeDiscoveryError(Exception):
    pass
```

bkeryan (changes requested):

tests/integration/test_service_manager.py
MounikaBattu17: Thanks for this approach! I have added the custom exception `FakeDiscoveryError` and expected it in test.

ni_measurementlink_service/_internal/service_manager.py
MounikaBattu17: Removed the `try`/`except` block.

ni_measurementlink_service/_internal/discovery_client.py
MounikaBattu17: Got it! I have only replace return False with raise. The sample exception is like this
```
2023-08-21 23:34:26,302 ERROR: Unable to register with discovery service. Possible reason: discovery service not running.
  File "Python\Python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "Python\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File ".vscode\extensions\ms-python.python-2023.4.1\pythonFiles\lib\python\debugpy\adapter/../..\debugpy\launcher/../..\debugpy\__main__.py", line 39, in <module>
    cli.main()
  File ".vscode\extensions\ms-python.python-2023.4.1\pythonFiles\lib\python\debugpy\adapter/../..\debugpy\launcher/../..\debugpy/..\debugpy\server\cli.py", line 430, in main
    run()
  File ".vscode\extensions\ms-python.python-2023.4.1\pythonFiles\lib\python\debugpy\adapter/../..\debugpy\launcher/../..\debugpy/..\debugpy\server\cli.py", line 284, in run_file
    runpy.run_path(target, run_name="__main__")
  File ".vscode\extensions\ms-python.python-2023.4.1\pythonFiles\lib\python\debugpy\_vendored\pydevd\_pydevd_bundle\pydevd_runpy.py", line 321, in run_path
    return _run_module_code(code, init_globals, run_name,
  File ".vscode\extensions\ms-python.python-2023.4.1\pythonFiles\lib\python\debugpy\_vendored\pydevd\_pydevd_bundle\pydevd_runpy.py", line 135, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File ".vscode\extensions\ms-python.python-2023.4.1\pythonFiles\lib\python\debugpy\_vendored\pydevd\_pydevd_bundle\pydevd_runpy.py", line 124, in _run_code
    exec(code, run_globals)
  File "measurementlink-python\examples\sample_measurement\measurement.py", line 113, in 
<module>
    main()
  File "measurementlink-python\examples\sample_measurement\.venv\lib\site-packages\click\core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "measurementlink-python\examples\sample_measurement\.venv\lib\site-packages\click\core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "measurementlink-python\examples\sample_measurement\.venv\lib\site-packages\click\core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "measurementlink-python\examples\sample_measurement\.venv\lib\site-packages\click\core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "measurementlink-python\examples\sample_measurement\measurement.py", line 108, in 
main
    with sample_measurement_service.host_service():
  File "measurementlink-python\ni_measurementlink_service\measurement\service.py", line 379, in host_service
    self.grpc_service.start(
  File "measurementlink-python\ni_measurementlink_service\_internal\service_manager.py", 
line 105, in start
    self.discovery_client.register_measurement_service(port, service_info, measurement_info)
  File "measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 125, in register_measurement_service
    register_response = self.stub.RegisterService(request)
  File "measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 84, in stub
    address = _get_discovery_service_address()
  File "measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 220, in _get_discovery_service_address
    _ensure_discovery_service_started(key_file_path)
  File "measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 232, in _ensure_discovery_service_started
    exe_file_path = _get_discovery_service_location()
  File "measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 240, in _get_discovery_service_location
    registration_json_obj = json.loads(registration_json_path.read_text())
  File "Python\Python38\lib\pathlib.py", line 1236, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors) as f:
  File "Python\Python38\lib\pathlib.py", line 1222, in open
    return io.open(self, mode, buffering, encoding, errors, newline,
  File "Python\Python38\lib\pathlib.py", line 1078, in _opener
    return self._accessor.open(self, flags, mode)
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Program Files\\National Instruments\\Shared\\MeasurementLink\\MeasurementLinkService.json'
```

bkeryan (approved):

PR #354, opened 2023-08-16
Refactor DataType enum so it doesn't contain all of the data type information

dixonjoel:

Refactor DataType. It is currently defined as an enum, but the enum values are unnamed tuples of the grpc field type, a repeated boolean, and the TypeSpecialization. This refactoring makes `DataType` a simple enum and adds a `NamedTuple` to hold that information. It also adds a static lookup method to get the information from the `DataType` enum.

Simplifies the `DataType` enum and avoids publicly exposing all the information in the enum tuples. Also adds named fields for the information related to data types so the code is clearer about the fields.

Ran all tests locally. No new tests since this is a straight refactoring.


ni_measurementlink_service/measurement/info.py
```diff
@@ -67,26 +67,104 @@ class TypeSpecialization(enum.Enum):
...
-    UInt32Array1D = (type_pb2.Field.TYPE_UINT32, True, TypeSpecialization.NoType)
-    UInt64Array1D = (type_pb2.Field.TYPE_UINT64, True, TypeSpecialization.NoType)
-    FloatArray1D = (type_pb2.Field.TYPE_FLOAT, True, TypeSpecialization.NoType)
-    DoubleArray1D = (type_pb2.Field.TYPE_DOUBLE, True, TypeSpecialization.NoType)
-    BooleanArray1D = (type_pb2.Field.TYPE_BOOL, True, TypeSpecialization.NoType)
-    StringArray1D = (type_pb2.Field.TYPE_STRING, True, TypeSpecialization.NoType)
-    PinArray1D = (type_pb2.Field.TYPE_STRING, True, TypeSpecialization.Pin)
-    PathArray1D = (type_pb2.Field.TYPE_STRING, True, TypeSpecialization.Path)
-    EnumArray1D = (type_pb2.Field.TYPE_ENUM, True, TypeSpecialization.Enum)
+    Int32 = (0,)
```
dixonjoel, line 70: Hmm, I think the ni-python-styleguide did this. Not sure why.

ni_measurementlink_service/measurement/info.py
```diff
@@ -67,26 +67,107 @@ class TypeSpecialization(enum.Enum):
...
+    FloatArray1D = 104
+    DoubleArray1D = 105
+    BooleanArray1D = 106
+    StringArray1D = 107
+    PinArray1D = 108
+    PathArray1D = 109
+    EnumArray1D = 110
+
+
+class DataTypeInfo(NamedTuple):
```
bkeryan, line 95: I don't think these implementation details should be public. Please move them to a private module (one that starts with an underscore). Suggestion: something like `ni_measurementlink_service/_datatypeinfo.py`

ni_measurementlink_service/measurement/info.py
```diff
@@ -67,26 +67,107 @@ class TypeSpecialization(enum.Enum):
...
+FloatArray1DTypeInfo = DataTypeInfo(type_pb2.Field.TYPE_FLOAT, True)
+DoubleArray1DTypeInfo = DataTypeInfo(type_pb2.Field.TYPE_DOUBLE, True)
+BooleanArray1DTypeInfo = DataTypeInfo(type_pb2.Field.TYPE_BOOL, True)
+StringArray1DTypeInfo = DataTypeInfo(type_pb2.Field.TYPE_STRING, True)
+PinArray1DTypeInfo = DataTypeInfo(type_pb2.Field.TYPE_STRING, True, TypeSpecialization.Pin)
+PathArray1DTypeInfo = DataTypeInfo(type_pb2.Field.TYPE_STRING, True, TypeSpecialization.Path)
+EnumArray1DTypeInfo = DataTypeInfo(type_pb2.Field.TYPE_ENUM, True, TypeSpecialization.Enum)
+
+
+class DataTypeInfoLookup:
```
bkeryan, line 139: I don't think it's necessary for the lookup to have a separate class.

Options:
- Define the lookup function at file scope
- Define the lookup function as a static method of the DataTypeInfo class

ni_measurementlink_service/measurement/info.py
```diff
@@ -67,26 +67,107 @@ class TypeSpecialization(enum.Enum):
...
+        DataType.PathArray1D: PathArray1DTypeInfo,
+        DataType.EnumArray1D: EnumArray1DTypeInfo,
+    }
+
+    @staticmethod
+    def get_type_info(data_type: DataType) -> DataTypeInfo:
+        """Get information about a DataType."""
+        if data_type not in DataTypeInfoLookup._DATATYPE_TO_DATATYPEINFO_LOOKUP:
+            raise Exception(f"Data type information not found '{data_type}'")
+        data_type_info = DataTypeInfoLookup._DATATYPE_TO_DATATYPEINFO_LOOKUP[data_type]
```
bkeryan, line 172: This does two dictionary lookups when it only needs one.

```suggestion
        data_type_info = DataTypeInfoLookup._DATATYPE_TO_DATATYPEINFO_LOOKUP.get(data_type)
        if data_type_info is None:
            raise Exception(f"Data type information not found '{data_type}'")
```

I'm not really worried about performance because this lookup is only done at initialization time, but IMO using get() is a "best practice" and is also simpler than doing two lookups.

ni_measurementlink_service/measurement/info.py
bkeryan, line 1: This change technically breaks compatibility, so I guess we should label it accordingly. However, I don't expect any clients to be affected.

bkeryan (changes requested):

ni_measurementlink_service/measurement/info.py
```diff
@@ -67,26 +67,107 @@ class TypeSpecialization(enum.Enum):
...
+        can have more than one use
+
+    """
+
+    grpc_field_type: type_pb2.Field.Kind.ValueType
+    repeated: bool
+    type_specialization: TypeSpecialization = TypeSpecialization.NoType
+
+
+Int32TypeInfo = DataTypeInfo(type_pb2.Field.TYPE_INT32, False)
```
bkeryan, line 114: These are constants, not types, so they should be in UPPER_SNAKE_CASE

https://pep8.org/#constants

bkeryan: This also applies to the enum values but it's too late to fix those

bkeryan: Also, these constants aren't accessed from outside this module, so you can make them private. Or you can delete them and move the data into the dictionary initializer.

dixonjoel: I'm going to delete these and use the dictionary initializer, but how do you make a constant private in Python?

dixonjoel: Just put them in a private module?

ni_measurementlink_service/measurement/info.py
dixonjoel: Marked with the 'breaking change' tag.

ni_measurementlink_service/measurement/info.py
bkeryan: It still does two lookups. Delete the 2nd lookup.

ni_measurementlink_service/_datatypeinfo.py
```diff
@@ -0,0 +1,59 @@
...
+    """Class that represents the information for each of the DataType enum values.
+
+    Attributes
+    ----------
+        grpc_field_type (type_pb2.Field.Kind): Field.Kind associated with the DataType
+
+        repeated (bool): Whether the DataType is a repeated field
+
+        type_specialization (TypeSpecialization): Specific type when value_type
+        can have more than one use
```
bkeryan, line 18: FYI, Sphinx uses type hints, so listing the attribute types in the class docstring is redundant.

Also FYI, you can write docstrings for class attributes: https://stackoverflow.com/questions/6060813/how-to-document-fields-and-properties-in-python

ni_measurementlink_service/measurement/info.py
dixonjoel: Oops. Fixed.

ni_measurementlink_service/measurement/info.py
bkeryan: Use a leading underscore.
https://pep8.org/#public-and-internal-interfaces

bkeryan (approved):

PR #358, opened 2023-08-17
Add encoder and decoder for DoubleXYData

dixonjoel:

Add encoder / decoder for DoubleXYData. This is the last step for supporting DoubleXYData with a follow-on PR coming that has an example (Game of Life) to showcase / exercise this capability.

Supports measurement services using the DoubleXYData type.

Adding automated tests for encoding / decoding. Expanded test.proto to include DoubleXYData and tested with that.


ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -54,6 +56,53 @@ def vector_encoder(field_index):
...
+
+
+def _encode_varint(write, value, unused_deterministic=None):
+    bits = value & 0x7F
+    value >>= 7
+    while value:
+        write(local_int2byte(0x80 | bits))
+        bits = value & 0x7F
+        value >>= 7
+    return write(local_int2byte(bits))
```
bkeryan, line 69: This is copied from google.protobuf. 

1. Do we really need to copy it?
2. If we really need to copy it, we should keep it in a separate module with a comment about its origin.
3. Can we copy it without restructuring it as much?

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -54,6 +56,53 @@ def vector_encoder(field_index):
     return vector_encoder
 
 
+local_int2byte = struct.Struct(">B").pack
```
bkeryan, line 59: This was local in the original code, but it's not local here. Also, it's public for no reason.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -54,6 +56,53 @@ def vector_encoder(field_index):
...
+    value >>= 7
+    while value:
+        write(local_int2byte(0x80 | bits))
+        bits = value & 0x7F
+        value >>= 7
+    return write(local_int2byte(bits))
+
+
+def _inner_message_encoder(field_index):
+    """Mimics google.protobuf._internal.MessageEncoder."""
```
bkeryan, line 73: google.protobuf.internal doesn't have an underscore.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -154,6 +272,8 @@ class Context:
         type_pb2.Field.TYPE_BOOL: (BoolEncoder, BoolArrayEncoder),
         type_pb2.Field.TYPE_STRING: (StringEncoder, StringArrayEncoder),
         type_pb2.Field.TYPE_ENUM: (IntEncoder, IntArrayEncoder),
+        # Array encoder is not correct, but 'None' violates expected return value
```
bkeryan, line 275: Instead of using the wrong encoder, please write an encoder that returns an error.

```
def _unsupported_encoder(field_number, is_repeated, is_packed):
    raise NotImplementedError(f"Unsupported data type for field {field_number}")
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -229,6 +349,12 @@ def get_decoder(
...
 
         """
+        if message_type == xydata_pb2.DoubleXYData.DESCRIPTOR.full_name:
+            if type != type_pb2.Field.Kind.TYPE_MESSAGE:
+                raise Exception(f"Message type must have a TYPE_MESSAGE kind '{message_type}'")
+            if repeated:
+                raise Exception(f"Repeated message types are not supported '{message_type}'")
+            return XYDataDecoder
         if type not in Context._FIELD_TYPE_TO_DECODER_MAPPING:
             raise Exception(f"Error can not decode type '{type}'")
```
bkeryan, line 359: Something like ValueError or TypeError would be more appropriate.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -107,12 +156,80 @@ def vector_decoder(field_index, key):
...
+        b = buffer[pos]
+        result |= (b & 0x7F) << shift
+        pos += 1
+        if not (b & 0x80):
+            result &= mask
+            result = result_type(result)
+            return (result, pos)
+        shift += 7
+        if shift >= 64:
+            raise Exception("Too many bytes when decoding varint.")
```
bkeryan, line 174: This was a _DecodeError in the original code. Can we use something more descriptive than Exception, like ValueError?

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -107,12 +156,80 @@ def vector_decoder(field_index, key):
...
+        (size, pos) = _decode_varint(buffer, pos)
+        new_pos = pos + size
+        thestring = _convert_to_byte_string(buffer[pos:new_pos])
+        value.ParseFromString(thestring)
+        return new_pos
+
+    return decode_message
+
+
+def _double_xy_data_decoder(decoder) -> Callable[[int, str], Callable]:
```
bkeryan, line 199: The type hint doesn't describe the signature of the returned callable. Creating some type aliases might make this code more understandable.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -107,12 +156,80 @@ def vector_decoder(field_index, key):
...
+    """Decoder for DoubleXYData.
+
+    Args
+    ----
+        decoder (Callable[[int, bool, bool], Callable]): Specific decoder(Callable) that takes in
```
bkeryan, line 204: Don't duplicate type hints in the docstring. Sphinx can parse type hints.

You don't really need to write docstrings for private/internal stuff. Sphinx doesn't generate any docs for these functions. If you do write docstrings for private/internal stuff, please try to avoid burying the content in boilerplate.

bkeryan (changes requested):

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -107,12 +156,80 @@ def vector_decoder(field_index, key):
...
+        if not (b & 0x80):
+            result &= mask
+            result = result_type(result)
+            return (result, pos)
+        shift += 7
+        if shift >= 64:
+            raise Exception("Too many bytes when decoding varint.")
+
+
+def _inner_message_decoder(field_index, is_repeated, is_packed, key, new_default):
```
bkeryan, line 177: type hints would help explain what this stuff is doing

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
dixonjoel: Moved to _message.py with comments about where they came from.

bkeryan: 3. I meant copy `_VarintEncoder` so what's local stays local

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
dixonjoel: Added type hints for Encoder / Decoder factory and low level encoder / decoder.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,23 +1,34 @@
...
+LowLevelEncoder = Callable[[Callable[[Buffer], int], Any, None], None]
+EncoderFactory = Callable[[int], LowLevelEncoder]
+LowLevelDecoder = Callable[
+    [memoryview, int, int, type_pb2.Field.Kind.ValueType, Dict[FieldDescriptor, Any]], int
+]
+DecoderFactory = Callable[[int, str], LowLevelDecoder]
+
+
+def _scalar_encoder(encoder) -> EncoderFactory:
     """Abstract Specific Encoder(Callable) as Scalar Encoder Callable that takes in field index.
```
bkeryan, line 22: I think the docstrings in this file do more harm than good. They are overly verbose, surface-level descriptions of the functions, and they are Littered with distracting Random Caps.

What does `_scalar_encoder` do? It performs partial application of is_repeated=False is_packed=False.

What does `_vector_encoder` do? It performs partial application of is_repeated=True and makes is_packed optional.

Do we need to list the arguments and their types in tedious Google docstring format? I'd rather have type hints in the function signature.

PEP8 says docstrings are not required for non-public methods. https://pep8.org/#documentation-strings

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,23 +1,34 @@
...
-from google.protobuf.internal import encoder
+from google.protobuf.descriptor import FieldDescriptor
+from google.protobuf.internal import decoder, encoder
+from typing_extensions import Buffer
 
+from ni_measurementlink_service._internal.parameter import _message
+from ni_measurementlink_service._internal.stubs.ni.protobuf.types import xydata_pb2
 
-def _scalar_encoder(encoder) -> Callable[[int], Callable]:
+LowLevelEncoder = Callable[[Callable[[Buffer], int], Any, None], None]
```
bkeryan, line 13: `LowLevelEncoder` and `LowLevelDecoder` look suspiciously like `google.protobuf.internal.encoder._Encoder` and `google.protobuf.internal.decoder._Decoder`: https://github.com/python/typeshed/blob/main/stubs/protobuf/google/protobuf/internal/decoder.pyi

Are they the same as the types in typeshed? Why are they different?

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,23 +1,34 @@
...
+EncoderFactory = Callable[[int], LowLevelEncoder]
```
bkeryan, line 14: `EncoderFactory` and `DecoderFactory` describe our partially-applied versions of `google.protobuf.internal.decoder.Int32Decoder`, etc. I'm not 100% onboard with these names, but I don't have a better suggestion yet.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,23 +1,34 @@
...
+LowLevelDecoder = Callable[
+    [memoryview, int, int, type_pb2.Field.Kind.ValueType, Dict[FieldDescriptor, Any]], int
+]
+DecoderFactory = Callable[[int, str], LowLevelDecoder]
+
+
+def _scalar_encoder(encoder) -> EncoderFactory:
```
bkeryan, line 21: It would be helpful to annotate the parameter types, particularly `encoder`, which could be anything.

Also nested functions.

bkeryan, line 21: This file imports `encoder` and `decoder` and then shadows them with parameters. We should rename one or the other:
- We could `import google.protobuf.internal.decoder as _decoder`
- Or we could rename the parameter to something like `inner_decoder` or `decoder_factory` (if that's what it is; it's the same as our `DecoderFactory` with more parameters)

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -54,7 +65,32 @@ def vector_encoder(field_index):
...
+
+    """
+
+    def message_encoder(field_index):
+        return encoder(field_index)
+
+    return message_encoder
+
+
+def _unsupported_encoder(field_index, is_repeated, is_packed):
```
bkeryan, line 89:
```suggestion
def _unsupported_encoder(field_index: int, is_repeated: bool, is_packed: bool):
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -229,6 +295,12 @@ def get_decoder(
             Callable[[int], Callable]: ScalarDecoder or VectorDecoder.
 
         """
+        if message_type == xydata_pb2.DoubleXYData.DESCRIPTOR.full_name:
```
bkeryan, line 298: I think this is called for each request, so we should move the message type comparisons (of which there will be more in the future) out of the hot path.

Example:
```
decoder_mapping = Context._FIELD_TYPE_TO_DECODER_MAPPING.get(type)
if decoder_mapping is not None:
    scalar_decoder, array_decoder = decoder_mapping
    return array_decoder if repeated else scalar_decoder
elif message_type == blah blah blah:
elif message_type == blah blah blah2:
elif message_type == blah blah blah3:
else:
    raise Exception(f"Error can not decode type '{type}'")
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -183,7 +249,7 @@ class Context:
     @staticmethod
     def get_encoder(
         type: type_pb2.Field.Kind.ValueType, repeated: bool
-    ) -> Callable[[int], Callable]:
+    ) -> Callable[[int], LowLevelEncoder]:
```
bkeryan, line 252:
```suggestion
    ) -> EncoderFactory:
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -210,7 +276,7 @@ def get_encoder(
 
     @staticmethod
     def get_decoder(
-        type: type_pb2.Field.Kind.ValueType, repeated: bool
+        type: type_pb2.Field.Kind.ValueType, repeated: bool, message_type: str = ""
     ) -> Callable[[int, str], Callable]:
```
bkeryan, line 280:
```suggestion
    ) -> DecoderFactory:
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -229,6 +295,12 @@ def get_decoder(
...
 
         """
+        if message_type == xydata_pb2.DoubleXYData.DESCRIPTOR.full_name:
+            if type != type_pb2.Field.Kind.TYPE_MESSAGE:
+                raise ValueError(f"Message type must have a TYPE_MESSAGE kind '{message_type}'")
+            if repeated:
+                raise ValueError(f"Repeated message types are not supported '{message_type}'")
+            return XYDataDecoder
         if type not in Context._FIELD_TYPE_TO_DECODER_MAPPING:
             raise Exception(f"Error can not decode type '{type}'")
```
bkeryan, line 305: Seems like `TypeError` or `ValueError`

ni_measurementlink_service/_internal/parameter/serializer.py
```diff
@@ -158,7 +158,9 @@ def _get_overlapping_parameters(
             )
         type = parameter_metadata_dict[field_index].type
         is_repeated = parameter_metadata_dict[field_index].repeated
-        decoder = serialization_strategy.Context.get_decoder(type, is_repeated)
+        decoder = serialization_strategy.Context.get_decoder(
+            type, is_repeated, parameter_metadata_dict[field_index].message_type
+        )
```
bkeryan, line 163: One of these things is not like the others. (Two are assigned to local variables and the third is not.)

`field_metadata = parameter_metadata_dict[field_index]` would eliminate the duplicated lookups.

ni_measurementlink_service/_internal/parameter/serializer.py
```diff
@@ -158,7 +158,9 @@ def _get_overlapping_parameters(
...
         inner_decoder = decoder(field_index, field_index)
```
bkeryan, line 164: The 2nd parameter is `key`, which is typed as `str` in `serialization_strategy.py`. 

Here, we are passing an `int`.

typeshed annotates `key` as a `FieldDescriptor`. 

Are these the same thing? Which is right?

ni_measurementlink_service/_internal/parameter/serializer.py
```diff
@@ -158,7 +158,9 @@ def _get_overlapping_parameters(
...
         parameter_bytes_io = BytesIO(parameter_bytes)
         parameter_bytes_memory_view = parameter_bytes_io.getbuffer()
```
bkeryan, line 166: I am not sure what type this should be (`getbuffer()` is a memory view, `getvalue()` is bytes) or whether the typeshed for `google.protobuf.internal.decoder._Decoder` is correct (it uses `str`?)

ni_measurementlink_service/_internal/parameter/_message.py
```diff
@@ -0,0 +1,94 @@
...
+    return encode_varint
+
+
+def _inner_message_decoder(field_index, is_repeated, is_packed, key, new_default):
+    """Based on google.protobuf.internal.MessageDecoder.
+
+    See DecodeField
+    """
+
+    def _convert_to_byte_string(memview):
```
bkeryan, line 49: Is this here because we are calling `getbuffer()` instead of `getvalue()`?

ni_measurementlink_service/_internal/parameter/_message.py
```diff
@@ -0,0 +1,94 @@
...
+            write(local_int2byte(0x80 | bits))
+            bits = value & 0x7F
+            value >>= 7
+        return write(local_int2byte(bits))
+
+    return encode_varint
+
+
+def _inner_message_decoder(field_index, is_repeated, is_packed, key, new_default):
+    """Based on google.protobuf.internal.MessageDecoder.
```
bkeryan, line 44: Is the point of this function to call `ParseFromString` instead of `_InternalParse`? That would be relevant information to document.

Ditto for the encoder and `_InternalSerialize`.

I would still like to know why this is the case. Is it due to upb vs. the old pure-Python parser? If you set the environment variable to use the old pure-Python parser, does it define `_InternalParse` and `_InternalSerialize`?

ni_measurementlink_service/_internal/parameter/_message.py
```diff
@@ -0,0 +1,94 @@
...
+        return byte_str
+
+    def _decode_message(buffer, pos, end, message, field_dict):
+        decode_varint = _varint_decoder(mask=(1 << 64) - 1, result_type=int)
+        value = field_dict.get(key)
+        if value is None:
+            value = field_dict.setdefault(key, new_default(message))
+        # Read length.
+        (size, pos) = decode_varint(buffer, pos)
+        new_pos = pos + size
```
bkeryan, line 60: You removed two error checks:
- Bounds check
- Did parsing consume the entire buffer subset?

ni_measurementlink_service/_internal/parameter/_message.py
```diff
@@ -0,0 +1,94 @@
+import struct
+
+from google.protobuf.internal import encoder, wire_format
+
+
+def _inner_message_encoder(field_index):
```
bkeryan, line 6: You customized the signature by removing is_repeated and is_packed, which duplicates the work of `_scalar_encoder` and `_vector_encoder`.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,23 +1,34 @@
...
 
-from typing import Any, Callable
+from typing import Any, Callable, Dict
 
 from google.protobuf import type_pb2
-from google.protobuf.internal import decoder
-from google.protobuf.internal import encoder
+from google.protobuf.descriptor import FieldDescriptor
+from google.protobuf.internal import decoder, encoder
+from typing_extensions import Buffer
```
bkeryan, line 8: Let's not require `typing_extensions` at runtime.

```
...
if typing.TYPE_CHECKING:
    from typing_extensions import Buffer
```

In type aliases, you can quote types that are not available at runtime.
```
LowLevelEncoder = Callable[[Callable[["Buffer"], int], Any, None], None]
```

bkeryan (changes requested):
At what point do we split out the refactoring/cleanup into a separate PR and rebase this one on top of it?

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -154,6 +219,7 @@ class Context:
         type_pb2.Field.TYPE_BOOL: (BoolEncoder, BoolArrayEncoder),
```
bkeryan, line 219: I'm not sure what purpose this `Context` class serves. It seems to have only class data/methods, not instance data/methods. This is not C# or Java, so we don't need to wrap all data/methods in classes.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
bkeryan: Also, a 2nd dict of message_type -> decoder would make sense.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
dixonjoel: Should be resolved by the refactoring PR.

ni_measurementlink_service/_internal/parameter/_message.py
dixonjoel: Unneeded. I think I may have gotten this from StringDecoder, but I don't remember.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
dixonjoel: Do you still feel we want to rename this to, e.g. decoder_constructor, or is it clear enough with the type hint next to it? I think it's clear as is.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
dixonjoel: Cover4ed in the refactoring PR

ni_measurementlink_service/_internal/parameter/_message.py
dixonjoel: Added comments to the method.

ni_measurementlink_service/_internal/parameter/_message.py
dixonjoel: Added back in the checks. ParseFromString is not like _InternalSerialize and seems to return the size of bytes it parsed. I can't find where ParseFromString is in the code to confirm this.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
dixonjoel: Deferring the addition of a message_type -> decoder dictionary until more messages are added.

ni_measurementlink_service/_internal/parameter/_message.py
```diff
@@ -0,0 +1,113 @@
...
+    Key,
+    NewDefault,
+)
+
+
+def _message_encoder_constructor(field_index: int, is_repeated: bool, is_packed: bool) -> Encoder:
+    """Mimics google.protobuf.internal.MessageEncoder.
+
+    We cannot use MessageEncoder because it is for the pure-Python implementation
+    of messages. Our messages are upb messages, so we need to call 'SerializeToString'
```
bkeryan, line 19: > Our messages are upb messages

It's not up to us.

Also, "our messages" is confusing. I think you are refering to DoubleXY and not the configuration/output parameters?

```suggestion
    This function was forked in order to call SerializeToString instead of _InternalSerialize.
    
    _InternalSerialize is only defined for the pure-Python protobuf implementation. Our child messages
    (like DoubleXY) are defined in .proto files, so they use whichever protobuf implementation that
    google.protobuf.internal.api_implementation chooses (usually upb).
```

ni_measurementlink_service/_internal/parameter/_message.py
```diff
@@ -0,0 +1,113 @@
...
+    return encode_varint
+
+
+def _message_decoder_constructor(
+    field_index: int, is_repeated: bool, is_packed: bool, key: Key, new_default: NewDefault
+) -> Decoder:
+    """Mimics google.protobuf.internal.MessageDecoder.
+
+    We cannot use MessageDecoder because it is for the pure-Python implementation
+    of messages. Our messages are upb messages, so we need to call 'ParseFromString'
```
bkeryan, line 61:
```suggestion
    This function was forked in order to call ParseFromString instead of _InternalParse.
    
    _InternalParse is only defined for the pure-Python protobuf implementation. Our child messages
    (like DoubleXY) are defined in .proto files, so they use whichever protobuf implementation that
    google.protobuf.internal.api_implementation chooses (usually upb).
```
```

File: ni_measurementlink_service/_internal/parameter/_message.py [hunk 30]
```diff
@@ -0,0 +1,113 @@
...
+    We cannot use MessageEncoder because it is for the pure-Python implementation
+    of messages. Our messages are upb messages, so we need to call 'SerializeToString'
+
+    See EncodeField:
+    https://github.com/protocolbuffers/protobuf/blob/0b817d46d4ca1977d3dccf2442aeee3c9e98e3a1/python/google/protobuf/internal/encoder.py#L765C15-L765C15
+    """
+    tag = encoder.TagBytes(field_index, wire_format.WIRETYPE_LENGTH_DELIMITED)
+    encode_varint = _varint_encoder()
+
+    def _encode_message(write, value, deterministic):
```
bkeryan, line 27:
```suggestion
    def _encode_message(write: WriteFunction, value: Message, deterministic: bool):
```

ni_measurementlink_service/_internal/parameter/_message.py
```diff
@@ -0,0 +1,113 @@
...
+
+from ni_measurementlink_service._internal.parameter._serializer_types import (
+    Decoder,
+    Encoder,
+    Key,
+    NewDefault,
+)
+
+
+def _message_encoder_constructor(field_index: int, is_repeated: bool, is_packed: bool) -> Encoder:
```
bkeryan, line 15: This doesn't return an `Encoder` because `Encoder`'s 2nd parameter is `bytes` not `Message`
```suggestion
def _message_encoder_constructor(field_index: int, is_repeated: bool, is_packed: bool) -> Callable[[WriteFunction, Message, bool], int]:
```

ni_measurementlink_service/_internal/parameter/_message.py
```diff
@@ -0,0 +1,113 @@
...
+            raise ValueError("Error decoding a message. Message is truncated.")
+        parsed_bytes = value.ParseFromString(buffer[pos:new_pos])
+        if parsed_bytes != size:
+            raise ValueError("Parsed incorrect number of bytes.")
+        return new_pos
+
+    return _decode_message
+
+
+def _varint_decoder(mask, result_type):
```
bkeryan, line 86:
```suggestion
def _varint_decoder(mask: int, result_type: Type[T]) -> Callable[[memoryview, int], Tuple[T, int]]:
```
where T is a TypeVar

ni_measurementlink_service/_internal/parameter/_message.py
```diff
@@ -0,0 +1,113 @@
...
+    def _encode_message(write, value, deterministic):
+        write(tag)
+        bytes = value.SerializeToString()
+        encode_varint(write, len(bytes), deterministic)
+        write(bytes)
+
+    return _encode_message
+
+
+def _varint_encoder():
```
bkeryan, line 36:
```suggestion
def _varint_encoder() -> Callable[[WriteFunction, int, Optional[bool]], int]:
```

ni_measurementlink_service/_internal/parameter/_message.py
```diff
@@ -0,0 +1,113 @@
...
+    """Return an encoder for a basic varint value (does not include tag).
+
+    From google.protobuf.internal.encoder.py _VarintEncoder
+    """
+    local_int2byte = struct.Struct(">B").pack
+
+    def encode_varint(write, value, unused_deterministic=None):
```
bkeryan, line 43:
```suggestion
    def encode_varint(write: WriteFunction, value: int, unused_deterministic: Optional[bool] = None) -> int:
```

ni_measurementlink_service/_internal/parameter/_message.py
```diff
@@ -0,0 +1,113 @@
...
+    Decoded values will be bitwise-anded with the given mask before being
+    returned, e.g. to limit them to 32 bits.  The returned decoder does not
+    take the usual "end" parameter -- the caller is expected to do bounds checking
+    after the fact (often the caller can defer such checking until later).  The
+    decoder returns a (value, new_pos) pair.
+
+    From google.protobuf.internal.decoder.py _VarintDecoder
+    """
+
+    def decode_varint(buffer, pos):
```
bkeryan, line 98:
```suggestion
    def decode_varint(buffer: memoryview, pos: int) -> Tuple[T, int]:
```
(assuming the inner function can use the outer function's TypeVar)

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -63,6 +55,10 @@ def vector_encoder(field_index: int) -> Encoder:
     return vector_encoder
 
 
+def _unsupported_encoder(field_index: int, is_repeated: bool, is_packed: bool):
```
bkeryan, line 58:
```suggestion
def _unsupported_encoder(field_index: int, is_repeated: bool, is_packed: bool) -> Encoder:
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -106,6 +102,23 @@ def vector_decoder(field_index: int, key: Key) -> Decoder:
     return vector_decoder
 
 
+def _double_xy_data_decoder(decoder: DecoderConstructor) -> PartialDecoderConstructor:
+    """Constructs a DoubleXYData decoder constructor.
+
+    Takes a field index and a key and returns a Decoder for DoubleXYData.
+    """
+
+    def _new_default(unused_message=None):
```
bkeryan, line 111:
```suggestion
    def _new_default(unused_message: Optional[Message] = None) -> Any:
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -106,6 +102,23 @@ def vector_decoder(field_index: int, key: Key) -> Decoder:
...
+        return xydata_pb2.DoubleXYData()
+
+    def message_decoder(field_index, key):
```
bkeryan, line 114:
```suggestion
    def message_decoder(field_index: int, key; Key) -> Decoder:
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -195,17 +212,22 @@ def get_encoder(type: type_pb2.Field.Kind.ValueType, repeated: bool) -> PartialE
...
+    decoder_mapping = _FIELD_TYPE_TO_DECODER_MAPPING.get(type)
+    if decoder_mapping is not None:
+        scalar_decoder, array_decoder = decoder_mapping
+        return array_decoder if repeated else scalar_decoder
+    elif message_type == xydata_pb2.DoubleXYData.DESCRIPTOR.full_name:
+        if type != type_pb2.Field.Kind.TYPE_MESSAGE:
+            raise ValueError(f"Message type must have a TYPE_MESSAGE kind '{message_type}'")
+        if repeated:
+            raise ValueError(f"Repeated message types are not supported '{message_type}'")
+        return XYDataDecoder
```
bkeryan, line 228: Let's structure this so we can add more message types without duplicating the error checks. 

```suggestion
    elif type == type_pb2.Field.Kind.TYPE_MESSAGE:
        if repeated:
            raise ValueError(f"Repeated message types are not supported '{message_type}'")
        if message_type == xydata_pb2.DoubleXYData.DESCRIPTOR.full_name:
            return XYDataDecoder
        else:
            raise ValueError(f"Unknown message type '{message_type}'")
```

ni_measurementlink_service/_internal/parameter/serializer.py
```diff
@@ -158,9 +158,10 @@ def _get_overlapping_parameters(
             raise Exception(
                 f"Error occurred while reading the parameter - given protobuf index '{field_index}' is invalid."
             )
-        type = parameter_metadata_dict[field_index].type
-        is_repeated = parameter_metadata_dict[field_index].repeated
-        decoder = serialization_strategy.get_decoder(type, is_repeated)
+        field_metadata = parameter_metadata_dict[field_index]
+        type = field_metadata.type
+        is_repeated = field_metadata.repeated
+        decoder = serialization_strategy.get_decoder(type, is_repeated, field_metadata.message_type)
```
bkeryan, line 164: We don't use these again, do we?
```suggestion
        field_metadata = parameter_metadata_dict[field_index]
        decoder = serialization_strategy.get_decoder(field_metadata.type, field_metadata.repeated, field_metadata.message_type)
```

bkeryan (changes requested):

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
bkeryan: Even better... (didn't I already suggest this?)
```suggestion
    elif type == type_pb2.Field.Kind.TYPE_MESSAGE:
        if repeated:
            raise ValueError(f"Repeated message types are not supported '{message_type}'")
        decoder = _MESSAGE_TYPE_TO_DECODER.get(message_type)
        if decoder is None:
            raise ValueError(f"Unknown message type '{message_type}'")
        return decoder
```

dixonjoel: Yeah, you did suggest a second dictionary for _MESSAGE_TYPE_TO_DECODER. I was going to defer this until we add more message types, but I'll go ahead and add it.

ni_measurementlink_service/_internal/parameter/_message.py
```diff
@@ -1,30 +1,32 @@
...
+    implementation that google.protobuf.internal.api_implementation chooses (usually upb).
     """
     tag = encoder.TagBytes(field_index, wire_format.WIRETYPE_LENGTH_DELIMITED)
     encode_varint = _varint_encoder()
 
-    def _encode_message(write, value, deterministic):
+    def _encode_message(write: WriteFunction, value: Message, deterministic: bool):
         write(tag)
         bytes = value.SerializeToString()
         encode_varint(write, len(bytes), deterministic)
```
bkeryan, line 32: This is missing the return type hint, so it's implicitly `Any`. The `Encoder` typedef suggests that it returns an `int`, so you should return the value of `write(bytes)`.

Protobuf seems to return an int for scalar writes and None for repeated writes, so maybe the correct return type is `Optional[int]`, but we don't have to worry about that in this case.

bkeryan (approved):
Approved with suggestions

PR #360, opened 2023-08-21
Game of Life example

dixonjoel:

This adds a Game of Life example similar to the one we have in C#.

Demonstrates that XY data can be used in Python as well as C#.

Interactive testing with poetry and running the example in InstrumentStudio.


examples/game_of_life/measurement.py
```diff
@@ -0,0 +1,126 @@
...
+)
+
+
+@measurement_service.register_measurement
+@measurement_service.configuration("width", nims.DataType.UInt32, 100)
+@measurement_service.configuration("height", nims.DataType.UInt32, 100)
+@measurement_service.configuration("update_interval", nims.DataType.UInt32, 100)
+@measurement_service.output("game_of_life", nims.DataType.DoubleXYData)
+@measurement_service.output("generation", nims.DataType.UInt32)
+def measure(width, height, update_interval):
```
bkeryan, line 26: Please add type hints.

examples/game_of_life/measurement.py
```diff
@@ -0,0 +1,126 @@
...
+)
+def main(verbose: int) -> None:
+    """Host the game_of_life service."""
+    if verbose > 1:
+        level = logging.DEBUG
+    elif verbose == 1:
+        level = logging.INFO
+    else:
+        level = logging.WARNING
+    logging.basicConfig(format="%(asctime)s %(levelname)s: %(message)s", level=level)
```
bkeryan, line 119: _helpers.py

examples/game_of_life/measurement.py
```diff
@@ -0,0 +1,126 @@
...
+    xydata.y_data.append(height + 1)
+    return xydata
+
+
+def _initialize_grid_with_seeded_data(rows, cols, probability=0.6):
+    return [[random.random() < probability for _ in range(cols)] for _ in range(rows)]
+
+
+def _initialize_grid(rows, cols):
+    return [[False for _ in range(cols)] for _ in range(rows)]
```
bkeryan, line 71: FYI, I was going to suggest `[[False] * cols] * rows` but unfortunately this is wrong because it returns N references to the same row.

bkeryan (changes requested):

examples/game_of_life/_helpers.py
```diff
@@ -0,0 +1,302 @@
...
+def get_service_options(**kwargs) -> ServiceOptions:
+    """Get service options from keyword arguments."""
+    return ServiceOptions(
+        use_grpc_device=kwargs.get("use_grpc_device", False),
+        grpc_device_address=kwargs.get("grpc_device_address", ""),
+        use_simulation=kwargs.get("use_simulation", False),
+    )
+
+
+T = TypeVar("T")
```
mshafer-NI, line 51: I'm not seeing T used anywhere. Can we remove it?

dixonjoel: I also don't see it used. We have 'vendored' this file in all our examples and the template, so we have been trying to keep all of them identical. @bkeryan can we remove this from all of the _helpers.py example files and from the template? If so, I'll plan to do this in a separate PR.

bkeryan: `T` was used by `str_to_enum`. When we added support for real enums, we removed `str_to_enum` but we overlooked `T`.

dixonjoel: Removed. Also from the rest of the _helpers.py in #363

examples/game_of_life/pyproject.toml
```diff
@@ -0,0 +1,24 @@
...
+click = ">=7.1.2, !=8.1.4" # mypy fails with click 8.1.4: https://github.com/pallets/click/issues/2558
+
+[tool.poetry.group.dev.dependencies]
+mypy = ">=1.0"
+grpc-stubs = "^1.53"
+types-protobuf = "^4.21"
+# Uncomment to use prerelease dependencies.
+ni-measurementlink-service = {path = "../..", develop = true}
+
+[tool.mypy]
```
dixonjoel, line 19: @bkeryan I put this to try to enforce type hints in this example. However, it allowed this where **kwargs doesn't have a type hint and it doesn't check functions prepended with '_' (private methods). Is this expected or is mypy not checking things correctly?
```python
def main(verbosity: int, **kwargs) -> None:
```

dixonjoel: Looks like the PR checks caught these things. I must have been invoking it incorrectly locally.

bkeryan (approved):

PR #361, opened 2023-08-21
service: Add support for ETW logging

bkeryan:

Add support for ETW logging using Microsoft's `traceloggingdynamic` package.

Notes:
- PerfView is able to correctly calculate elapsed time for these log messages. 
- I wasn't able to get NI's internal ETW viewer to identify these as paired events. I tried specifying event ids, but sometimes the viewer mismatched client/server calls together.
- This code currently doesn't push/pop the current thread's activity id, so the related activity id is always 0. 

As a developer, I want better visibility into MeasurementLink measurement execution time

Manually tested with PerfView and NI's internal ETW viewer.

### DIFF
--- a/ni_measurementlink_service/_loggers.py
+++ b/ni_measurementlink_service/_loggers.py
@@ -20,6 +20,13 @@ from typing import (
 import grpc
 from typing_extensions import Self
 
+try:
+    import traceloggingdynamic
+
+    _event_provider = traceloggingdynamic.Provider(b"NI-MeasurementLink-Python")
+except ImportError:
+    _event_provider = None
+
 _logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


bkeryan:
Example of PerfView output:
image

jasonmreding:
> I wasn't able to get NI's internal ETW viewer to identify these as paired events. I tried specifying event ids, but sometimes the viewer mismatched client/server calls together.

Can multiple start events be emitted before the corresponding stop is emitted? If so, there should be a way to annotate the event as recursive, and you can see if that makes a difference. You can look at the TransactionStart event on the PF event source for an example of this. I'm not sure if a similar concept exists or not in the Python library you're using.

The pairing logic in the viewer has always been a little finicky. I thought there were three different options for trying to pair up event start/stop data. Do none of them work? I guess if the lack of recursive option is causing the events to get dropped, it would probably throw off all of the pairing algorithms. 

You could also generate a unique ID that is included with the output of the start/stop event and then use that unique ID to pair start/stop events. That requires hard coding knowledge of that pairing for that event source in ETW Viewer so it's kind of a pain. We could probably generalize the pattern if we find it useful. That might also incur some boxing overhead when logging unless you also provide a matching override of WriteEvent in the event source.

jasonmreding (approved):

dixonjoel (approved):

bkeryan:
```
Run poetry run pytest ./tests -v --cov=ni_measurementlink_service --junitxml=test_results/nims-windows-latest-py3.8.xml
ImportError while loading conftest 'D:\a\measurementlink-python\measurementlink-python\tests\conftest.py'.
tests\conftest.py:9: in <module>
    from ni_measurementlink_service._internal.discovery_client import _get_registration_json_file_path
ni_measurementlink_service\__init__.py:12: in <module>
    from ni_measurementlink_service.measurement.service import (
ni_measurementlink_service\measurement\service.py:30: in <module>
    from ni_measurementlink_service._internal.discovery_client import DiscoveryClient
ni_measurementlink_service\_internal\discovery_client.py:19: in <module>
    from ni_measurementlink_service._loggers import ClientLogger
ni_measurementlink_service\_loggers.py:29: in <module>
    from ni_measurementlink_service import _tracelogging
ni_measurementlink_service\_tracelogging.py:11: in <module>
    _event_provider = traceloggingdynamic.Provider(b"NI-MeasurementLink-Python")
.venv\lib\site-packages\traceloggingdynamic.py:1265: in __init__
    self.id = id if id != None else providerid_from_name_bytes_utf[8](https://github.com/ni/measurementlink-python/actions/runs/5966151523/job/16185221550#step:7:9)(name_utf8)
.venv\lib\site-packages\traceloggingdynamic.py:113: in providerid_from_name_bytes_utf8
    return providerid_from_name(str(providername_utf8, encoding='utf-8'))
.venv\lib\site-packages\traceloggingdynamic.py:[9](https://github.com/ni/measurementlink-python/actions/runs/5966151523/job/16185221550#step:7:10)3: in providerid_from_name
    sha1 = hashlib.sha1(usedforsecurity = False)
E   TypeError: 'usedforsecurity' is an invalid keyword argument for openssl_sha1()
```

I filed an issue about this: https://github.com/microsoft/tracelogging/issues/58

PR #367, opened 2023-09-05
Add type hints and clean up documentation for encoders and decoders

dixonjoel:

serialization_strategy.py was confusing because it had Callables returning Callables without any type hints showing how they were related.

This adds some type hints for the 'InnerEncoder' and 'InnerDecoder', which we get from google.protobuf.internal. It also adds hints for 'EncoderFactory' and 'DecoderFactory', which are our wrappers that hold on to the InnerEncoder and InnerDecoder.

Also removes the unnecessary 'Context' class from serialization_strategy.

Clarify the code we use for serialization and deserialization.

Ran 'poetry run pytest -v' locally.


ni_measurementlink_service/_internal/parameter/EncodeDecodeNotes.md
dixonjoel, line 1: I'll remove this file. Ignore please.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,25 +1,24 @@
...
+from typing import Any, Callable, Dict
 
 from google.protobuf import type_pb2
-from google.protobuf.internal import decoder
-from google.protobuf.internal import encoder
+from google.protobuf.internal import decoder, encoder
+from typing_extensions import TypeAlias
 
+InnerEncoder: TypeAlias = Callable[[Callable[[bytes], int], bytes, bool], int]
+InnerDecoder: TypeAlias = Callable[
```
dixonjoel, line 10: [Typeshed](https://github.com/python/typeshed/blob/main/stubs/protobuf/google/protobuf/internal/decoder.pyi) has _Decoder as `_Decoder: TypeAlias = Callable[[str, int, int, Message, dict[FieldDescriptor, Any]], int]`. However, this flags our [one call of it](https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/_internal/parameter/serializer.py#L165) as having invalid arguments. We're passing 'memoryview' instead of 'str', 'ValueType' instead of 'Message', and 'dict[int, Any]' instead of 'dict[FieldDescriptor, Any]'.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,25 +1,24 @@
...
-        Callable[[int],Callable]: Callable Encoder for scalar types that takes
-        in field_index and returns the Low-level Encode Callable.
+def _scalar_encoder(encoder) -> Callable[[int], InnerEncoder]:
+    """Constructs a scalar encoder factory that takes a field index and returns an InnerEncoder.
 
+    This class returns the InnerEncoder callable with is_repeated set to False
+    and is_packed set to False.
     """
 
     def scalar_encoder(field_index):
```
bkeryan, line 24: Untyped return values are also treated as `Any`. When you add type hints to the outer function, the inner function will still allow type mismatches.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -141,117 +115,76 @@ def vector_decoder(field_index, key):
...
+    type_pb2.Field.TYPE_UINT64: int(),
+    type_pb2.Field.TYPE_BOOL: bool(),
+    type_pb2.Field.TYPE_STRING: str(),
+    type_pb2.Field.TYPE_ENUM: int(),
+}
+
+
+def get_encoder(
+    type: type_pb2.Field.Kind.ValueType, repeated: bool
+) -> Callable[[int], InnerEncoder]:
```
bkeryan, line 157:
```suggestion
) -> EncoderFactory:
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -141,117 +115,76 @@ def vector_decoder(field_index, key):
...
+        raise Exception(f"Error can not encode type '{type}'")
+    scalar, array = _FIELD_TYPE_TO_ENCODER_MAPPING[type]
+    if repeated:
+        return array
+    return scalar
+
+
+def get_decoder(
+    type: type_pb2.Field.Kind.ValueType, repeated: bool
+) -> Callable[[int, str], InnerDecoder]:
```
bkeryan, line 172:
```suggestion
) -> DecoderFactory:
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -141,117 +115,76 @@ def vector_decoder(field_index, key):
...
+    """
+    if type not in _FIELD_TYPE_TO_DECODER_MAPPING:
+        raise Exception(f"Error can not decode type '{type}'")
+    scalar, array = _FIELD_TYPE_TO_DECODER_MAPPING[type]
+    if repeated:
+        return array
+    return scalar
+
+
+def get_type_default(type: type_pb2.Field.Kind.ValueType, repeated: bool) -> Any:
```
bkeryan, line 185: [maybe outside the scope of this PR] This function's return value is another source of `Any`.

One alternative is to return `object`. This should still allow comparisons, but it will prevent dereferencing fields/properties/methods or passing this value anywhere without a cast.

I don't think defining overloads for literal Kind.ValueType enum values will work because we call this with a variable Kind.ValueType enum value.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
bkeryan: `memoryview` instead of `str`: I brought this up in the previous PR:

> I am not sure what type this should be (getbuffer() is a memory view, getvalue() is bytes) or whether the typeshed for google.protobuf.internal.decoder._Decoder is correct (it uses str?)

`ValueType` instead of `Message`: `Kind` is a protobuf enum. `Kind.ValueType` is an opaque alias for `int`: https://github.com/python/typeshed/blob/main/stubs/protobuf/google/protobuf/type_pb2.pyi#L121 . I don't understand how we are able to pass an int instead of a message. Either our code is incorrect or typeshed is incorrect.

`int` instead of `FieldDescriptor`: I brought this up in the previous PR:

> The 2nd parameter is key, which is typed as str in serialization_strategy.py.
>
>Here, we are passing an int.
>
>typeshed annotates key as a FieldDescriptor.
>
>Are these the same thing? Which is right?

By "are these the same thing" I meant "am I even looking at the correct type?"

bkeryan (changes requested):

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,25 +1,24 @@
 """Serialization Strategy."""
 
-from typing import Any, Callable
+from typing import Any, Callable, Dict
 
 from google.protobuf import type_pb2
-from google.protobuf.internal import decoder
-from google.protobuf.internal import encoder
+from google.protobuf.internal import decoder, encoder
+from typing_extensions import TypeAlias
```
bkeryan, line 7: Let's not add runtime dependencies on typing_extensions. Also, we don't need typing_extensions with newer Python versions.

```suggestion
if typing.TYPE_CHECKING:
    if sys.version_info >= (3, 10):
        from typing import TypeAlias
    else:
        from typing_extensions import TypeAlias
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,25 +1,24 @@
...
-    Args
-    ----
-       encoder (Callable[[int, bool, bool], Callable]): Specific encoder that takes in
-       field_index, is_repeated, is_packed and returns the Low-level Encode Callable.
-
-    Returns
-    -------
-        Callable[[int],Callable]: Callable Encoder for scalar types that takes
-        in field_index and returns the Low-level Encode Callable.
+def _scalar_encoder(encoder) -> Callable[[int], InnerEncoder]:
```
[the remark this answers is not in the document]
bkeryan: Another way to double-check this stuff is to inspect the runtime types. We should do this when filing a bug against typeshed: use `inspect.signature` to print the runtime type and use `typing.reveal_type` to print the annotated type. Note that `typing.reveal_type` only works when running mypy or pyright, not at runtime.

```
PS D:\dev\measurementlink-python> poetry run python
Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from google.protobuf.internal import encoder
>>> encoder.FloatEncoder
<function _FloatingPointEncoder.<locals>.SpecificEncoder at 0x0000023D876F3EE0>
>>> import inspect
>>> inspect.signature(encoder.FloatEncoder)
<Signature (field_number, is_repeated, is_packed)>
>>> inner_encoder = encoder.FloatEncoder(1, False, False)
>>> inner_encoder
<function _FloatingPointEncoder.<locals>.SpecificEncoder.<locals>.EncodeField at 0x0000023D8721F0D0>
>>> inspect.signature(inner_encoder)
<Signature (write, value, unused_deterministic=None)>
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
bkeryan: I looked into whether we can construct `FieldDescriptor`. We can't construct descriptors directly anymore, but if we use `descriptor.proto` types to add a `FileDescriptorProto` to the descriptor pool, then the protobuf package will generate classes at run time like it does when you import a `_pb2.py` file.

https://github.com/ni/measurementlink-python/issues/35

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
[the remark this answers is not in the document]
dixonjoel: This from typeshed seems to match what I'm getting from reveal_type.
```bat
FloatEncoder: _Encoder
def BoolEncoder(field_number: int, is_repeated: bool, is_packed: bool) -> _Encoder

reveal_type(encoder.FloatEncoder)
Revealed type is "def (def (builtins.bytes) -> builtins.int, builtins.bytes, builtins.bool) -> builtins.int"
reveal_type(encoder.BoolEncoder)
Revealed type is "def (field_number: builtins.int, is_repeated: builtins.bool, is_packed: builtins.bool) -> def (def (builtins.bytes) -> builtins.int, builtins.bytes, builtins.bool) -> builtins.int"
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
dixonjoel: Changed this to what I'm calling `PartialEncoderConstructor`

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
dixonjoel: Changed to `PartialDecoderConstructor`

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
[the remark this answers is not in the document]
dixonjoel: I don't understand how we can call both of these with field_number, is_repeated, and is_packed if these are really the types.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
dixonjoel: I get these errors if I use this instead of importing TypeAlias. Should I just remove the 'if typing.TYPE_CHECKING`?
```bat
ni_measurementlink_service\_internal\parameter\serialization_strategy.py:18: in <module>
    WriteFunction: TypeAlias = Callable[[bytes], int]
E   NameError: name 'TypeAlias' is not defined
```

bkeryan: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING gives two solutions to this problem.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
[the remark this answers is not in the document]
dixonjoel: How is it that typing.reveal_type and mypy can even know what these are? If there's no type hints in the original code, how do we know that this is int, bool, bool? Does it go into the code and see, e.g. that 'is_repeated' is used like so `if is_repeated:` so therefore it must be a bool?
```python
  def SpecificEncoder(field_number, is_repeated, is_packed):
```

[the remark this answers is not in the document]
dixonjoel: And if mypy and typing.reveal_type give types that match typeshed, then can we really file a bug to them? I don't know how they generate their stub types - using runtime types or using mypy / 'structural subtyping'?

dixonjoel:
> memoryview instead of str: I brought this up in the previous PR:

Ultimately, I believe whatever is passed in just has to be 'sliceable' so I think either str or memoryview would work at runtime.
https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/decoder.py#L287C11-L287C11

InnerDecode in decoder.py says "Args: buffer: memoryview of the serialized bytes"
https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/decoder.py#L277C49-L277C49

However, the docs at the top of the file say "The arguments are:  buffer:     The string containing the encoded message."
https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/decoder.py#L14C1-L16C1

So I don't know which to believe. I'm inclined to leave it as `memoryview` in our TypeAlias

dixonjoel:
@bkeryan 

> ValueType instead of Message: Kind is a protobuf enum. Kind.ValueType is an opaque alias for int: https://github.com/python/typeshed/blob/main/stubs/protobuf/google/protobuf/type_pb2.pyi#L121 . I don't understand how we are able to pass an int instead of a message. Either our code is incorrect or typeshed is incorrect.

The code simply uses the `key` that we pass it to index the dictionary, both to check if the value is None and call new_default in order to initialize the value in the dictionary and also to store the parsed value into the dictionary, so reading the code it looks to me like we could use any key as long as we're consistent.
https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/decoder.py#L175C56-L175C66

Since it isn't easy for us to be constructing FieldDescriptors (I think) and since we're already using int, I would like to leave the key in the TypeAlias as int or even change it to Any? What do you think?

dixonjoel:
@bkeryan 

> ValueType instead of Message: Kind is a protobuf enum. Kind.ValueType is an opaque alias for int: https://github.com/python/typeshed/blob/main/stubs/protobuf/google/protobuf/type_pb2.pyi#L121 . I don't understand how we are able to pass an int instead of a message. Either our code is incorrect or typeshed is incorrect.

This parameter is only used to pass back to the `new_default` method as a parameter, which is unused. So we were passing the 'type' incorrectly, I think. I would like to change this to pass `cast(Message, None)`?

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,257 +1,201 @@
...
+NewDefault: TypeAlias = Callable[[Message], Message]
+DecoderConstructor: TypeAlias = Callable[[int, bool, bool, FieldDescriptor, NewDefault], Decoder]
 
-    Returns
-    -------
-        Callable[[int],Callable]: Callable Encoder for scalar types that takes
-        in field_index and returns the Low-level Encode Callable.
 
+def _scalar_encoder(encoder: EncoderConstructor) -> PartialEncoderConstructor:
+    """Constructs a scalar encoder factory that takes a field index and returns an Encoder.
```
dixonjoel, line 31: TODO: Update terminology for 'constructor' vs 'factory'.

bkeryan:
> @bkeryan
> 
> > memoryview instead of str: I brought this up in the previous PR:
> 
> Ultimately, I believe whatever is passed in just has to be 'sliceable' so I think either str or memoryview would work at runtime. https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/decoder.py#L287C11-L287C11
> 
> InnerDecode in decoder.py says "Args: buffer: memoryview of the serialized bytes" https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/decoder.py#L277C49-L277C49
> 
> However, the docs at the top of the file say "The arguments are: buffer: The string containing the encoded message." https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/decoder.py#L14C1-L16C1
> 
> So I don't know which to believe. I'm inclined to leave it as `memoryview` in our TypeAlias

Sliceable is not enough. The element type also matters.

`str` might have been correct in Python 2.7, but `str` is Unicode and cannot store bytes in Python 3.x.

```python
>>> "abc"[0] & 0x80
TypeError: unsupported operand type(s) for &: 'str' and 'int'
```

Both `bytes` and `memoryview` implement `Sequence[int]`, but there are at least two significant differences:
- `bytes` is immutable and `memoryview` is mutable
- Slicing `bytes` makes a copy. Slicing `memoryview` returns a new `memoryview` referencing the same buffer.

It looks like protobuf creates a `memoryview` in `Message.MergeFromString`. Both `Message.FromString` and `Message.ParseFromString` use `Message.MergeFromString` (at least in the pure-Python implementation). Typeshed annotates all three methods as taking `bytes`.

Based on this, here's what I think:
- "The arguments are: buffer: The string containing the encoded message." is incorrect and should say memoryview.
- The typeshed hints for decoder.pyi are incorrect and should say `memoryview`, not `str`.
- We should file GitHub issues for both of these, as well as the decoder vs. decoder-constructor mixup.

bkeryan:
> @bkeryan
> 
> > int instead of FieldDescriptor: I brought this up in the previous PR:
> 
> The code simply uses the `key` that we pass it to index the dictionary, both to check if the value is None and call new_default in order to initialize the value in the dictionary and also to store the parsed value into the dictionary, so reading the code it looks to me like we could use any key as long as we're consistent. https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/decoder.py#L175C56-L175C66
> 
> Since it isn't easy for us to be constructing FieldDescriptors (I think) and since we're already using int, I would like to leave the key in the TypeAlias as int or even change it to Any? What do you think?

I think we should add a type alias called `Key` with a comment explaining why we are using a different type than in typeshed, use it in our annotations, and cast when necessary.

I also think that in the long run we should scrap this code and create `FileDescriptorProtos`, but that's outside the scope of the current feature.

bkeryan:
> > @bkeryan
> > > int instead of FieldDescriptor: I brought this up in the previous PR:
> > 
> > 
> > The code simply uses the `key` that we pass it to index the dictionary, both to check if the value is None and call new_default in order to initialize the value in the dictionary and also to store the parsed value into the dictionary, so reading the code it looks to me like we could use any key as long as we're consistent. https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/decoder.py#L175C56-L175C66
> > Since it isn't easy for us to be constructing FieldDescriptors (I think) and since we're already using int, I would like to leave the key in the TypeAlias as int or even change it to Any? What do you think?
> 
> I think we should add a type alias called `Key` with a comment explaining why we are using a different type than in typeshed, use it in our annotations, and cast when necessary.
> 
> I also think that in the long run we should scrap this code and create `FileDescriptorProtos`, but that's outside the scope of the current feature.

One more thought: perhaps `Key` should be a `TypeVar`, not a `TypeAlias`.

bkeryan:
> Based on this, here's what I think:
> 
> * "The arguments are: buffer: The string containing the encoded message." is incorrect and should say memoryview.
> * The typeshed hints for decoder.pyi are incorrect and should say `memoryview`, not `str`.
> * We should file GitHub issues for both of these, as well as the decoder vs. decoder-constructor mixup.

BTW, there are some new types for working with bytes-like objects:
- `collections.abc.Buffer` / `typing_extensions.Buffer`: https://github.com/python/typing_extensions/blob/main/src/typing_extensions.py#L2522
`_typeshed.ReadableBuffer`: https://github.com/python/typeshed/blob/32535ee8952dffecbe7a9fd04e5ed8ba0f1c8da2/stdlib/_typeshed/__init__.pyi#L239

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,118 +1,133 @@
...
 
+WriteFunction: TypeAlias = Callable[[bytes], int]
+Encoder: TypeAlias = Callable[[WriteFunction, bytes, bool], int]
+PartialEncoderConstructor: TypeAlias = Callable[[int], Encoder]
+EncoderConstructor: TypeAlias = Callable[[int, bool, bool], Encoder]
 
-def _scalar_encoder(encoder) -> Callable[[int], InnerEncoder]:
-    """Constructs a scalar encoder factory that takes a field index and returns an InnerEncoder.
+Decoder: TypeAlias = Callable[[memoryview, int, int, Message, Dict[int, Any]], int]
+PartialDecoderConstructor: TypeAlias = Callable[[int, str], Decoder]
```
bkeryan, line 25: Create a `TypeAlias` or `TypeVar` for `Key` and use it here:
- `Dict[Key, Any]`
- `Callable[[int, Key], Decoder]`

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,118 +1,133 @@
...
+        from typing_extensions import TypeAlias
 
+WriteFunction: TypeAlias = Callable[[bytes], int]
+Encoder: TypeAlias = Callable[[WriteFunction, bytes, bool], int]
+PartialEncoderConstructor: TypeAlias = Callable[[int], Encoder]
+EncoderConstructor: TypeAlias = Callable[[int, bool, bool], Encoder]
 
-def _scalar_encoder(encoder) -> Callable[[int], InnerEncoder]:
-    """Constructs a scalar encoder factory that takes a field index and returns an InnerEncoder.
+Decoder: TypeAlias = Callable[[memoryview, int, int, Message, Dict[int, Any]], int]
```
bkeryan, line 24: FYI: it is possible to override a 3rd party type stub by forking it and putting it in the first directory listed in `mypy_path`. We could do this to override decoder.pyi when we run mypy on our project. However, I don't know if running mypy on a measurement service will also need the forked type stub. We shouldn't distribute it as part of our package.

bkeryan (changes requested):
Thanks, this is very helpful for understanding the differences between the various encoder/decoder functions.

ni_measurementlink_service/_internal/parameter/serializer.py
```diff
@@ -158,15 +159,15 @@ def _get_overlapping_parameters(
...
+        decoder = serialization_strategy.get_decoder(type, is_repeated)
         inner_decoder = decoder(field_index, field_index)
         parameter_bytes_io = BytesIO(parameter_bytes)
         parameter_bytes_memory_view = parameter_bytes_io.getbuffer()
         position = inner_decoder(
             parameter_bytes_memory_view,
             position + encoder._TagSize(field_index),  # type: ignore[attr-defined]
             len(parameter_bytes),
-            type,
+            cast(Message, None),  # unused
```
bkeryan, line 170: This is another situation where we are assuming that new_default is never used, and if it is used, customers will get a cryptic error about `NoneType`.

Would it be better to pass a dummy message? (The answer may be "no")

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,118 +1,133 @@
...
-    Takes a field index and a key and returns an InnerDecoder.
+    Takes a field index and a key and returns an Decoder.
 
-    This class returns the InnerDecoder callable with is_repeated set to False
+    This class returns the Decoder callable with is_repeated set to False
     and is_packed set to False.
     """
 
-    def scalar_decoder(field_index, key):
+    def scalar_decoder(field_index: int, key) -> Decoder:
```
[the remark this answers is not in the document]
bkeryan: If we annotate it as a TypeVar, we probably don't have to cast.

dixonjoel:
@bkeryan FYI looks like there's some related work around bytes going on in typeshed. I plan to file an issue also specific to encoder.
https://github.com/python/typeshed/issues/10656
https://github.com/python/typeshed/issues/9006
https://github.com/python/typeshed/issues/9001

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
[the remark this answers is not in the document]
dixonjoel: https://github.com/python/typeshed/issues/10695
Issue for encoder types in typeshed

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
dixonjoel: Deferring this for future refactoring.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
dixonjoel: Added type hints for this inner method.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,25 +1,24 @@
...
-from google.protobuf.internal import decoder
-from google.protobuf.internal import encoder
+from google.protobuf.internal import decoder, encoder
+from typing_extensions import TypeAlias
 
+InnerEncoder: TypeAlias = Callable[[Callable[[bytes], int], bytes, bool], int]
+InnerDecoder: TypeAlias = Callable[
+    [memoryview, int, int, type_pb2.Field.Kind.ValueType, Dict[int, Any]], int
+]
+EncoderFactory: TypeAlias = Callable[[int], InnerEncoder]
```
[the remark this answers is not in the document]
dixonjoel: Now using partial, constructor, and encoder / decoder for these terms consistently. Lmk if I missed anything or if you want other changes on the terminology.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,118 +1,133 @@
...
-StringArrayDecoder = _vector_decoder(decoder.StringDecoder, is_packed=False)
+FloatArrayDecoder = _vector_decoder(cast(DecoderConstructor, decoder.FloatDecoder))
+DoubleArrayDecoder = _vector_decoder(cast(DecoderConstructor, decoder.DoubleDecoder))
+Int32ArrayDecoder = _vector_decoder(cast(DecoderConstructor, decoder.Int32Decoder))
+UInt32ArrayDecoder = _vector_decoder(cast(DecoderConstructor, decoder.UInt32Decoder))
+Int64ArrayDecoder = _vector_decoder(cast(DecoderConstructor, decoder.Int64Decoder))
+UInt64ArrayDecoder = _vector_decoder(cast(DecoderConstructor, decoder.UInt64Decoder))
+BoolArrayDecoder = _vector_decoder(cast(DecoderConstructor, decoder.BoolDecoder))
+StringArrayDecoder = _vector_decoder(
+    cast(DecoderConstructor, decoder.StringDecoder), is_packed=False
```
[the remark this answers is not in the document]
dixonjoel: BoolDecoder *is* typed as _Decoder
https://github.com/python/typeshed/blob/main/stubs/protobuf/google/protobuf/internal/decoder.pyi#L25

But yeah the StringDecoder cast is because of the str vs memoryview issue.

dixonjoel:
> We should file GitHub issues for both of these, as well as the decoder vs. decoder-constructor mixup.

https://github.com/protocolbuffers/protobuf/issues/14037
https://github.com/python/typeshed/issues/10697

ni_measurementlink_service/_internal/parameter/serializer.py
dixonjoel: Clarified the comment as opposed to a dummy message. The parameter can be passed back to us in _new_default, but we're the ones that aren't using the parameter.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,257 +1,212 @@
...
-    -------
-        Callable[[int,str],Callable]: Callable Decoder for 1D Array types that takes in
-        field_index , key and returns the Low-level Decode Callable.
+    Takes a field index and a key and returns a Decoder.
 
+    This class returns the Decoder with is_repeated set to True
+    and is_packed defaulting to True.
     """
 
     def _new_default(unused_message=None):
```
bkeryan, line 95: Oops, I overlooked this function. It has no annotations, so it's implicitly `Callable[[Any], Any]`.

It's supposed to be `Callable[[Message], Message]`:
```suggestion
    def _new_default(unused_message: Message = None) -> Message:
```

But this has problems:
```
ni_measurementlink_service\_internal\parameter\serialization_strategy.py:95: error: Incompatible default for argument "unused_message" (default has type "None", argument has type "Message")  [assignment]
ni_measurementlink_service\_internal\parameter\serialization_strategy.py:95: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
ni_measurementlink_service\_internal\parameter\serialization_strategy.py:95: note: Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
ni_measurementlink_service\_internal\parameter\serialization_strategy.py:96: error: Incompatible return value type (got "list[<nothing>]", expected "Message")  [return-value]
```

We can fix the 1st error by adding `Optional`: `unused_message: Optional[Message] = None`. It's okay to pass a `Callable[[Optional[Message]], Message]` to a function that expects `Callable[[Message], Message]`.

The 2nd error suggests that our code is wrong, typeshed is wrong, or both.

```
  field_dict: message._fields (avoids a hashtable lookup).
The decoder reads the field and stores it into field_dict, returning the new
buffer position.  A decoder for a repeated field may proactively decode all of
the elements of that field, if they appear consecutively.

  new_default:   A function which takes a message object as a parameter and
                 returns a new instance of the default value for this field.
                 (This is called for repeated fields and sub-messages, when an
                 instance does not already exist.)
```

 This is how `new_default` is used:
```python
        value = field_dict.get(key)
        if value is None:
          value = field_dict.setdefault(key, new_default(message))
```

This is where `new_default` comes from, in one case:
```
    elif decode_type == _FieldDescriptor.TYPE_STRING:
      field_decoder = decoder.StringDecoder(
          field_descriptor.number, is_repeated, is_packed,
          field_descriptor, field_descriptor._default_constructor,
          not field_descriptor.has_presence)
```

`field_descriptor._default_constructor` is constructed by calling `_DefaultValueConstructorForField`:
```
  field_descriptor._default_constructor = _DefaultValueConstructorForField(
      field_descriptor
  )
```

[_DefaultValueConstructorForField](https://github.com/protocolbuffers/protobuf/blob/45ed02805f9ea103d30d8f310e837c2a1468bf99/python/google/protobuf/internal/python_message.py#L408) returns a function that returns a `RepeatedScalarFieldContainer` or similar object.

[RepeatedScalarFieldContainer](https://github.com/protocolbuffers/protobuf/blob/45ed02805f9ea103d30d8f310e837c2a1468bf99/python/google/protobuf/internal/containers.py#L103) is a sequence but it's not a `Message`.

So, returning an empty list is not exactly the same thing that protobuf does, but it's close. It seems like the return value of `new_default` probably ought to be `Any`, not `Message`.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,257 +1,212 @@
...
         is_packed = False
-        return decoder(field_index, is_repeated, is_packed, key, None)
+        return decoder(field_index, is_repeated, is_packed, key, _unsupported_new_default)
 
     return scalar_decoder
 
 
-def _vector_decoder(decoder, is_packed=True) -> Callable[[int, str], Callable]:
-    """Abstract Specific Decoder(Callable) as Vector Decoder Callable that takes in field index,key.
+def _vector_decoder(decoder: DecoderConstructor, is_packed=True) -> PartialDecoderConstructor:
```
bkeryan, line 86: Using a TypeVar in a TypeAlias such as DecoderConstructor has consequences that I didn't consider: it makes the alias generic. If you use DecoderConstructor without specifying a type variable, it's equivalent to DecoderConstructor[Any].

Here's what reveal_type(_vector_decoder) says: 
```
ni_measurementlink_service\_internal\parameter\serialization_strategy.py:105: note: Revealed type is "def (decoder: def (builtins.int, builtins.bool, builtins.bool, Any, def (google.protobuf.message.Message) -> google.protobuf.message.Message) -> def (builtins.memoryview, builtins.int, builtins.int, google.protobuf.message.Message, builtins.dict[Any, Any]) -> builtins.int, is_packed: Any =) -> def (builtins.int, Any) -> def (builtins.memoryview, builtins.int, builtins.int, google.protobuf.message.Message, builtins.dict[Any, Any]) -> builtins.int"
```

One solution is to add `[Key]` to all references to these type aliases, including the references to Decoder inside the other type aliases:
```
PartialDecoderConstructor: TypeAlias = Callable[[int, Key], Decoder[Key]]
DecoderConstructor: TypeAlias = Callable[[int, bool, bool, Key, NewDefault], Decoder[Key]]
```

However, this is really starting to diverge from the (corrected) types in typeshed, so maybe we're better off specifying `Key = FieldDescriptor` and casting whenever we pass in an `int`.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,257 +1,212 @@
...
 
 
-def _vector_encoder(encoder, is_packed=True) -> Callable[[int], Callable]:
-    """Abstract Specific Encoder(Callable) as Vector Encoder Callable that takes in field index.
-
-    Args
-    ----
-       encoder (Callable[[int, bool, bool], Callable]): Specific encoder(Callable) that takes in
-       field_index, is_repeated, is_packed and returns the Low-level Encode Callable.
+def _vector_encoder(encoder: EncoderConstructor, is_packed=True) -> PartialEncoderConstructor:
```
bkeryan, line 48: mypy doesn't seem to infer types based on default arguments. We probably get this wrong a lot, but at least annotate is_repeated and is_packed everywhere.

```suggestion
def _vector_encoder(encoder: EncoderConstructor, is_packed: bool = True) -> PartialEncoderConstructor:
```

bkeryan (changes requested):

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
dixonjoel: Changed to Optional and Any return value.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
bkeryan: > How do you cast an int to a FieldDescriptor? They're very different things.

```python
        inner_decoder = decoder(field_index, cast(FieldDescriptor, field_index))
        ...
        position = inner_decoder(
            parameter_bytes_memory_view,
            position + encoder._TagSize(field_index),  # type: ignore[attr-defined]
            len(parameter_bytes),
            cast(Message, None),  # unused - See serialization_strategy._vector_decoder._new_default
            cast(Dict[FieldDescriptor, Any], overlapping_parameters_by_id),
        )
```

This has no effect at run time, but it lies to the type checker, saying "trust me, this is a FieldDescriptor". 

> Also you're talking about going back to a TypeAlias like this?
>
> Key: TypeAlias = FieldDescriptor

Yes.

> And does that mean this will return Dict[FieldDescriptor, Any]? As well as many other methods.
>
> ```
>def deserialize_parameters(
>    parameter_metadata_dict: Dict[int, ParameterMetadata], parameter_bytes: bytes
>) -> Dict[int, Any]:
> ```

No. Try to use int in the ni-measurementlink-python code. Cast to FieldDescriptor when directly interfacing with protobuf. I would be surprised if any code outside of serialization_strategy.py and serializer.py needs to use the FieldDescriptor version of these types.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,257 +1,216 @@
...
+}
+
+
+def get_encoder(type: type_pb2.Field.Kind.ValueType, repeated: bool) -> PartialEncoderConstructor:
+    """Get the appropriate partial encoder constructor for the specified type.
+
+    A scalar or vector constructor is returned based on the 'repeated' parameter.
+    """
+    if type not in _FIELD_TYPE_TO_ENCODER_MAPPING:
+        raise Exception(f"Error can not encode type '{type}'")
```
bkeryan, line 191:
```suggestion
        raise ValueError(f"Error can not encode type '{type}'")
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,257 +1,216 @@
...
-
-        """
-        if repeated:
-            return list()
-        type_default_value = Context._TYPE_DEFAULT_MAPPING.get(type)
-        return type_default_value
+    A scalar or vector constructor is returned based on the 'repeated' parameter.
+    """
+    if type not in _FIELD_TYPE_TO_DECODER_MAPPING:
+        raise Exception(f"Error can not decode type '{type}'")
```
bkeryan, line 204:
```suggestion
        raise ValueError(f"Error can not decode type '{type}'")
```

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
```diff
@@ -1,257 +1,216 @@
...
-        Callable[[int,str],Callable]: Callable Decoder for scalar types that takes in
-        field_index, key and returns the Low-level Decode Callable.
+    Takes a field index and a key and returns a Decoder.
 
+    This class returns the Decoder with is_repeated set to False
+    and is_packed set to False.
     """
 
-    def scalar_decoder(field_index, key):
+    def _unsupported_new_default(message: Message) -> Message:
```
bkeryan, line 75: Match the other one.
```suggestion
    def _unsupported_new_default(message: Optional[Message] = None) -> Any:
```

bkeryan (approved):
Looks good!

Approved with a few trivial suggestions.

ni_measurementlink_service/_internal/parameter/serialization_strategy.py
bkeryan: Return value is Any

bkeryan:
> @bkeryan FYI looks like there's some related work around bytes going on in typeshed. I plan to file an issue also specific to encoder. [python/typeshed#10656](https://github.com/python/typeshed/issues/10656) [python/typeshed#9006](https://github.com/python/typeshed/issues/9006) [python/typeshed#9001](https://github.com/python/typeshed/issues/9001)

One more:
https://github.com/python/typeshed/issues/10712

PR #368, opened 2023-09-06
Only service interfaces listed in .serviceconfig file

### LINKED ISSUE #359: `MeasurementService` always supports both v1 and v2 interfaces, even if `supportedInterfaces` specifies a subset
# Bug Report

`MeasurementService` creates servicers for both the measurement.v1 and measurement.v2 interfaces: https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/_internal/service_manager.py#L83 

However, the `.serviceconfig` file's `supportedInterfaces` key may specify a subset of interfaces. The measurement.v1 interface doesn't support UI updates or enums, so service authors may not necessarily want their services to support the measurement.v1 interface.

This leads to inconsistency between which interfaces are registered with the discovery service and which interfaces are actually supported by the service.

## Repro or Code Sample

Edit the `.serviceconfig` for an example service to remove the v1 or v2 interface.

## Expected Behavior

The server only accepts the interfaces listed in the `.serviceconfig` file.

## Current Behavior

The server accepts both interfaces.

## Possible Solution

Update `service_manager.py` to look at `service_info.provided_interfaces` when creating and adding servicers.

## Context

Identified in a proposal to split protobuf codegen into separate packages. The current behavior would always depend on both measurement.v1 codegen and measurement.v2 codegen.

## Your Environment

N/A

[AB#2491921](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2491921)

dixonjoel:

Instead of always creating servers for V1 and V2 measurement services, use the .serviceconfig file 'providedInterfaces' to determine which ones to serve.

Fixes #359 

Ran automated tests. Adding an automated test for v2 only service.

### DIFF
--- a/ni_measurementlink_service/_internal/service_manager.py
+++ b/ni_measurementlink_service/_internal/service_manager.py
@@ -87,26 +87,31 @@ class GrpcService:
                 ("grpc.max_send_message_length", -1),
             ],
         )
-        if _V1_INTERFACE in service_info.provided_interfaces:
-            servicer_v1 = MeasurementServiceServicerV1(
-                measurement_info,
-                configuration_parameter_list,
-                output_parameter_list,
-                measure_function,
-            )
-            v1_measurement_service_pb2_grpc.add_MeasurementServiceServicer_to_server(
-                servicer_v1, self.server
-            )
-        if _V2_INTERFACE in service_info.provided_interfaces:
-            servicer_v2 = MeasurementServiceServicerV2(
-                measurement_info,
-                configuration_parameter_list,
-                output_parameter_list,
-                measure_function,
-            )
-            v2_measurement_service_pb2_grpc.add_MeasurementServiceServicer_to_server(
-                servicer_v2, self.server
-            )
+        for interface in service_info.provided_interfaces:
+            if interface == _V1_INTERFACE:
+                servicer_v1 = MeasurementServiceServicerV1(
+                    measurement_info,
+                    configuration_parameter_list,
+                    output_parameter_list,
+                    measure_function,
+                )
+                v1_measurement_service_pb2_grpc.add_MeasurementServiceServicer_to_server(
+                    servicer_v1, self.server
+                )
+            elif interface == _V2_INTERFACE:
+                servicer_v2 = MeasurementServiceServicerV2(
+                    measurement_info,
+                    configuration_parameter_list,
+                    output_parameter_list,
+                    measure_function,
+                )
+                v2_measurement_service_pb2_grpc.add_MeasurementServiceServicer_to_server(
+                    servicer_v2, self.server
+                )
+            else:
+                raise Exception(
+                    f"Unknown interface was provided in the .serviceconfig file: {interface}"
+                )
         port = str(self.server.add_insecure_port("[::]:0"))
         self.server.start()
         _logger.info("Measurement service hosted on port: %s", port)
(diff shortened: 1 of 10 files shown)


ni_measurementlink_service/_internal/service_manager.py
```diff
@@ -85,24 +87,26 @@ def start(
...
-        servicer_v2 = MeasurementServiceServicerV2(
-            measurement_info,
-            configuration_parameter_list,
-            output_parameter_list,
-            measure_function,
-        )
-        v2_measurement_service_pb2_grpc.add_MeasurementServiceServicer_to_server(
-            servicer_v2, self.server
-        )
+        if _V1_INTERFACE in service_info.provided_interfaces:
```
bkeryan, line 90: I think this should error if any of the provided interfaces are unrecognized.

What about custom interfaces? 
- This class doesn't support adding custom gRPC servicers to its gRPC server at this time.
- You can create a separate gRPC server for your custom interfaces, but it will have a different listening port than the one we are registering here, so it's incorrect to include it in this method's call to `RegisterService`.

The easiest way to implement this is to iterate over provided_interfaces and do if v1 / elif v2 / else error. 

Also add a test case.

bkeryan (changes requested):

ni_measurementlink_service/_internal/service_manager.py
```diff
@@ -85,24 +87,31 @@ def start(
...
+                    measurement_info,
+                    configuration_parameter_list,
+                    output_parameter_list,
+                    measure_function,
+                )
+                v2_measurement_service_pb2_grpc.add_MeasurementServiceServicer_to_server(
+                    servicer_v2, self.server
+                )
+            else:
+                raise Exception(
```
bkeryan, line 112: Please don't raise the base `Exception` type. Use a more specific type. I think `ValueError` or `RuntimeError` would be ok here.

tests/integration/test_service_manager.py
```diff
@@ -76,6 +81,48 @@ def test___grpc_service_started___stop_service___service_stopped(grpc_service: G
...
+        v1_only_measurement.measurement_service.measure_function,
+    )
+
+    _validate_if_service_running_by_making_rpc(port_number)
+
+
+def test___grpc_service_unknown_interface___start_service_and_check_v1___raises_error(
+    grpc_service: GrpcService,
+):
+    with pytest.raises(Exception):
```
bkeryan, line 116: Use the specific exception type here.

tests/integration/test_service_manager.py
```diff
@@ -76,6 +81,48 @@ def test___grpc_service_started___stop_service___service_stopped(grpc_service: G
...
+):
+    port_number = grpc_service.start(
+        v2_only_measurement.measurement_service.measurement_info,
+        v2_only_measurement.measurement_service.service_info,
+        v2_only_measurement.measurement_service.configuration_parameter_list,
+        v2_only_measurement.measurement_service.output_parameter_list,
+        v2_only_measurement.measurement_service.measure_function,
+    )
+
+    with pytest.raises(Exception):
```
bkeryan, line 95: Is this RpcError?

bkeryan (changes requested):

tests/integration/test_service_manager.py
dixonjoel: Yes, it's RpcError.

bkeryan (approved):

PR #377, opened 2023-09-14
Use `poethepoet` to run tasks on multiple projects

### LINKED ISSUE #374: ni-python-styleguide treats ni_measurementlink_service as 3rd party for projects that copied Examples' helpers
# Bug Report

Running `ni-python-styleguide lint` on poetry projects that have copied the helpers files and teststand_fixture.py to use will treat ni-measurementlink-service incorrectly as 3rd party when organizing imports. Example of changes the ni styleguide wants to make on the daqmx helpers and the teststand_fixture.py.

image

## Repro or Code Sample

You can run `ni-python-styleguide lint` on any of the projects from[ this repo](https://dev.azure.com/ni/DevCentral/_git/srd-port-mlink-measurements?path=/measurements/src) to see the suggested changes on the helpers and teststand_fixture.py. See the errors from [this pipeline run.](https://dev.azure.com/ni/DevCentral/_build/results?buildId=6103995&view=logs&j=c956bc35-6b93-50d3-1f47-1db8bc2d8085&t=a0bb955e-8aaa-5009-0fbe-45867d0b1d75&l=303)

## Expected Behavior

Anyone using a copied helpers or teststand_fixture.py shouldn't run into any suggested lint changes if they also follow ni-python-styleguide.

## Current Behavior

ni-measurementlink-service is being prompted to be organized as a 3rd party import.

## Possible Solution

This was already addressed for ni_measurementlink_generator but not for the examples

## Context

The Test Workflow team copied the helpers files from the 2023 Q3 Examples and when running the pipeline that includes the ni-python-styleguide check, we fail that step because it thinks the driver helpers and teststand_fixture.py need changes. 

## Your Environment

*  Windows 10
* `ni-measurementlink-service` version 2023 Q3
* MeasurementLink version 2023 Q3
* Python version 3.9

[AB#2522357](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2522357)

bkeryan:

Add a dev dependency on `poethepoet` and use it to:
- Run `poetry install`, `ni-python-styleguide`, or `mypy` for `ni-measurementlink-service`, `ni-measurementlink-generator`, and all examples
- Run tests for `ni-measurementlink-service` and `ni-measurementlink-generator`

Add a dev dependency on `ni-python-styleguide` to all example projects.

Run `ni-python-styleguide fix` on each example project separately.
- Before, ni-python-styleguide put `ni_measurementlink_service` imports in the "first party" group, because we ran it using `ni_measurementlink_service`'s `pyproject.toml` file.
- Running ni-python-styleguide on each example project moves the `ni_measurementlink_service` imports from the "first party" group to the "third party" group.

Provides a new way to run tasks on all of the examples.

Fixes #374 

Manually tested Poe tasks.


pyproject.toml
```diff
@@ -92,3 +93,25 @@ module = [
   "traceloggingdynamic",
 ]
 ignore_missing_imports = true
+
+[tool.poe.tasks.install]
```
bkeryan, line 97: @mshafer-NI I would appreciate your feedback on these `poethepoet` tasks.

If this seems like a good practice, we could start using it in our GitHub workflows.

dixonjoel:
This seems fine to me. I've been doing these 4 commands each time I'm about to commit and I'd like to see if this will help or if we can have one poe task that does these tasks or a similar set that we consider a good workflow. We could add something to CONTRIBUTING.md to show typical workflows with these commands.
```bat
poetry run mypy .\ni_measurementlink_service
poetry run ni-python-styleguide fix
poetry run ni-python-styleguide lint
poetry run pytest -v
```

dixonjoel (approved):

pyproject.toml
```diff
@@ -92,3 +93,25 @@ module = [
...
+shell = "for dir in . ni_measurementlink_generator/ examples/*/; do pushd $dir; poetry install; popd; done"
+
+[tool.poe.tasks.ni-python-styleguide]
+args = [
+  { name = "command", positional = true, required = true }
```
mshafer-NI, line 102: does this allow flags? (like `fix --aggressive`)

pyproject.toml
```diff
@@ -92,3 +93,25 @@ module = [
...
+  { cmd = "mypy -p ni_measurementlink_service" },
+  { cmd = "mypy -p ni_measurementlink_generator", cwd = "ni_measurementlink_generator" },
+  { shell = "for dir in examples/*/; do pushd $dir; mypy .; popd; done" }
+]
+
+[tool.poe.tasks.test]
+sequence = [
+  { cmd = "pytest --cov=ni_measurementlink_service -v" },
+  { cmd = "pytest --cov=ni_measurementlink_generator -v", cwd = "ni_measurementlink_generator" }
+]
```
mshafer-NI, line 117: consider adding a
```suggestion
]

[tool.poe.tasks.all]
sequence = ["install", "ni-python-styleguide", "mypy", "test"]
```

pyproject.toml
mshafer-NI: This does seem like a good practice, it is on my backlog to experiment/adopt as standard tooling.

pyproject.toml
bkeryan: Not yet. Since this is a sequence task, I can add support for specific flags, but I don't think I can automatically pass extra arguments to the underlying tasks. It sounds like that is only supported for cmd and script tasks.

bkeryan: I wasn't able to get this to work.

`{ name = "extra_args", positional = true, multiple = true }` doesn't accept options.

`{ name = "aggressive", options = ["--aggressive"], type = "boolean" }` sets `$aggressive` to `True`, which is not useful for passing the argument unchanged.

bkeryan: I'm splitting the "ni-python-styleguide" task into "check-style" and "fix-style"

pyproject.toml
bkeryan: I added a "check" task that runs all the checks.

I don't really want to combine "install" because "install" is pretty slow.

pyproject.toml
```diff
@@ -92,3 +93,25 @@ module = [
...
+args = [
+  { name = "command", positional = true, required = true }
+]
+shell = "for dir in . ni_measurementlink_generator/ examples/*/; do pushd $dir; ni-python-styleguide $command; popd; done"
+
+[tool.poe.tasks.mypy]
+sequence = [
+  { cmd = "mypy -p ni_measurementlink_service" },
+  { cmd = "mypy -p ni_measurementlink_generator", cwd = "ni_measurementlink_generator" },
+  { shell = "for dir in examples/*/; do pushd $dir; mypy .; popd; done" }
```
bkeryan, line 110: This is actually running the copy of mypy that is installed in the root venv, not the copy in `examples/sample_measurement/.venv` or whatever. Someone (@mshafer-NI?) asked about this and I gave the wrong answer. 

For most of the examples, it still works out, but for `game_of_life` and `sample_measurement` it currently generates additional warnings that don't occur when you cd to the example directory and run `poetry run mypy .`.

This also affects ni-python-styleguide, but it doesn't matter because using the example's pyproject.toml is sufficient.

mshafer-NI: Yes, that was me.

We need to tell poetry to ignore the root .venv and make new .venv's (maybe clearing the VIRTUALENV_PATH would do this?), then here, get it to use the individual envs...

I would suggest posting a discussion on the poethepoet repo, the maintainer has been pretty responsive, and if someone else has figured this out they may have ideas.

bkeryan: > We need to tell poetry to ignore the root .venv and make new .venv's (maybe clearing the VIRTUALENV_PATH would do this?), then here, get it to use the individual envs...

Yes, unsetting `VIRTUAL_ENV` seems like a potential path forward. There are a couple of caveats with this approach, which I listed in the discussion post.

> I would suggest posting a discussion on the poethepoet repo

https://github.com/nat-n/poethepoet/discussions/166

bkeryan:
I haven't had a chance to get back to my PoeThePoet venv problems yet, so I posted a separate PR with the example-linting bugfix: https://github.com/ni/measurementlink-python/pull/443

bkeryan:
I won't be able to get back to this PR anytime soon.

PR #379, opened 2023-09-15
mypy: Disallow untyped defs

bkeryan:

Specify the mypy `--disallow-untyped-defs` option for `ni-measurementlink-service`, `ni-measurementlink-generator`, and all example projects.
- This causes mypy to display errors when type hints are missing.
- I initially was going to use `--disallow-incomplete-defs`, which would only error for functions with partial type hints, but the number of completely untyped functions was pretty low so I switched to `--disallow-untyped-defs`.

Fix all new mypy warnings and errors.

Specify more accurate type hints in some cases. For example, decorators should use type variables to represent the functions that they are decorating, rather than an unbracketed `Callable` (which is equivalent to `Callable[[Any, ...], Any]`).

Help the development team remember to add type hints, so we don't have to discuss it in code reviews.

Ran mypy on ni-measurementlink-service, ni-measurementlink-generator, and all example projects.


examples/nidaqmx_analog_input/_nidaqmx_helpers.py
```diff
@@ -11,7 +11,7 @@
 def create_task(
     session_info: nims.session_management.SessionInformation,
     session_grpc_channel: Optional[grpc.Channel] = None,
-    initialization_behavior=nidaqmx.SessionInitializationBehavior.AUTO,
+    initialization_behavior: nidaqmx.SessionInitializationBehavior = nidaqmx.SessionInitializationBehavior.AUTO,
```
bkeryan, line 14: I was surprised to learn that mypy doesn't infer the parameter type from its default value.

pyproject.toml
```diff
@@ -92,3 +92,9 @@ module = [
   "traceloggingdynamic",
 ]
 ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+# The tests are not yet ready for --disallow-untyped-defs (or even --disallow-incomplete-defs)
```
bkeryan, line 97: Oops, I forgot to run mypy on the tests. Let's use a less strict configuration for the tests, for now.

dixonjoel (approved):

PR #386, opened 2023-09-20
service: Add SessionManagementClient to MeasurementService and rework initialization code

bkeryan:

Add a `session_management_client` property to the `MeasurementService` object, which lazily constructs a `SessionManagementClient` and caches it.

Rework initialization code:
- `DiscoveryClient`: lazily create stub and allow injecting a `GrpcChannelPool`.
- `SessionManagementClient`: lazily create stub, allow injecting a `DiscoveryClient` and `GrpcChannelPool`, and automatically resolve service.
- `MeasurementService`:
  - Lazily create `DiscoveryClient`.
  - Defer `GrpcService` creation until `host_service()` is called.
  - Update `close_service()` to clear lazily created objects, in order to allow reusing the `MeasurementService` object.
  - Deprecate `grpc_service` property because it leaks implementation details.
  - Add `service_location` property as a replacement for `grpc_service` that uses public types.

The examples will be updated in [AB#2527580](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2527580).

Provide a simple, official API for creating the session management client, to replace the `create_session_management_client` function in `_helpers.py`.

Prepare to inject a `DiscoveryClient` and `GrpcChannelPool` into the `Reservation` classes to support improved session management APIs.

Ran `poetry run pytest -v`
Manually tested `nidcpower_source_dc_voltage` and `sample_measurement` with debug logging (`-vv`) enabled.


ni_measurementlink_service/measurement/service.py
```diff
@@ -183,10 +182,54 @@ def convert_value_to_str(value: object) -> str:
 
         self.configuration_parameter_list: List[Any] = []
         self.output_parameter_list: List[Any] = []
-        self.grpc_service = GrpcService()
-        self.context: MeasurementContext = MeasurementContext()
-        self.channel_pool: GrpcChannelPool = GrpcChannelPool()
-        self.discovery_client: DiscoveryClient = DiscoveryClient()
+        self.context = MeasurementContext()
```
dixonjoel, line 184: I notice you're omitting the type hints for a couple of these. Is this like C# where it's ok to use var if it's obvious from the rest of the line what the type is?

ni_measurementlink_service/measurement/service.py
```diff
@@ -183,10 +182,54 @@ def convert_value_to_str(value: object) -> str:
...
+
+        self._initialization_lock = threading.RLock()
```
dixonjoel, line 186: I read a little about RLock vs Lock, but can you explain why the different choice in service and discovery_client?

dixonjoel (approved):

ni_measurementlink_service/measurement/service.py
bkeryan: RLock is recursive. Lock isn't.

service has multiple lazily-initialized properties that call each other, so it needs a recursive lock.

discovery_client has a single lazily-initialized property, so it doesn't need a recursive lock.

dixonjoel: Ahh, thus the 'R'. :)

ni_measurementlink_service/measurement/service.py
bkeryan: Type checkers infer the attribute type based on the initialization code. https://mypy.readthedocs.io/en/stable/class_basics.html#instance-and-class-attributes

> ```
> class A:
>     def __init__(self, x: int) -> None:
>         self.x = x  # Aha, attribute 'x' of type 'int'
> ```

ni_measurementlink_service/measurement/service.py
bkeryan: Yeah, also not to be confused with RWLock (which is standard in some languages but not Python)

PR #389, opened 2023-09-20
service: Update reserve_session(s) to accept a single pin and explicitly specify default timeout

bkeryan:

`SessionManagementClient`:
- Update `reserve_session(s)` to accept a single pin for the `pin_or_relay_names` argument.
- Change `reserve_session(s)` to explicitly specify a default timeout of 0.0.
- Update documentation.
- Add more unit test cases. 

Accepting a single pin allows calling `reserve_session(context, "Vdd")` without wrapping the pin name in a list or tuple.

In Python, `str` is a `Iterable[str]` that iterates over the characters of the string. Before this change, `reserve_session(context, "Vdd")` would pass type checking, but at run time it translated `"Vdd"` to `["V", "d", "d"]`, which is surprising behavior.

The default timeout behavior was previously undocumented. Using a default value of `None` was ambiguous, because `None` could plausibly be interpreted as either -1 (infinite) or 0.0 (non-blocking).

Ran `poetry run pytest -v`.
Manually tested a modified version of `nidmm_measurement.
Ran `poetry run sphinx-build _docs_source docs -b html -W --keep-going` and looked at the output.

### DIFF
--- a/ni_measurementlink_service/session_management.py
+++ b/ni_measurementlink_service/session_management.py
@@ -121,9 +121,19 @@ class SessionInformation(NamedTuple):
     """
 
     session_exists: bool
-    """Indicates whether the session exists in the Session Manager. 
+    """Indicates whether the session has been registered with the session management service.
     
-    This indicates whether the session has been created."""
+    When calling measurements from TestStand, the test sequence's ``ProcessSetup`` callback
+    creates instrument sessions and registers them with the session management service so that
+    they can be shared between multiple measurement steps. In this case, the `session_exists`
+    attribute is ``True``, indicating that the instrument sessions were already created and any
+    one-time setup (such as creating NI-DAQmx channels or loading NI-Digital files) has been
+    performed.
+    
+    When calling measurements outside of TestStand, the `session_exists` attribute is ``False``,
+    indicating that the measurement is responsible for creating the instrument sessions and
+    performing any one-time setup.
+    """
 
     channel_mappings: Iterable[ChannelMapping]
     """List of site and pin/relay mappings that correspond to each channel in the channel_list.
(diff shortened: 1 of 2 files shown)


ni_measurementlink_service/session_management.py
```diff
@@ -305,40 +316,44 @@ def _get_stub(self) -> session_management_service_pb2_grpc.SessionManagementServ
...
                 For NI instruments, use instrument type id constants, such as
                 :py:const:`INSTRUMENT_TYPE_NI_DCPOWER` or :py:const:`INSTRUMENT_TYPE_NI_DMM`.
 
                 For custom instruments, use the instrument type id defined in the pin map file.
 
-            timeout (float): Timeout in seconds. Allowed values,
-                0 (non-blocking, fails immediately if resources cannot be reserved),
-                -1 or negative (infinite timeout), or
-                any positive numeric value (wait for that number of second).
+            timeout: Timeout in seconds.
```
dixonjoel, line 348: Milliseconds

ni_measurementlink_service/session_management.py
```diff
@@ -356,40 +371,44 @@ def reserve_session(
...
```
dixonjoel, line 403: Milliseconds

ni_measurementlink_service/session_management.py
```diff
@@ -499,26 +515,28 @@ def unregister_sessions(self, session_info: Iterable[SessionInformation]) -> Non
...
```
dixonjoel, line 531: Milliseconds

dixonjoel: The .proto files uses `timeout_in_milliseconds`. Maybe we should change to that for clarity. But that would be a breaking change, I guess.

dixonjoel: Oh, it looks from the tests like this is seconds. So maybe just fix the sentences below to be consistent.

dixonjoel (approved):

ni_measurementlink_service/session_management.py
dixonjoel: Correction: Change 'milliseconds' below to seconds.

ni_measurementlink_service/session_management.py
dixonjoel: Correction: Change 'milliseconds' below to seconds.

ni_measurementlink_service/session_management.py
bkeryan: Oops, I updated this from the .proto file and didn't notice that. Thanks.

ni_measurementlink_service/session_management.py
bkeryan: Other Python APIs like [threading.Lock](https://docs.python.org/3/library/threading.html#lock-objects) and [grpc calls](https://grpc.github.io/grpc/python/grpc.html#grpc.UnaryUnaryMultiCallable) use `timeout` in seconds, so the Python API is fine as it is. I just didn't notice the difference when fixing the formatting & removing "or negative" (which is not really allowed).

In the .proto file, we arguably should have used `google.protobuf.Duration`, but it is what it is.

bkeryan: Also, now I see that while the session management service doesn't accept a timeout of -2, the Python API converts -2 to -1. Sigh.

dixonjoel (approved):

Issue #369, opened 2023-09-06
Auto-launching discovery service fails after clean install


bkeryan:
# Bug Report

After installing MeasurementLink on a clean system, running a Python measurement service fails.

## Repro or Code Sample

- Reimage PC
- Install MeasurementLink and drivers
- Do not run InstrumentStudio or MeasurementLink UI Editor
- Install Python, Poetry, etc.
- Run a Python measurement service

## Expected Behavior

The Python measurement service starts the MeasurementLink discovery service.

## Current Behavior

The Python measurement service starts the MeasurementLink discovery service (visible in Task Manager), but it also raises an unhandled exception:
```
PS D:\dev\measurementlink-python\examples\sample_measurement> poetry run python .\measurement.py -v
2023-09-06 13:52:04,696 INFO: Measurement service hosted on port: 52461
2023-09-06 13:52:04,723 ERROR: Error in registering with discovery service.
Traceback (most recent call last):
  File "D:\dev\measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 128, in register_measurement_service
    register_response = self.stub.RegisterService(request)
  File "D:\dev\measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 85, in stub
    address = _get_discovery_service_address()
  File "D:\dev\measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 223, in _get_discovery_service_address
    _ensure_discovery_service_started(key_file_path)
  File "D:\dev\measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 237, in _ensure_discovery_service_started
    _discovery_service_subprocess = _start_service(exe_file_path, key_file_path)
  File "D:\dev\measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 284, in _start_service
    with _open_key_file(str(key_file_path)) as _:
  File "D:\dev\measurementlink-python\ni_measurementlink_service\_internal\discovery_client.py", line 335, in _open_key_file
    win32_file_handle = win32file.CreateFile(
pywintypes.error: (3, 'CreateFile', 'The system cannot find the path specified.')
Traceback (most recent call last):
  File "D:\dev\measurementlink-python\examples\sample_measurement\measurement.py", line 115, in <module>
    main()
  File "D:\dev\measurementlink-python\examples\sample_measurement\.venv\lib\site-packages\click\core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
[...]

PR #397, opened 2023-09-22
Fix pywintypes error during the launch of the discovery service after a fresh install of MeasurementLink

samchris007:

When attempting to open the discovery service key file after a fresh install of MeasurementLink (i.e., folders such as Discovery and Logs are not present in the fresh PC installation), the `winerror.ERROR_PATH_NOT_FOUND` exception was thrown, which hadn't been handled in `_open_key_file()`. This unhandled error resulted in the exception being thrown in the command prompt during the first run of a measurement service alone.

To fix this issue, I have made the following suggestions provided by @bkeryan.
- Updated `_open_key_file` to translate `ERROR_PATH_NOT_FOUND` to `FileNotFoundError`, which is handled by its caller (`_start_service`)
- Updated `_start_service` to catch `OSError`.

Fixes [Bug 2514643](https://dev.azure.com/ni/DevCentral/_workitems/edit/2514643): Auto-launching discovery service fails after clean install (GitHub Issue #369)

Added/Updated unit tests to test the fix.

### DIFF
--- a/ni_measurementlink_service/_internal/discovery_client.py
+++ b/ni_measurementlink_service/_internal/discovery_client.py
@@ -329,7 +329,7 @@ def _start_service(
         try:
             with _open_key_file(str(key_file_path)) as _:
                 return discovery_service_subprocess
-        except IOError:
+        except OSError:
             pass
         if time.time() >= timeout_time:
             raise TimeoutError("Timed out waiting for discovery service to start")
@@ -390,14 +390,17 @@ def _open_key_file(path: str) -> typing.TextIO:
                 None,
             )
         except win32file.error as e:
-            if e.winerror == winerror.ERROR_FILE_NOT_FOUND:
+            if (
+                e.winerror == winerror.ERROR_FILE_NOT_FOUND
+                or e.winerror == winerror.ERROR_PATH_NOT_FOUND
+            ):
                 raise FileNotFoundError(errno.ENOENT, e.strerror, path) from e
             elif (
                 e.winerror == winerror.ERROR_ACCESS_DENIED
                 or e.winerror == winerror.ERROR_SHARING_VIOLATION
             ):
                 raise PermissionError(errno.EACCES, e.strerror, path) from e
-            raise
+            raise WindowsError(errno.ENONET, e.strerror, path) from e
 
         # The CRT file descriptor takes ownership of the Win32 file handle.
         # os.O_TEXT is unnecessary because Python handles newline conversion.
--- a/tests/unit/test_discovery_client.py
+++ b/tests/unit/test_discovery_client.py
@@ -9,6 +9,8 @@ from unittest.mock import Mock
 
 import grpc
 import pytest
+import win32file
+import winerror
 from pytest_mock import MockerFixture
 
 from ni_measurementlink_service._channelpool import GrpcChannelPool
@@ -16,6 +18,7 @@ from ni_measurementlink_service._internal.discovery_client import (
     DiscoveryClient,
     ServiceLocation,
     _get_discovery_service_address,
+    _open_key_file,
     _start_service,
 )
 from ni_measurementlink_service._internal.stubs.ni.measurementlink.discovery.v1.discovery_service_pb2 import (
@@ -223,10 +226,12 @@ def test___get_discovery_service_address___start_service_jit___returns_expected_
     assert _TEST_SERVICE_PORT in discovery_service_address
 
 
+@pytest.mark.parametrize("key_file_error", [IOError, WindowsError])
 def test___get_discovery_service_address___key_file_not_exist___throws_timeouterror(
     mocker: MockerFixture,
     temp_discovery_key_file_path: pathlib.Path,
     temp_registration_json_file_path: pathlib.Path,
+    key_file_error,
 ):
     mocker.patch(
         "ni_measurementlink_service._internal.discovery_client._get_key_file_path",
@@ -236,7 +241,8 @@ def test___get_discovery_service_address___key_file_not_exist___throws_timeouter
         "ni_measurementlink_service._internal.discovery_client._START_SERVICE_TIMEOUT", 5.0
     )
     mocker.patch(
-        "ni_measurementlink_service._internal.discovery_client._open_key_file", side_effect=IOError
+        "ni_measurementlink_service._internal.discovery_client._open_key_file",
+        side_effect=key_file_error,
     )
     mocker.patch(
         "ni_measurementlink_service._internal.discovery_client._get_registration_json_file_path",
@@ -247,6 +253,18 @@ def test___get_discovery_service_address___key_file_not_exist___throws_timeouter
     with pytest.raises(IOError) as exc_info:
         _get_discovery_service_address()
     assert exc_info.type is TimeoutError
+
+
+def test___key_file_not_exist___open_key_file___raises_file_not_found_error(
+    mocker: MockerFixture, temp_discovery_key_file_path: pathlib.Path
+):
+    mocker.patch(
+        "win32file.CreateFile",
+        side_effect=win32file.error(winerror.ERROR_PATH_NOT_FOUND, None, None),
+    )
+
+    with pytest.raises(FileNotFoundError):
+        _open_key_file(temp_discovery_key_file_path)
 
 
 def test___start_discovery_service___key_file_exist_after_poll___service_start_success(


tests/unit/test_discovery_client.py
```diff
@@ -223,7 +228,7 @@ def test___get_discovery_service_address___start_service_jit___returns_expected_
     assert _TEST_SERVICE_PORT in discovery_service_address
 
 
-def test___get_discovery_service_address___key_file_not_exist___throws_timeouterror(
+def test___open_key_file_that_throws_ioerror___get_discovery_service_address___throws_timeouterror(
```
Tharun-Sundar, line 231:
```suggestion
def test___get_discovery_service_address___open_key_file_throws_ioerror___throws_timeouterror(
```

tests/unit/test_discovery_client.py
```diff
@@ -249,6 +254,51 @@ def test___get_discovery_service_address___key_file_not_exist___throws_timeouter
     assert exc_info.type is TimeoutError
 
 
+def test___open_key_file_that_throws_windowserror___get_discovery_service_address___throws_timeouterror(
```
Tharun-Sundar, line 257:
```suggestion
def test___get_discovery_service_address___open_key_file_throws_windowserror___throws_timeouterror(
```

Tharun-Sundar (approved):

dixonjoel:
If we wanted to reproduce this locally and test the fix without having to do a fresh install, would you recommend just deleting the 'Logs' or 'Discovery' folder in programdata?

samchris007:
@dixonjoel 
> If we wanted to reproduce this locally and test the fix without having to do a fresh install, would you recommend just deleting the 'Logs' or 'Discovery' folder in programdata?

Yes, we can terminate any existing running discovery service, delete the `Discovery` folder `C:\ProgramData\National Instruments\MeasurementLink\` and then test this use case

ni_measurementlink_service/_internal/discovery_client.py
```diff
@@ -329,7 +329,7 @@ def _start_service(
         try:
             with _open_key_file(str(key_file_path)) as _:
                 return discovery_service_subprocess
-        except IOError:
+        except OSError:
```
dixonjoel, line 332: What was the scenario where we were getting 'IOError' before and how are we handling this case now? Because it looks like we're no longer catching 'IOError'.

bkeryan: https://docs.python.org/3/library/exceptions.html#ZeroDivisionError

> The following exceptions are kept for compatibility with previous versions; starting from Python 3.3, they are aliases of [OSError](https://docs.python.org/3/library/exceptions.html#OSError).
> 
> - exception EnvironmentError
> - exception IOError
> - exception WindowsError: Only available on Windows.

ni_measurementlink_service/_internal/discovery_client.py
```diff
@@ -390,14 +390,17 @@ def _open_key_file(path: str) -> typing.TextIO:
...
+                or e.winerror == winerror.ERROR_PATH_NOT_FOUND
+            ):
                 raise FileNotFoundError(errno.ENOENT, e.strerror, path) from e
             elif (
                 e.winerror == winerror.ERROR_ACCESS_DENIED
                 or e.winerror == winerror.ERROR_SHARING_VIOLATION
             ):
                 raise PermissionError(errno.EACCES, e.strerror, path) from e
-            raise
+            raise WindowsError(errno.ENONET, e.strerror, path) from e
```
bkeryan, line 403: ENONET does not make any sense. How did you choose this error code? Is it a typo for ENOENT?

https://www.man7.org/linux/man-pages/man3/errno.3.html:

>  ENONET Machine is not on the network.

When I suggested raising WindowsError, I didn't realize it's another alias for OSError. 

> exception OSError(errno, strerror[, filename[, winerror[, filename2]]])
>
> winerror
> Under Windows, this gives you the native Windows error code. The [errno](https://docs.python.org/3/library/exceptions.html#OSError.errno) attribute is then an approximate translation, in POSIX terms, of that native error code.
> 
> Under Windows, if the winerror constructor argument is an integer, the [errno](https://docs.python.org/3/library/exceptions.html#OSError.errno) attribute is determined from the Windows error code, and the errno argument is ignored. On other platforms, the winerror argument is ignored, and the [winerror](https://docs.python.org/3/library/exceptions.html#OSError.winerror) attribute does not exist. 

It turns out that OSError can do all the work for us.

```python
>>> OSError(None, "message", None, winerror.ERROR_FILE_NOT_FOUND)
FileNotFoundError(2, 'message', None, 2)
>>> OSError(None, "message", None, winerror.ERROR_PATH_NOT_FOUND)
FileNotFoundError(2, 'message', None, 3)
>>> OSError(None, "message", None, winerror.ERROR_ACCESS_DENIED)
PermissionError(13, 'message', None, 5)
>>> OSError(None, "message", None, winerror.ERROR_SHARING_VIOLATION)
PermissionError(13, 'message', None, 32)
>>> OSError(None, "message", None, winerror.ERROR_INVALID_FUNCTION)
OSError(22, 'message', None, 1)
```

dixonjoel (approved):

tests/unit/test_discovery_client.py
```diff
@@ -249,6 +254,51 @@ def test___get_discovery_service_address___key_file_not_exist___throws_timeouter
...
+        with pytest.raises(IOError) as exc_info:
+            _get_discovery_service_address()
+        assert exc_info.type is TimeoutError
+
+
+def test___key_file_not_exist___open_key_file___raises_file_not_found_error(
+    mocker: MockerFixture, temp_discovery_key_file_path: pathlib.Path
+):
+    if sys.platform != "win32":
+        pytest.skip(f"Platform {sys.platform} is not supported")
```
bkeryan, line 291: Use the decorator: `@pytest.mark.skipif(sys.platform != "win32", reason="Windows-only test")`

tests/unit/test_discovery_client.py
```diff
@@ -249,6 +254,51 @@ def test___get_discovery_service_address___key_file_not_exist___throws_timeouter
     assert exc_info.type is TimeoutError
 
 
+def test___get_discovery_service_address___open_key_file_throws_windowserror___throws_timeouterror(
+    mocker: MockerFixture,
+    temp_discovery_key_file_path: pathlib.Path,
+    temp_registration_json_file_path: pathlib.Path,
+):
+    if sys.platform != "win32":
+        pytest.skip(f"Platform {sys.platform} is not supported")
```
bkeryan, line 263: The only Windows-specific thing in this test case is `WindowsError`, which we now know is just an alias for `OSError`. If you change this test to use `OSError` or `FileNotFoundError`, I think it will pass on Linux. Change the test name too.

tests/unit/test_discovery_client.py
```diff
@@ -249,6 +254,51 @@ def test___get_discovery_service_address___key_file_not_exist___throws_timeouter
...
+            "ni_measurementlink_service._internal.discovery_client._open_key_file",
+            side_effect=WindowsError,
+        )
+        mocker.patch(
+            "ni_measurementlink_service._internal.discovery_client._get_registration_json_file_path",
+            return_value=temp_registration_json_file_path,
+        )
+        mocker.patch("subprocess.Popen")
+
+        with pytest.raises(IOError) as exc_info:
```
bkeryan, line 282: Search/replace all instances of `IOError` with `OSError`

tests/unit/test_discovery_client.py
```diff
@@ -223,7 +228,7 @@ def test___get_discovery_service_address___start_service_jit___returns_expected_
     assert _TEST_SERVICE_PORT in discovery_service_address
 
 
-def test___get_discovery_service_address___key_file_not_exist___throws_timeouterror(
+def test___get_discovery_service_address___open_key_file_throws_ioerror___throws_timeouterror(
```
bkeryan, line 231: This test name and test___get_discovery_service_address___start_service_jit___returns_expected_value have the given and when clauses swapped.

tests/unit/test_discovery_client.py
```diff
@@ -249,6 +254,51 @@ def test___get_discovery_service_address___key_file_not_exist___throws_timeouter
...
+
+def test___key_file_not_exist___open_key_file___raises_file_not_found_error(
+    mocker: MockerFixture, temp_discovery_key_file_path: pathlib.Path
+):
+    if sys.platform != "win32":
+        pytest.skip(f"Platform {sys.platform} is not supported")
+    else:
+        mocker.patch(
+            "win32file.CreateFile",
+            side_effect=win32file.error(winerror.ERROR_PATH_NOT_FOUND, None, None),
```
bkeryan, line 295: You can parameterize this on winerror, in order to test ERROR_FILE_NOT_FOUND as well.

bkeryan (changes requested):

tests/unit/test_discovery_client.py
bkeryan: Actually, we already have a test that always throws IOError, which you will change to OSError. So maybe this test case isn't needed.

ni_measurementlink_service/_internal/discovery_client.py
samchris007: Sure. I have changed all the existing raised errors to OSError with appropriate error codes.

tests/unit/test_discovery_client.py
samchris007: Sure, I removed the test since I have changed all occurrences of `IOError` to `OSError`.

ni_measurementlink_service/_internal/discovery_client.py
bkeryan: I don't think you understand. We don't need any of these cases because `OSError` already handles these. Replace all of the cases with an `OSError`:

```python
        except win32file.error as e:
            raise OSError(None, e.strerror, path, e.winerror) from e
```

Don't replace `e.winerror` with `ERROR_INVALID FUNCTION`. That doesn't make any sense.

tests/unit/test_discovery_client.py
bkeryan: https://github.com/ni/measurementlink-python/pull/404

tests/unit/test_discovery_client.py
```diff
@@ -236,19 +240,35 @@ def test___get_discovery_service_address___key_file_not_exist___throws_timeouter
...
     assert exc_info.type is TimeoutError
 
 
+@pytest.mark.skipif(sys.platform != "win32", reason="Windows-only test")
+@pytest.mark.parametrize(
+    "windows_error_code", [2, 3]
+)  # ERROR_FILE_NOT_FOUND = 2, ERROR_PATH_NOT_FOUND = 3
+def test___key_file_not_exist___open_key_file___raises_file_not_found_error(
+    mocker: MockerFixture, temp_discovery_key_file_path: pathlib.Path, windows_error_code
+):
```
bkeryan, line 262:
```suggestion
def test___key_file_not_exist___open_key_file___raises_file_not_found_error(
    mocker: MockerFixture, temp_discovery_key_file_path: pathlib.Path, windows_error_code: int
) -> None:
```

bkeryan (approved):

tests/unit/test_discovery_client.py
```diff
@@ -33,6 +34,9 @@
 from ni_measurementlink_service.measurement.info import MeasurementInfo, ServiceInfo
 from tests.utilities.fake_rpc_error import FakeRpcError
 
+if sys.platform.startswith("win"):
+    import win32file
```
bkeryan, line 38:
```
Run poetry run mypy tests
tests/unit/test_discovery_client.py:265: error: Name "win32file" is not defined  [name-defined]
Found 1 error in 1 file (checked 33 source files)
```

```suggestion
if sys.platform == "win32":
    import win32file
```

samchris007: Actually, I tried both, and I can see in the pipeline that the tests are failing in both scenarios.

bkeryan: nevermind, this part is fine

tests/unit/test_discovery_client.py
```diff
@@ -236,19 +240,35 @@ def test___get_discovery_service_address___key_file_not_exist___throws_timeouter
...
+@pytest.mark.skipif(sys.platform != "win32", reason="Windows-only test")
+@pytest.mark.parametrize(
+    "windows_error_code", [2, 3]
+)  # ERROR_FILE_NOT_FOUND = 2, ERROR_PATH_NOT_FOUND = 3
+def test___key_file_not_exist___open_key_file___raises_file_not_found_error(
+    mocker: MockerFixture, temp_discovery_key_file_path: pathlib.Path, windows_error_code
+):
+    mocker.patch(
+        "win32file.CreateFile",
+        side_effect=win32file.error(windows_error_code, None, None),
```
bkeryan, line 265:
```
Run poetry run mypy tests
tests/unit/test_discovery_client.py:265: error: Name "win32file" is not defined  [name-defined]
Found 1 error in 1 file (checked 33 source files)
```

I guess we need `if sys.platform == "win32"` around either this code or the entire test body because mypy doesn't interpret `@pytest.skipif`.

samchris007: Sure. Included the check `if sys.platform == "win32"`

tests/unit/test_discovery_client.py
```diff
@@ -236,19 +240,37 @@ def test___get_discovery_service_address___key_file_not_exist___throws_timeouter
...
+)  # ERROR_FILE_NOT_FOUND = 2, ERROR_PATH_NOT_FOUND = 3
+def test___key_file_not_exist___open_key_file___raises_file_not_found_error(
+    mocker: MockerFixture, temp_discovery_key_file_path: pathlib.Path, windows_error_code: int
+) -> None:
+    if sys.platform != "win32":
+        pytest.skip("Windows-only test")
+    else:
+        mocker.patch(
+            "win32file.CreateFile",
+            side_effect=win32file.error(windows_error_code, None, None),
```
bkeryan, line 267: This is causing mypy errors in my dev branches.

https://github.com/ni/measurementlink-python/pull/422

bkeryan: I'm updating the PR workflow to check Windows-specific code.

https://github.com/ni/measurementlink-python/pull/423

samchris007: Sure.

PR #398, opened 2023-09-22
Remove use_simulation_options command line parameter from examples that don't use it

### LINKED ISSUE #395: `use_simulation_option` cmd line option in examples has no effect
# Bug Report

The `use_simulation_option` (aka `--use-simulation`/`--no-use-simulation`) cmd line option has no effect on scope example, and possibly other examples. Even when the `--no-use-simulation` option was specified when starting up the measurement service IO Trace showed session initialization was specifying `Simulate=1`, and the captured data from the scope was clearly simulated data. This would occur as long as `USE_SIMULATION` in `_constants.py` was true.

## Repro or Code Sample

The main culprit from what I can gather is that this line in the [_niscope_helpers.py](https://github.com/ni/measurementlink-python/blob/main/examples/niscope_acquire_waveform/_niscope_helpers.py#L19) which relies solely on a value defined in `_constants.py`. Since this doesn't check the values coming from the cmd line there's no way for it to be accounted for during session creation.

## Expected Behavior

I would expect `--use-simulation`/`--no-use-simulation` to either affect the session behavior, or to not exist as a cmd line arg since this behavior appears to be controlled by the `USE_SIMULATION` value in `_constants.py`

## Current Behavior

`py measurement.py --help` will list `--use-simulation`/`--no-use-simulation` as potential args but these args don't have any affect on the session creation behavior.

e.g.
`USE_SIMULATION = true` + `--no-use-simulation` will create simulated sessions
`USE_SIMULATION = false` + `--use-simulation` will create real hw sessions

## Possible Solution

Remove `use_simulation_option` as a cmd line arg since this behavior is already controlled by a value in another file.

## Your Environment

* OS & Device: Windows PC
* `ni-measurementlink-service` version [e.g. 0.12.3]
* MeasurementLink version =>23.8.0d697
* Python version => 3.11

[AB#2529511](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2529511)

dixonjoel:

Fixes #395. Remove the 'use_simulation' command-line option that wasn't working and replaces all of it with the USE_SIMULATION compile-time constant in _constants.py.

Further refactoring / renaming will be done in #408 

Incorrect command-line option that was never used.

poetry run pytest -v

### DIFF
--- a/examples/nidaqmx_analog_input/README.md
+++ b/examples/nidaqmx_analog_input/README.md
@@ -16,7 +16,6 @@ measurement with NI-DAQmx.
     registration and unregistration in the `Setup` and `Cleanup` sections of the main
     sequence. For **Test UUTs** and batch process model use cases, these steps should
     be moved to the `ProcessSetup` and `ProcessCleanup` callbacks.
-- Uses the NI gRPC Device Server to allow sharing instrument sessions with other measurement services when running measurements from TestStand.
 
 > **Note**
 >
--- a/examples/game_of_life/_helpers.py
+++ b/examples/game_of_life/_helpers.py
@@ -27,15 +27,12 @@ class ServiceOptions(NamedTuple):
     use_grpc_device: bool = False
     grpc_device_address: str = ""
 
-    use_simulation: bool = False
-
 
 def get_service_options(**kwargs: Any) -> ServiceOptions:
     """Get service options from keyword arguments."""
     return ServiceOptions(
         use_grpc_device=kwargs.get("use_grpc_device", False),
         grpc_device_address=kwargs.get("grpc_device_address", ""),
-        use_simulation=kwargs.get("use_simulation", False),
     )
 
 
@@ -211,16 +208,6 @@ def grpc_device_options(func: F) -> F:
         help="NI gRPC Device Server address (e.g. localhost:31763). If unspecified, use the discovery service to resolve the address.",
     )
     return grpc_device_address_option(use_grpc_device_option(func))
-
-
-def use_simulation_option(default: bool) -> Callable[[F], F]:
-    """Decorator for --use-simulation command line option."""
-    return click.option(
-        "--use-simulation/--no-use-simulation",
-        default=default,
-        is_flag=True,
-        help="Use simulated instruments.",
-    )
 
 
 def get_grpc_device_channel(
--- a/examples/nidaqmx_analog_input/_helpers.py
+++ b/examples/nidaqmx_analog_input/_helpers.py
@@ -36,15 +36,12 @@ class ServiceOptions(NamedTuple):
     use_grpc_device: bool = False
     grpc_device_address: str = ""
 
-    use_simulation: bool = False
-
 
 def get_service_options(**kwargs: Any) -> ServiceOptions:
     """Get service options from keyword arguments."""
     return ServiceOptions(
         use_grpc_device=kwargs.get("use_grpc_device", False),
         grpc_device_address=kwargs.get("grpc_device_address", ""),
-        use_simulation=kwargs.get("use_simulation", False),
     )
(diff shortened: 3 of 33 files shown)


bkeryan:
> I left it in 'output_voltage_measurement' and 'ni_visa_dmm' measurement because these lines look like they're using it.
> https://github.com/ni/measurementlink-python/blob/main/examples/output_voltage_measurement/measurement.py#L129
> https://github.com/ni/measurementlink-python/blob/main/examples/nivisa_dmm_measurement/measurement.py#L96

If we aren't going to make the effort to keep this command line option working, then let's remove it entirely from all the examples and `_helpers.py`, not just from the examples where we broke it. It's weird to keep it around for two examples just because those are the two we didn't mess up.

@cfuentes-ni @rfriedma Do you agree? All examples should require editing source code to disable simulation?

bkeryan:
It's also worth reevaluating whether the command line options to disable NI gRPC Device Server and/or hardcode its address/port are still useful.

@cfuentes-ni @rfriedma

cfuentes-ni:
> > I left it in 'output_voltage_measurement' and 'ni_visa_dmm' measurement because these lines look like they're using it.
> > https://github.com/ni/measurementlink-python/blob/main/examples/output_voltage_measurement/measurement.py#L129
> > https://github.com/ni/measurementlink-python/blob/main/examples/nivisa_dmm_measurement/measurement.py#L96
> 
> If we aren't going to make the effort to keep this command line option working, then let's remove it entirely from all the examples and `_helpers.py`, not just from the examples where we broke it. It's weird to keep it around for two examples just because those are the two we didn't mess up.

I agree with Brad for now, and we can revisit once we have the updated session management API if we need to.

dixonjoel:
> It's also worth reevaluating whether the command line options to disable NI gRPC Device Server and/or hardcode its address/port are still useful.

@cfuentes-ni does your thumbs up mean to remove these two options?

If I do that, it'll be in a separate PR.

examples/nidaqmx_analog_input/README.md
```diff
@@ -16,7 +16,6 @@ measurement with NI-DAQmx.
     registration and unregistration in the `Setup` and `Cleanup` sections of the main
     sequence. For **Test UUTs** and batch process model use cases, these steps should
     be moved to the `ProcessSetup` and `ProcessCleanup` callbacks.
-- Uses the NI gRPC Device Server to allow sharing instrument sessions with other measurement services when running measurements from TestStand.
```
dixonjoel, line 19: This is duplicated above.

dixonjoel:
@bkeryan @csjall Can we go ahead with this PR? More refactoring can be done when we start to use the .env file for configuration, but this removes some of the code that will need to go anyway.

bkeryan:
> @bkeryan @csjall Can we go ahead with this PR? More refactoring can be done when we start to use the .env file for configuration, but this removes some of the code that will need to go anyway.

https://github.com/ni/measurementlink-python/pull/442 removes the command line options that are replaced with variables in `.env`, including `--use-grpc-device` and `--grpc-device-address`. It doesn't include the `_helpers.py` churn, but soon there will be more stuff to remove, so I think `_helpers.py` can wait.

dixonjoel:
> > @bkeryan @csjall Can we go ahead with this PR? More refactoring can be done when we start to use the .env file for configuration, but this removes some of the code that will need to go anyway.
> 
> #442 removes the command line options that are replaced with variables in `.env`, including `--use-grpc-device` and `--grpc-device-address`. It doesn't include the `_helpers.py` churn, but soon there will be more stuff to remove, so I think `_helpers.py` can wait.

@bkeryan ok, I'll just go ahead and close this one.

PR #399, opened 2023-09-22
Importing nidcpower.session into output voltage measurement

DelpireNI:

When running the OutputVoltageMeasurement I got the following error:

> Traceback (most recent call last):
  File "C:\Users\DelpireTest1\Desktop\measurementlink-python\measurementlink-python\examples\output_voltage_measurement\measurement.py", line 177, in <module>
    channels: nidcpower._SessionBase,
AttributeError: module 'nidcpower' has no attribute '_SessionBase'

~I am not sure what has happened with `nidcpower._SessionBase`. A search in the ni/nimi-python repo shows this class for our other drivers, DMM, Scope, Switch, etc., but I do not see it for nidcpower.~

Importing nidcpower.session and fixing type hint.

Currently it seems the measurement cannot be run out of the box. This should fix it.

Was able to run the measurement successfully and it worked with both InstrumentStudio and TestStand.

### DIFF
--- a/examples/output_voltage_measurement/measurement.py
+++ b/examples/output_voltage_measurement/measurement.py
@@ -182,7 +182,7 @@ def _get_session_info_for_pin(
 
 def _wait_for_source_complete_event(
     measurement_service: nims.MeasurementService,
-    channels: nidcpower._SessionBase,
+    channels: nidcpower.Session,
     cancellation_event: threading.Event,
 ) -> None:
     deadline = time.time() + measurement_service.context.time_remaining
--- a/examples/output_voltage_measurement/measurement.py
+++ b/examples/output_voltage_measurement/measurement.py
@@ -1,7 +1,5 @@
 """ Source DC voltage as input with an NI SMU and measure output using NI-VISA DMM."""
 
-from __future__ import annotations
-
 import logging
 import pathlib
 import threading
@@ -15,6 +13,7 @@ import click
 import grpc
 import hightime
 import nidcpower
+import nidcpower.session
 import pyvisa
 from _constants import USE_SIMULATION
 from _helpers import (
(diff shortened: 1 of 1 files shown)


bkeryan:
I don't know what caused this. `_SessionBase` is still right there in the nimi-python main branch:

https://github.com/ni/nimi-python/blob/4437fdd53856ba886d543aa01dd6fc86ada383c9/generated/nidcpower/nidcpower/session.py#L95

I filed a bug about the fact that this public API returns an internal type: https://github.com/ni/nimi-python/issues/2019

examples/output_voltage_measurement/measurement.py
```diff
@@ -182,7 +182,7 @@ def _get_session_info_for_pin(
 
 def _wait_for_source_complete_event(
     measurement_service: nims.MeasurementService,
-    channels: nidcpower._SessionBase,
+    channels: nidcpower.Session,
```
bkeryan, line 185: This is incorrect. `session.channels[x]` returns a `_SessionBase`, not a `Session`. 

It only passes type checking because nimi-python doesn't have type hints and `mypy --ignore-missing-imports` treats missing modules/classes/methods as `Any`.

A better fix is to add `from __future__ import annotations` so that Python doesn't try to pass type hints at run time. However, I don't understand why this ever failed.

bkeryan (changes requested):

examples/output_voltage_measurement/measurement.py
DelpireNI: Strange. I wonder why _SessionBase didn't show up when I searched the repo. What you linked to above wasn't the main branch, but switching to the main branch it is there too. 

I'll try you suggested change and see if I can run the measurement.

bkeryan: Also, using the wrong type will cause editors like VS Code and PyCharm to offer the wrong auto-complete suggestions. For example, the `self_test` method is defined by `Session` but not `_SessionBase`, and using the wrong type will allow your editor to auto-complete a call to `channels.self_test`.

DelpireNI: Adding `from __future__ import annotations` fixes the issue.

bkeryan (approved):

PR #402, opened 2023-09-22
service: Add MeasurementContext.reserve_session(s)

bkeryan:

ni-measurementlink-service:
- Add `reserve_session` and `reserve_sessions` methods to the public `MeasurementContext` class
  - These have the same signature as in `SessionManagementClient`, except the `pin_map_context` parameter is removed
- Add a `session_management_client` property to the public `MeasurementService` class
- Enable the `MeasurementContext` to get a reference to the owning `MeasurementService`
  - Add an internal `_measurement_service` property to the public `MeasurementContext` class
  - Pass an opaque `owner` to the internal `GrpcService`, `GrpcServicer`, and `MeasurementServiceContext` classes
  - When saving the `owner` as an object attribute, use a weak reference to avoid introducing a reference cycle

tests:
- Add acceptance tests for session management
  - These use the real MeasurementLink session management and pin map services.
- Add a `pin_aware_measurement` test service
- Add a `PinMapClient` to the test utilities
  - This is an improved version of the one in `_helpers.py`: it can use the discovery client and gRPC channel pool.
  - In a separate PR, I would like to move this into a shipping package and delete the one in `_helpers.py`.
- Add `__init__.py` files to the acceptance/integration/unit directories so that `test_*.py` names don't have to be unique.
- Add unit tests for `MeasurementContext`

This is part of the updated session reservation API.

Ran `poetry run pytest -v`.
Updated `nidmm_measurement` to use the new APIs and manually tested.


ni_measurementlink_service/_featuretoggles.py
```diff
@@ -0,0 +1,2 @@
+SESSION_MANAGEMENT_2024Q1 = True
```
bkeryan, line 1: hmm, I should test setting this to False

ni_measurementlink_service/_internal/grpc_servicer.py
```diff
@@ -33,11 +34,13 @@ def __init__(
         self,
         grpc_context: grpc.ServicerContext,
         pin_map_context: PinMapContext,
+        owner: Optional[weakref.ref[object]],
```
bkeryan, line 37: Python 3.8 can't handle this syntax.

```
ImportError while loading conftest '/home/runner/work/measurementlink-python/measurementlink-python/tests/conftest.py'.
tests/conftest.py:[9](https://github.com/ni/measurementlink-python/actions/runs/6278845153/job/17053461040?pr=402#step:7:10): in <module>
    from ni_measurementlink_service._internal.discovery_client import (
ni_measurementlink_service/__init__.py:12: in <module>
    from ni_measurementlink_service.measurement.service import (
ni_measurementlink_service/measurement/service.py:34: in <module>
    from ni_measurementlink_service._internal import grpc_servicer
ni_measurementlink_service/_internal/grpc_servicer.py:30: in <module>
    class MeasurementServiceContext:
ni_measurementlink_service/_internal/grpc_servicer.py:37: in MeasurementServiceContext
    owner: Optional[weakref.ref[object]],
E   TypeError: 'type' object is not subscriptable
Error: Process completed with exit code 4.
```

ni_measurementlink_service/measurement/service.py
```diff
@@ -93,6 +99,101 @@ def abort(self, code: grpc.StatusCode, details: str) -> None:
...
+            instrument_type_id: Optional[str] = None,
+            timeout: Optional[float] = 0.0,
+        ) -> MultiSessionReservation:
+            """Reserve multiple sessions.
+
+            Reserve sessions matching the given pins, sites, and instrument type ID and return the
+            information needed to create or access the sessions.
+
+            Args:
+                context: Includes the pin map ID for the pin map in the Pin Map Service,
```
dixonjoel, line 164: Does 'context' refer to 'self', i.e. this as a MeasurementContext? You didn't include this as an arg in the `reserve_session` method.

dixonjoel (approved):

ni_measurementlink_service/measurement/service.py
bkeryan: No, this is copied from session_management.py where context means PinMapContext.

It's weird that Sphinx didn't error or warn about this.

ni_measurementlink_service/_featuretoggles.py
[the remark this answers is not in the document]
bkeryan: We can add those capabilities at run time (e.g. read an environment variable or registry key) but I don't know how to add them at type check time because the type checker doesn't run code. So maybe all features toggles should be enabled at type check time and opt-in at run time.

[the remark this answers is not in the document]
bkeryan: Wait, this might work: https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-always-true

ni_measurementlink_service/_featuretoggles.py
mshafer-NI, line 1: Consider using python-decouple: https://pypi.org/project/python-decouple/

ni_measurementlink_service/measurement/service.py
```diff
@@ -93,6 +99,101 @@ def abort(self, code: grpc.StatusCode, details: str) -> None:
...
 
+    @property
+    def _measurement_service(self) -> MeasurementService:
+        owner = grpc_servicer.measurement_service_context.get().owner
+        assert isinstance(owner, MeasurementService)
+        return owner
+
+    if _featuretoggles.SESSION_MANAGEMENT_2024Q1:
+
+        def reserve_session(
```
pbirkhol-ni, line 110: This function signature is different from LabVIEW's in a few ways:
- `pin_or_relay_names` is required in LabVIEW. 
- `instrument_type_id` is not an option in LabVIEW. This also deviates from your HLD, which specifically calls out removing `instrument_type_id`.

It seems like what you're proposing tries to cater to the code module use case more than it should and thus I would recommend removing `instrument_type_id`. I'm less sure about `pin_or_relay_names`. If we think the common use case is to reserve all of the pins in a pin map, then you can make an argument for `pin_or_relay_names` being optional. However, I don't know that that is the case.

I also don't think python and LabVIEW necessarily have to match, but I don't see a compelling reason for them to differ in this case. This isn't a difference in language features since both languages can support optional/required inputs.

pbirkhol-ni (changes requested):

ni_measurementlink_service/measurement/service.py
bkeryan: Oops, I copied from the SessionManagementClient class and not from the HLD

ni_measurementlink_service/_featuretoggles.py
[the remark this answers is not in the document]
bkeryan: https://github.com/ni/measurementlink-python/pull/421

ni_measurementlink_service/_featuretoggles.py
bkeryan: https://github.com/ni/measurementlink-python/pull/421

ni_measurementlink_service/measurement/service.py
bkeryan: I also added a runtime check that pin_or_relay_names isn't empty.

ni_measurementlink_service/_featuretoggles.py
```diff
@@ -141,3 +141,11 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _T:
         return wrapper
 
     return decorator
+
```
dixonjoel, line 144: Is this ready to be enabled by a decouple .env file? I'm just checking my understanding of where we're at on that front. And if so, would you put "MEASUREMENTLINK_SESSION_MANAGEMENT_2024Q1_ENABLE" in the file? Where would this file be placed?

dixonjoel (approved):

ni_measurementlink_service/_featuretoggles.py
```diff
@@ -141,3 +141,11 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _T:
...
+
+# --------------------------------------
+# Define feature toggle constants here:
+# --------------------------------------
+
+SESSION_MANAGEMENT_2024Q1 = FeatureToggle("SESSION_MANAGEMENT_2024Q1", CodeReadiness.INCOMPLETE)
```
mshafer-NI, line 150:
```suggestion
SESSION_MANAGEMENT_2024Q1 = FeatureToggle("SESSION_MANAGEMENT_2024Q1", CodeReadiness.INCOMPLETE, cast=bool)
```

ni_measurementlink_service/_featuretoggles.py
mshafer-NI: Speaking to how python-decouple works:

> And if so, would you put "MEASUREMENTLINK_SESSION_MANAGEMENT_2024Q1_ENABLE" in the file?

You would put:
```
SESSION_MANAGEMENT_2024Q1=True
```
(or "true", or "T", "1" -> these are all cast to boolean True, while basically any other value is cast to False)

Anywhere from your cwd up to the next to the .git folder at the top of your Git repo (or, if you're not in a Git repo, up to the top of your current drive).

# Also

`python-decouple` evaluation process:
find and load a `.env` or `setttings.ini` file (IIRC, first found wins, but I don't remember order of preference between the file names). Examine the environment variables, and overwrite anything from the file with those values.

When asked (via the function call), perform a string fetch (or cast/convert to the desired type)

ni_measurementlink_service/_featuretoggles.py
mshafer-NI: nope, you have this in the wrapper class already.

ni_measurementlink_service/_featuretoggles.py
dixonjoel: This line appears to be looking for MEASUREMENTLINK_ENABLE_<feature toggle name>:
https://github.com/ni/measurementlink-python/pull/402/files#diff-bf2ce48b68e8ea6ae8c565e0145a4198b87e3574c327848571ae34e7fd023f40R102

bkeryan: Yes, set `MEASUREMENTLINK_ENABLE_SESSION_MANAGEMENT_2024Q1=1`. Or `=True`. It accepts both.

I am going to create an `.env.sample` file for configuration settings, but I'm not sure I want to put all of the pre-release feature toggles in it.

mshafer-NI: oh, because we did a wrapper...

ok, so..., `_PREFIX` is `MEASUREMENTLINK`, and the string passed in is the `name`, so it would be:
`if config(f"{_PREFIX}_ENABLE_{name}", default=False, cast=bool):`
turns into
`if config(f"MEASUREMENTLINK_ENABLE_SESSION_MANAGEMENT_2024Q1", default=False, cast=bool):`

so then the .env file or environment vars would look like:
```
MEASUREMENTLINK_ENABLE_SESSION_MANAGEMENT_2024Q1=True
```

bkeryan: > while basically any other value is cast to False

It only accepts certain false values.
https://github.com/HBNetwork/python-decouple/blob/master/decouple.py#L29

mshafer-NI: > It only accepts certain false values.
> https://github.com/HBNetwork/python-decouple/blob/master/decouple.py#L29

Thank you for the correction, I've updated the original note.

bkeryan: Ok, I added `.env.sample` and tested it (which revealed that the default search path isn't what we want)

dixonjoel: Looks good! I like it. On the naming of the file, it just has to start with .env? Just want to be clear what's allowed. E.g. does '.environment.mysettings' work? Or settings.env?

mshafer-NI: Just `.env` or `settings.ini` -> https://github.com/HBNetwork/python-decouple/blob/7c9b250c53bdd0ae96f449d5d2be918ed90ba273/decouple.py#L202C1-L205C7

https://pypi.org/project/python-decouple/#usage

dixonjoel: That code seems to mean ".env.sample" won't work? So should we add some instruction in the .env.sample to say "Rename this file to .env"?

pbirkhol-ni (approved):

ni_measurementlink_service/_featuretoggles.py
mshafer-NI: or in the readme? -> but yeah, I would expect so.

bkeryan: I updated the instructions in `.env.sample` and will update the wiki in a bit.

I'm going to hold off on updating README.md until we update the examples.

mshafer-NI: Slight correction to my earlier comment, `decouple.config` apparently loads the file lazily: 
> [Since config is lazy and only opens the configuration file when it’s first needed, you have the chance to change its encoding right after import.](https://pypi.org/project/python-decouple/#:~:text=Since%20config%20is%20lazy%20and%20only%20opens%20the%20configuration%20file%20when%20it%E2%80%99s%20first%20needed%2C%20you%20have%20the%20chance%20to%20change%20its%20encoding%20right%20after%20import.)

Issue #410, opened 2023-09-25
VISA DMM examples cut off out-of-range values with 34401A DMM


bkeryan:
# Bug Report

The 34401A DMM represents out-of-range as 9.9e+37. The `.measui` files for the VISA DMM examples (`nivisa_dmm_measurement` and `output_voltage_measurement`) display 9.9e+37 as 99000000000000000000000000000000000000.000000, which does not fit in the **Measured value** control.

## Repro or Code Sample

- Install a VISA DMM.
- Input a voltage that is out of range.
- For a specific example, `nivisa_dmm_measurement` or `output_voltage_measurement`:
  - Configure the pin map to use the correct resource.
  - Set USE_SIMULATION = False
  - Run the example.

## Expected Behavior

**Measured value** displays using scientific or exponent format, such as "9.9e+37" or "99e+36".

## Current Behavior

**Measured value** displays 36 zeroes before the decimal point, which doesn't fit.

![image](https://github.com/ni/measurementlink-python/assets/17914399/449ef267-16fb-4086-a1e9-c6627b33459a)

## Possible Solution

Change display format in the `.measui` file.

## Context

Testing with real hardware.

## Your Environment

* OS & Device: Windows
* `ni-measurementlink-service` version: releases/1.2 branch
* MeasurementLink version: 23.8.0d731
* Python version: 3.9.13

[AB#2532272](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2532272)

PR #420, opened 2023-09-26
Changing measured value indicator to use significant digits instead of digits of precision

DelpireNI:

- Changes NI VISA DMM, NI DMM, Output Measurement Voltage examples' measui's "Measured value" indicator to use significant digits instead of digits of precision.
- They will have 7 significant digits.
- Updated width of indicators so that 7 digits and the E+XX fit well.
- Ensured left and right margins were the same.
- Updated output voltage measurement example so that the measurement voltage indicator is aligned vertically with the inputs.

Here is a demonstration of the issue:
image

Here are the updated UI in InstrumentStudio:
NI-VISA DMM
image

NI-DMM
image

Output Voltage Measurement:
image

See (#410)
Very large numbers take up too much space. It would be better to use scientific notation which the significant digits setting does.

Ran example in the measui editor.
Updated python measurement to force it to return huge number. Huge numbers are now in scientific notation.


dixonjoel (approved):

examples/nivisa_dmm_measurement/NIVisaDmmMeasurement.measui
```diff
@@ -1,31 +1,31 @@
...
+			<ChannelNumericText AdaptsToType="[bool]True" BaseName="[string]Numeric" Channel="[string]{26c56ad8-e31a-456d-979e-87865b3f5c61}/Output/measured_value" Height="[float]51" Id="b8c14ae283ec4bf8ad8a037705e08afc" IsLabelBoundToChannel="[bool]False" IsReadOnly="[bool]True" Label="[UIModel]71a08f4917d143caa1a9edb08cfff3c8" Left="[float]195" MinHeight="[float]51" TabIndex="[int]0" Top="[float]30" UnitAnnotation="[string]" ValueFormatter="[string]DisplayFormat=Automatic:Digits=6:DigitDisplayType=SignificantDigits:MinimumFieldWidth=0:AlwaysShowSign=False:ShowThousandsSeparator=False" ValueType="[Type]Double" Width="[float]215">
```
bkeryan, line 20: 6 decimal digits is not equivalent to 6 significant digits because significant digits also includes the left of the decimal point. That's why the display changed from "1.234560" to "1.23456".

Look at the 34401A LV example. I think I used 7 significant digits due to this.

examples/nivisa_dmm_measurement/NIVisaDmmMeasurement.measui
bkeryan, line 1: Please check/update `nidmm_measurement` and `output_voltage_measurement` too.

`nidmm_measurement` doesn't use the 9.9e37 format for out of range but it could still make sense to use exponential format.

`output_voltage_measurement` is a 2nd VISA DMM example, combined with an SMU example.

examples/nivisa_dmm_measurement/NIVisaDmmMeasurement.measui
bkeryan, line 20: Does this look the same in InstrumentStudio? When I filed the bug, I experimented with the display format settings and saw some weird output like "9.9e+13 Y" (yottavolt?)

bkeryan (changes requested):

examples/nivisa_dmm_measurement/NIVisaDmmMeasurement.measui
bkeryan, line 1: The screenshots show almost no margin on the right side of the panel. I think it is better for the left and right sides to have about the same margin.

examples/nivisa_dmm_measurement/NIVisaDmmMeasurement.measui
```diff
@@ -1,31 +1,31 @@
...
-				<RingSelectorInfo DisplayValue="[string]DC_VOLTS" IsEnabled="[bool]True" Value="[int]1" ValueType="[Type]Int32" xmlns="http://www.ni.com/Controls.LabVIEW.Design" />
-				<RingSelectorInfo DisplayValue="[string]AC_VOLTS" IsEnabled="[bool]True" Value="[int]2" ValueType="[Type]Int32" xmlns="http://www.ni.com/Controls.LabVIEW.Design" />
+			<ChannelEnumSelector AdaptsToType="[bool]True" AllowNonSequentialValues="[bool]True" BaseName="[string]Enum" Channel="[string]{26c56ad8-e31a-456d-979e-87865b3f5c61}/Configuration/measurement_type" Enabled="[bool]True" Height="[float]24" Id="dada7f23e27f4f468c0faf48553a7b66" IsLabelBoundToChannel="[bool]False" Label="[UIModel]570f36c6f42d4eb6a90f7e8a826ab54d" Left="[float]23" Top="[float]85" Width="[float]150" xmlns="http://www.ni.com/InstrumentFramework/ScreenDocument">
+				<RingSelectorInfo DisplayValue="[string]DC_VOLTS" IsEnabled="[bool]True" Value="[int]0" xmlns="http://www.ni.com/Controls.LabVIEW.Design" />
```
bkeryan, line 27: Hmm, this must be from an earlier version of the service that didn't have a 0 enum value.

examples/nivisa_dmm_measurement/NIVisaDmmMeasurement.measui
DelpireNI: image
Looks fine in IS to me.

DelpireNI: I will update to 7. 
image

You did use 7 for that example:
image

examples/nivisa_dmm_measurement/NIVisaDmmMeasurement.measui
DelpireNI: Yeah I wasn't sure about this change.

examples/nivisa_dmm_measurement/NIVisaDmmMeasurement.measui
DelpireNI: I have updated all three examples to ensure the margins on the left and right are the same.

DelpireNI: I have updated the NI DMM and output voltage measurement examples.

examples/nivisa_dmm_measurement/NIVisaDmmMeasurement.measui
DelpireNI: Is there any action I need to take here?

bkeryan: I don't think so.

bkeryan (approved):

PR #421, opened 2023-09-26
service: Add feature toggle support

bkeryan:

Add an internal `_featuretoggles` module which allows defining feature toggles, checking them at run time, and decorating functions to indicate that a feature toggle is required. You can control which feature toggles are enabled by setting environment variables or by setting variables in a `.env` file (implemented using `python-decouple`).

> **Note** I considered having the decorator add text to the function's docstring like `@deprecation.deprecated` does, but it turns out that this doesn't work with `sphinx-autoapi`. We may need to add filtering to `conf.py`.

Add pytest markers `disable_feature_toggle` and `enable_feature_toggle` that can be used to disable or enable feature toggles for specific test functions or test modules.

It provides basic run-time feature toggle support with similar conventions to those used by NI's application software.

Wrote unit test.


ni_measurementlink_service/_featuretoggles.py
```diff
@@ -0,0 +1,93 @@
...
+    _T = TypeVar("_T")
+
+
+class CodeReadiness(Enum):
+    """Indicates whether code is ready to be supported."""
+
+    RELEASE = 0
+    NEXT_RELEASE = 1
+    INCOMPLETE = 2
+    PROTOTYPE = 3
```
bkeryan, line 27: This uses the same enum ordering as Core. grpc-device [reverses this ordering](https://github.com/ni/grpc-device/blob/main/source/server/feature_toggles.h).

ni_measurementlink_service/_featuretoggles.py
```diff
@@ -0,0 +1,101 @@
...
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+_ENV_VAR_PREFIX = "MEASUREMENTLINK"
+_ALLOW_INCOMPLETE: bool = config(f"{_ENV_VAR_PREFIX}_ALLOW_INCOMPLETE", default=False, cast=bool)
+_ALLOW_NEXT_RELEASE: bool = config(
```
bkeryan, line 92: Would it be better to have a `MEASUREMENTLINK_CODE_READINESS` variable that you set to `INCOMPLETE`, etc.?

ni_measurementlink_service/_featuretoggles.py
```diff
@@ -0,0 +1,98 @@
...
+        # Only read the env var at initialization time.
+        if config(f"{_PREFIX}_ENABLE_{name}", default=False, cast=bool):
+            self._is_enabled_override = True
+
+    @property
+    def is_enabled(self) -> bool:
+        """Indicates whether the feature is currently enabled."""
+        if self._is_enabled_override is not None:
+            return self._is_enabled_override
+        return self.readiness.value <= READINESS_LEVEL.value
```
bkeryan, line 57: It would make sense for the enum to support comparisons.

tests/unit/test_featuretoggles.py
```diff
@@ -0,0 +1,98 @@
...
+def _prototype_function(x: int, y: str, z: List[int]) -> str:
+    return _prototype_function_impl(x, y, z)
+
+
+def _prototype_function_impl(x: int, y: str, z: List[int]) -> str:
+    return ""
+
+
+def test___current_readiness_level___is_enabled___reflects_readiness_level() -> None:
+    level = _featuretoggles.READINESS_LEVEL
```
bkeryan, line 30: `from _featuretoggles import READINESS_LEVEL` doesn't work with patching. Perhaps we should expose it as a function instead.

tests/unit/test_featuretoggles.py
```diff
@@ -0,0 +1,98 @@
...
+
+from ni_measurementlink_service import _featuretoggles
+from ni_measurementlink_service._featuretoggles import (
+    CodeReadiness,
+    FeatureNotSupportedError,
+    FeatureToggle,
+    requires_feature,
+)
+
+RELEASE_FEATURE = FeatureToggle("INCOMPLETE_FEATURE", CodeReadiness.RELEASE)
```
DelpireNI, line 14: The feature name here should be "RELEASE_FEATURE"

ni_measurementlink_service/_featuretoggles.py
bkeryan: https://docs.python.org/3/howto/enum.html#orderedenum

ni_measurementlink_service/_featuretoggles.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _raise_if_disabled(self) -> None:
+        if self.is_enabled:
+            return
+
+        env_vars = f"{_PREFIX}_ENABLE_{self.name}"
+        if self.readiness in [CodeReadiness.NEXT_RELEASE, CodeReadiness.INCOMPLETE]:
```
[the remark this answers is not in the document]
bkeryan: There is no "allow" variable for prototype.

DelpireNI:
We should update the docs on how to show how to use this when developing a feature behind a feature toggle, probably CONTRIBUTING.md. We should also update the README to show how to enable certain feature toggles. 

I imagine we'll also want a section on which features are available behind feature toggles.

ni_measurementlink_service/_featuretoggles.py
bkeryan: On second thought, I can't come up with a use case for enabling all prototype features, so no.

Also, `cast=CodeReadiness` doesn't work. It would have to be `cast=CodeReadiness.__getitem__` or `cast=lambda x: CodeReadiness[x]`.

bkeryan:
> We should update the docs on how to show how to use this when developing a feature behind a feature toggle, probably CONTRIBUTING.md. We should also update the README to show how to enable certain feature toggles.
> 
> I imagine we'll also want a section on which features are available behind feature toggles.

I added this content to the wiki: https://github.com/ni/measurementlink-python/wiki/Feature-Toggles

jasonmreding (approved):

tests/unit/test_featuretoggles.py
```diff
@@ -0,0 +1,100 @@
...
+    assert PROTOTYPE_FEATURE.is_enabled
+
+
+@pytest.mark.use_code_readiness(CodeReadiness.INCOMPLETE)
+def test___use_incomplete_readiness___is_enabled___reflects_code_readiness_level() -> None:
+    assert RELEASE_FEATURE.is_enabled
+    assert NEXT_RELEASE_FEATURE.is_enabled
+    assert INCOMPLETE_FEATURE.is_enabled
+    assert not PROTOTYPE_FEATURE.is_enabled
+
```
dixonjoel, line 56: Did you intentionally omit a next_release test?

dixonjoel (approved):

tests/unit/test_featuretoggles.py
bkeryan: These were originally spot checks but since they cover 3/4 cases I guess I'll add the 4th.

PR #425, opened 2023-09-29
service: Add _configuration submodule

bkeryan:

Add a `_configuration` submodule, which reads configuration options from the environment or a `.env` file. In later PRs, this will be used to replace the examples' command-line parameters and `ServiceOptions` class.

Update `_featuretoggles` to share the python-decouple `AutoConfig` with `_configuration`.

The new create_session(s) API will use this.

Ran mypy and new unit tests.

### DIFF
--- a/ni_measurementlink_service/_configuration.py
+++ b/ni_measurementlink_service/_configuration.py
@@ -1,0 +1,108 @@
+"""MeasurementLink configuration options."""
+from __future__ import annotations
+
+import pathlib
+import sys
+from typing import TYPE_CHECKING, Any, Callable, Dict, NamedTuple, TypeVar, Union
+
+from decouple import AutoConfig, Undefined, undefined
+
+if TYPE_CHECKING:
+    if sys.version_info >= (3, 11):
+        from typing import Self
+    else:
+        from typing_extensions import Self
+
+
+_PREFIX = "MEASUREMENTLINK"
+
+# Search for the `.env` file starting with the current directory.
+_config = AutoConfig(str(pathlib.Path.cwd()))
+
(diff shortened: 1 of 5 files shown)


bkeryan:
FYI @cfuentes-ni @rfriedma: This is my proposed replacement for the `USE_SIMULATION` variable as well as the hardcoded boardtype/model in `_nifoo_helpers.py`.

ni_measurementlink_service/_configuration.py
```diff
@@ -0,0 +1,108 @@
...
+if TYPE_CHECKING:
+    # Work around decouple's lack of type hints.
+    _T = TypeVar("_T")
+
+    def _config(
+        option: str,
+        default: Union[_T, Undefined] = undefined,
+        cast: Union[Callable[[str], _T], Undefined] = undefined,
+    ) -> _T:
+        ...
```
dixonjoel, line 31: Is this some kind of notation for 'pass'?

dixonjoel (approved):

ni_measurementlink_service/_configuration.py
mshafer-NI: It's effectively equivalent (in this context):
https://docs.python.org/3/library/constants.html#Ellipsis

Community hasn't decided on a convention for when to use which for empty contexts.
https://stackoverflow.com/a/55275342

cfuentes-ni (approved):

ni_measurementlink_service/_configuration.py
bkeryan: This is a type stub and type stubs use ellipses.

PR #426, opened 2023-09-30
service: Add generic create_session(s) API

bkeryan:

Add the generic create_session(s) API and some of the infrastructure needed for the driver-specific APIs.

session_management:
- Add `TSession` and `TSession_co` type variables
- Add `session: object = None` to `SessionInformation` to allow returning a session object along with the session info
- Add a `TypedSessionInformation[TSession_co]` protocol to allow returning session objects in a type-safe manner
  - Python versions <3.11 don't support generic named tuples, so I think a protocol is the best we can do.
  - The protocol doesn't exactly match the `SessionInformation` class because `TSession_co` is narrower than `object`, so a cast is needed.
- Add the generic `create_session` and `create_sessions` methods.
  - The core implementation is split out to hide the use of `@contextlib.contextmanager` from documentation and hover help
 
_drivers:
- New subpackage for driver-related code
- Provides a `closing_session` function to use to create a context manager to close the session

_drivers._grpcdevice:
- Provides a `get_insecure_grpc_device_channel` function

Part of the new session management API.

Ran mypy and tests.


ni_measurementlink_service/session_management.py
```diff
@@ -224,6 +294,115 @@ def unreserve(self) -> None:
...
+        try:
+            yield
+        finally:
+            del self._session_cache[session_name]
+
+    @contextlib.contextmanager
+    def _create_session_core(
+        self,
+        session_constructor: Callable[[SessionInformation], TSession],
+        instrument_type_id: str,
```
dixonjoel, line 311: I had the impression we were going to make `instrument_type_id` optional or pass empty string to get 'all' sessions, or in this case, 'the only' session. Am I misremembering? Similar for `_create_sessions_core`.

tests/unit/test_reservation.py
```diff
@@ -0,0 +1,224 @@
...
+
+def test___single_session_info___create_session___session_created_and_closed(
+    session_management_client: Mock,
+) -> None:
+    reservation = MultiSessionReservation(
+        session_management_client, _create_grpc_session_infos(1, "nifake")
+    )
+
+    with reservation.create_session(_construct_session, "nifake") as session_info:
+        assert isinstance(session_info.session, fake_driver.Session)
```
dixonjoel, line 38: The setup seems to be identical for the test before but the asserts are slightly different. This one has an assert for 'isinstance'. The previous has an assert on the 'session_name'. Is there a reason for them to be different? Maybe include all the assertions on both tests?

tests/unit/test_reservation.py
```diff
@@ -0,0 +1,224 @@
...
+        assert session_info.session.resource_name == "Dev0"
+        assert not session_info.session.is_closed
+
+    assert session_info.session.is_closed
+
+
+def test___single_session_info___create_session___session_cached(
```
dixonjoel, line 45: Since the final assert is that the _session_cache is empty, consider expanding the when clause like `session_cached_for_lifetime_of_create_session` (too wordy but hopefully you get the idea). Unless I'm misunderstanding the cache or test.

tests/unit/test_reservation.py
dixonjoel, line 1: Do we need any tests for a scenario where the user calls, e.g. `create_session` or `create_sessions` and then calls the opposite one?

dixonjoel (approved):
Approved with a few queries and relatively minor points. Since I'm out early this week, I don't want to hold anything up so feel free to bypass me. Maybe @pbirkhol-ni should take a quick look before submitting.

ni_measurementlink_service/_drivers/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+    Type,
+    TypeVar,
+    runtime_checkable,
+)
+
+import grpc
+
+
+class SessionInitializationBehavior(Protocol):
+    """IntEnum specifying whether to initialize a new session or attach to an existing one.
```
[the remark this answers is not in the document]
bkeryan: This is a protocol modeling an IntEnum.

`name: str` and `value: int` are instance properties for enums.

tests/utilities/fake_driver.py
```diff
@@ -0,0 +1,147 @@
...
+
+class GrpcSessionOptions:
+    """gRPC session options."""
+
+    def __init__(
+        self,
+        grpc_channel: grpc.Channel,
+        session_name: str,
+        *,
+        api_key: str = _API_KEY,
```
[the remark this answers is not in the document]
bkeryan: This field exists in the real GrpcSessionOptions classes. 
https://github.com/ni/nimi-python/blob/master/generated/nidcpower/nidcpower/grpc_session_options.py

tests/unit/test_reservation.py
bkeryan: "Tests should verify one thing."

ni_measurementlink_service/session_management.py
bkeryan: @pbirkhol-ni Are you supporting instrument_type_id="" in the LabVIEW Create Session(s) API?

ni_measurementlink_service/_drivers/__init__.py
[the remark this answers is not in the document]
bkeryan: I added some type-checking tests and this protocol doesn't work as well as I thought, so I'm deleting it and changing the TypeVar to have bound=int.

tests/unit/test_reservation.py
bkeryan: Renamed to `_session_lifetime_tracked` and moved the `is_closed` check into this test case.

ni_measurementlink_service/session_management.py
bkeryan: BTW, this parameter only exists in the generic functions. The driver-specific functions know what instrument type id to use.

pbirkhol-ni: `instrument_type_id` is exposed via the `IInstrument` interface, where there is a VI that you must implement to specify the instrument type id. Right now, we don't do any checking of the value and so a blank string is allowed. However, if you specify a blank string, we will not create any sessions and will not throw an error (at least in the version I am looking at now). For Get Connections, I am returning all connections if the instrument type id is blank. However, now that I think of it, maybe that isn't the right thing to do. Maybe we should error.

ni_measurementlink_service/session_management.py
```diff
@@ -224,6 +301,120 @@ def unreserve(self) -> None:
...
+        Args:
+            session_constructor: A function that constructs sessions based on session
+                information.
+            instrument_type_id: Instrument type ID for the session.
+                For NI instruments, use instrument type id constants, such as
+                :py:const:`INSTRUMENT_TYPE_NI_DCPOWER` or :py:const:`INSTRUMENT_TYPE_NI_DMM`.
+                For custom instruments, use the instrument type id defined in the pin map file.
+
+        Returns:
+            A context manager that yields a tuple of session information and session objects.
```
bkeryan, line 390: This is out of date

ni_measurementlink_service/_drivers/_grpcdevice.py
```diff
@@ -0,0 +1,41 @@
...
+from ni_measurementlink_service._internal.discovery_client import DiscoveryClient
+
+SERVICE_CLASS = "ni.measurementlink.v1.grpcdeviceserver"
+"""The service class for NI gRPC Device Server."""
+
+
+def get_insecure_grpc_device_channel(
+    discovery_client: DiscoveryClient,
+    grpc_channel_pool: GrpcChannelPool,
+    driver_module: DriverModule,
```
bkeryan, line 18: I don't think I'll need this DriverModule abstraction after all.

ni_measurementlink_service/_drivers/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+    #
+    # Python 3.11 doesn't have a good way to describe read-only class variables
+    # or protocols that inherit from built-in types such as int.
+
+    @property
+    def AUTO(  # noqa: N802 - function name should be lowercase
+        self,
+    ) -> TSessionInitializationBehavior_co:
+        """Automatically initialize a new session or attach to an existing one."""
+        ...
```
pbirkhol-ni, line 40: Is this the same as using `pass`? I've never seen this syntax before and it is hard to come up with a google search to figure out what it means.

pbirkhol-ni (approved):
Looks good to me. I didn't see any functional discrepancies between LabVIEW and python, although the implementation itself is radically different.  :)

ni_measurementlink_service/_drivers/__init__.py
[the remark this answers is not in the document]
bkeryan: I ended up using a different approach in the next PR: define a driver-independent SessionInitializationBehavior and keep the GRPC_DEVICE_INTERFACE reference in driver-specific code, so these protocols have been deleted.

ni_measurementlink_service/_drivers/_grpcdevice.py
bkeryan: Removed

ni_measurementlink_service/session_management.py
```diff
@@ -143,6 +162,63 @@ class SessionInformation(NamedTuple):
...
+
+
+# Python versions <3.11 do not support generic named tuples, so we use a generic
+# protocol to return typed session information.
+class TypedSessionInformation(Protocol, Generic[TSession_co]):
+    """Generic version of :any:`SessionInformation` that preserves the session type."""
+
+    @property
+    def session_name(self) -> str:
+        """See :any:`SessionInformation.session_name`."""
```
bkeryan, line 189: This does not make for a great user experience in VS Code

bkeryan: I duplicated the 1-line summary

PR #436, opened 2023-10-06
service: Fix session type checks

bkeryan:

`_with_typed_session()` was incorrectly assuming that the runtime type matches the static type. Also, it was trying to infer the static type variable `TSession` based on the runtime type expression `type(session)`.

Fixing this revealed a problem with the tests: patching `driver.Session` and replacing it with a mock causes `isinstance` checks to fail. The 2nd argument to `isinstance` must be a type, so it raises an error if `driver.Session` is a `Mock` instance rather than a type. Patching `driver.Session.__new__` instead of `driver.Session` avoids this problem because it ensures that `driver.Session` is still a type. (Patching `driver.Session` to replace it with a subclass that overrides `__new__` would also work.)

These type checks are needed for `get_connection(s)`

Ran updated unit tests.

### DIFF
--- a/ni_measurementlink_service/session_management/_reservation.py
+++ b/ni_measurementlink_service/session_management/_reservation.py
@@ -136,7 +136,6 @@ class BaseReservation(abc.ABC):
         self,
         session_constructor: Callable[[SessionInformation], TSession],
         instrument_type_id: str,
-        session_type: Optional[Type[TSession]] = None,
     ) -> Generator[TypedSessionInformation[TSession], None, None]:
         if not instrument_type_id:
             raise ValueError("This method requires an instrument type ID.")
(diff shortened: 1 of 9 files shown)


ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -153,13 +155,17 @@ def _create_session_core(
         session_info = session_infos[0]
         with closing_session(session_constructor(session_info)) as session:
             with self._cache_session(session_info.session_name, session):
-                yield session_info._with_typed_session(session)
+                new_session_info = session_info._with_session(session)
+                if session_type:
+                    new_session_info._check_runtime_type(session_type)
```
bkeryan, line 160: On second thought, create_session(s) doesn't really need a runtime type check. Only get_connection(s) does.

dixonjoel (approved):

PR #437, opened 2023-10-09
service: Stop leaking internal types from the MeasurementService class

### LINKED ISSUE #334: `MeasurementService` class attributes leak internal types
# Bug Report

The `MeasurementService` class has public attributes with private/internal types.

## Repro or Code Sample

https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/measurement/service.py#L125
```
    Attributes
    ----------
        measurement_info (info.MeasurementInfo): Measurement info

        service_info(info.ServiceInfo) : Service Info

        configuration_parameter_list (List): List of configuration parameters.

        output_parameter_list (list): List of output parameters.

        measure_function (Callable): Registered measurement function.

        context (MeasurementContext): Accessor for context-local state.

        discovery_client (DiscoveryClient): Client for accessing the MeasurementLink discovery
            service.

        channel_pool (GrpcChannelPool): Pool of gRPC channels used by the service.

    """
```

## Expected Behavior

All public attributes have public types.

## Current Behavior

`configuration_parameter_list` and `output_parameter_list` have an underspecified type signature. They are really lists of `ni_measurementlink_service._internal.parameter.metadata.ParameterMetadata`, but that is an internal type.

`discovery_client` has an internal type.

`measure_function` is arguably an internal attribute, even though it doesn't leak any internal types.

## Your Environment

* OS & Device: Windows
* `ni-measurementlink-service` version: 1.0 through 1.2.0-dev0
* MeasurementLink version: N/A
* Python version: N/A

[AB#2457615](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2457615)

bkeryan:

Make discovery API public:
- Move `ni_measurementlink_service/_internal/discovery_client.py` to `ni_measurementlink_service/discovery/` and split into sub-modules.
- Deprecate `ni_measurementlink_service._internal.discovery_client`.
- Update tests, generator, and examples to use the new sub-package.

Deprecate more `MeasurementService` attributes that should not be public:
- Replace `configuration_parameters`, `output_parameters`, and `measure_function` with deprecated properties.
- Change internal `_configuration_parameters` and `_output_parameters` to have correct type hints.
- Change internal `_measure_function` initialization checks to use a sentinel function that raises an exception when called. Before, we checked for `None`, but the attribute was not assigned in the constructor, so accessing it would have raised `AttributeError`

Fixes #334 

Ran mypy and tests.


ni_measurementlink_service/discovery/_client.py
bkeryan, line 1: @dixonjoel @pbirkhol-ni To review this split, look at the 1st two commits. The 1st commit copies the old file to the new locations, and the 2nd commit removes the duplicated parts.

ni_measurementlink_service/discovery/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Client for accessing the MeasurementLink discovery service."""
```
dixonjoel, line 1: Seems like this should be similar to
"""Redirect internal discovery client API to the public API."""

bkeryan: This **is** the public API, so I updated the docstring to """Public API for accessing the MeasurementLink discovery service."""

PR #438, opened 2023-10-10
service: Add generic get_connection(s) API

bkeryan:

Add `reservation.get_connection` and `reservation.get_connections` for mapping a pin and site to a specific instrument session and channel.

Update the reservation classes to eagerly convert the gRPC session infos to API session infos.

Part of the new session management API.

Ran new unit tests.


ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -183,6 +275,61 @@ def _create_sessions_core(
...
+        sites: Union[int, Iterable[int], None] = None,
+        instrument_type_id: Optional[str] = None,
+    ) -> Sequence[TypedConnection[TSession]]:
+        pin_order = _to_iterable(pin_or_relay_names, self._reserved_pin_or_relay_names)
+
+        if sites == SITE_ALL_SITES:
+            site_order = self._reserved_sites
+        else:
+            site_order = _to_iterable(sites, self._reserved_sites)
+            if SITE_SYSTEM_PINS in self._reserved_sites and SITE_SYSTEM_PINS not in site_order:
```
bkeryan, line 308: @pbirkhol-ni @jasonmreding @jonathanmendez Does this approach to handling system pins make sense to you? If any of the reserved pins are system pins and the specified list of sites does not include -1, add -1 to the end.

tests/unit/test_reservation.py
```diff
@@ -208,5 +215,402 @@ def test___heterogenous_session_infos___create_sessions___grouped_by_instrument_
...
+    grpc_session_infos[1].channel_mappings.add(pin_or_relay_name="Pin1", site=1, channel="3")
+    grpc_session_infos[2].channel_mappings.add(pin_or_relay_name="Pin2", site=1, channel="4")
+    grpc_session_infos[2].channel_mappings.add(pin_or_relay_name="Pin3", site=1, channel="5")
+    grpc_session_infos[3].channel_mappings.add(pin_or_relay_name="Pin4", site=0, channel="6")
+    grpc_session_infos[3].channel_mappings.add(pin_or_relay_name="Pin4", site=1, channel="7")
+    grpc_session_infos[3].instrument_type_id = "nibar"
+    return grpc_session_infos
+
+
+def _create_grpc_session_infos_with_system_pins() -> (
```
bkeryan, line 606: @pbirkhol-ni @jasonmreding @jonathanmendez What test cases am I missing? Shared pins? Do we need to update the session management service for that?

tests/unit/test_reservation.py
```diff
@@ -208,5 +215,402 @@ def test___heterogenous_session_infos___create_sessions___grouped_by_instrument_
...
+    grpc_session_infos = _create_grpc_session_infos_for_ordering()
+    reservation = MultiSessionReservation(
+        session_management_client,
+        grpc_session_infos,
+        reserved_pin_or_relay_names=["Pin3", "Pin1", "Pin4", "Pin2"],
+        reserved_sites=[1, 0],
+        reserved_instrument_type_ids=["nifoo", "nibar"],
+    )
+
+    nifoo_connections = reservation.get_connections(
```
bkeryan, line 414: @pbirkhol-ni It's currently not an error to get connections for instrument_type_id=nifoo with pins that only match instrument_type_id=nibar, as long as some of the pins match nifoo. Please let me know if this is inconsistent with LV.

It now occurs to me that this also tolerates entirely made-up pin names, as long as some of the pins are real and match the session info. Perhaps we should validate that the pins and sites exist in the reservation data.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -183,6 +275,61 @@ def _create_sessions_core(
...
+                site_order = site_order + [SITE_SYSTEM_PINS]
```
bkeryan, line 309:
```
ni_measurementlink_service/session_management/_reservation.py:30[9](https://github.com/ni/measurementlink-python/actions/runs/6463055419/job/17545646010?pr=438#step:9:10): error: Unsupported left operand type for + ("Iterable[int]")  [operator]
```

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -76,12 +106,30 @@ def __init__(
         self,
         session_manager: SessionManagementClient,
         session_info: Sequence[session_management_service_pb2.SessionInformation],
+        reserved_pin_or_relay_names: Union[str, Iterable[str], None] = None,
```
bkeryan, line 109: I also need to update the SessionManagementClient to pass these parameters.

ni_measurementlink_service/session_management/_reservation.py
pbirkhol-ni: LabVIEW does something similar. However, I just always add -1 to the end of the sites list (unless -1 is already included). I suppose excluding -1 if there are no reserved system pins is reasonable, but it requires tracking extra state that I don't currently track.

tests/unit/test_reservation.py
pbirkhol-ni: Yes, this is inconsistent. In LabVIEW, if you get connections for a pin that does not have any connections, an error is thrown.

tests/unit/test_reservation.py
pbirkhol-ni: The LabVIEW implementation does not handle shared pins well. There is a follow up user story for how to handle shared pins with the following comment:

`
There isn't a great way to handle shared pins. If you have a shared pin across multiple sites, the channel mapping will only contain the shared pin on a single site. That means that if you call Get Connections on a single site, you may or may not get the shared pin. The workaround would be to always query all sites when trying to get a shared pin.
`

I think the solution may require an update to the session manager service since nothing in the session information currently indicates that a pin is shared.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -94,6 +142,52 @@ def _grpc_channel_pool(self) -> GrpcChannelPool:
             raise ValueError("This method requires a gRPC channel pool.")
         return self._session_manager._grpc_channel_pool
 
+    @cached_property
```
pbirkhol-ni, line 145: I'm really jealous of this decorator. It allows you to write the code in an intuitive way and have python make it efficient for you. LabVIEW, on the other hand......

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -183,6 +275,61 @@ def _create_sessions_core(
...
+            site_order = _to_iterable(sites, self._reserved_sites)
+            if SITE_SYSTEM_PINS in self._reserved_sites and SITE_SYSTEM_PINS not in site_order:
+                site_order = site_order + [SITE_SYSTEM_PINS]
+
+        instrument_type_id_order = _to_iterable(
+            instrument_type_id, self._reserved_instrument_type_ids
+        )
+
+        results: List[TypedConnection[TSession]] = []
+        for instrument_type in instrument_type_id_order:
```
bkeryan, line 316: Instrument type should not be the top-level sort because that overrides the specified site/pin order in unexpected ways.

@pbirkhol-ni I realize the LV version requires the instrument type id, so it doesn't even have this situation.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -94,6 +142,52 @@ def _grpc_channel_pool(self) -> GrpcChannelPool:
...
+                sites[channel_mapping.site] = True
+        return list(sites.keys())
+
+    @cached_property
+    def _reserved_instrument_type_ids(self) -> Iterable[str]:
+        # Use dict keys to preserve insertion order.
+        ids = {}
+        for session_info in self._session_info:
+            ids[session_info.instrument_type_id] = True
+        return list(ids.keys())
```
mshafer-NI, line 169: So..., dictionary insertion is O(1),..., but so is list appending...

( note: that `dict.keys()` is O(n), so current implementation is O(n<sup>2</sup>)!! )

Which is simpler?
```suggestion
    @cached_property
    def _reserved_instrument_type_ids(self) -> Iterable[str]:
        # Use dict keys to preserve insertion order.
        ids = list()
        for session_info in self._session_info:
            ids.append(session_info.instrument_type_id)
        return ids
```

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -94,6 +142,52 @@ def _grpc_channel_pool(self) -> GrpcChannelPool:
...
         return self._session_manager._grpc_channel_pool
 
+    @cached_property
+    def _reserved_pin_or_relay_names(self) -> Iterable[str]:
+        # Use dict keys to preserve insertion order.
+        names = {}
+        for session_info in self._session_info:
+            for channel_mapping in session_info.channel_mappings:
+                names[channel_mapping.pin_or_relay_name] = True
+        return list(names.keys())
```
mshafer-NI, line 152:
```suggestion
    def _reserved_pin_or_relay_names(self) -> Iterable[str]:
        names = []
        for session_info in self._session_info:
            for channel_mapping in session_info.channel_mappings:
                names.append(channel_mapping.pin_or_relay_name)
        return names
```

mshafer-NI: (see other comment for why)

mshafer-NI (changes requested):

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -94,6 +142,52 @@ def _grpc_channel_pool(self) -> GrpcChannelPool:
...
+
+    @cached_property
+    def _reserved_sites(self) -> Iterable[int]:
```
pbirkhol-ni, line 155: In LabVIEW, the pin map context has an array of sites and I use that as the order. Similarly, for the pin order, I just cache off the array of pins passed when reserving the session. I don't have enough intuition on the order of the channel mappings, but I suspect you won't necessarily get the same ordering with this logic as we do in LabVIEW. It also seems like it would be difficult for a user to intuit the order based on this logic.

Did you do it this way because it allows you to also order by instrument type id in a consistent way? If so, I might prefer that instrument type id ordering should either be non guaranteed or simply be alphabetical order. We don't run into this problem in LabVIEW (yet) because we always pass an instrument type id. When we implement the generic Get Connections, I haven't thought about whether the instrument type id should be required or not. If not, then we'll have to tackle ordering in LabVIEW too.

tests/unit/test_reservation.py
```diff
@@ -208,5 +215,402 @@ def test___heterogenous_session_infos___create_sessions___grouped_by_instrument_
...
+            _ = reservation.get_connection(int)
+
+        assert "Incorrect type for session 'MySession0'." in exc_info.value.args[0]
+        assert (
+            "Expected <class 'int'>, got <class 'tests.utilities.fake_driver.Session'>."
+            in exc_info.value.args[0]
+        )
+
+
+def test___session_not_created___get_connection___type_error_raised(
```
pbirkhol-ni, line 370: This got me thinking about LabVIEW. I thought something like this would work (it would just return an invalid session), but it doesn't. Casting an empty variant to an IVI session errors. So I'll have to file a bug for that.

Then the question becomes, should it be valid to get connections, even if you haven't created the session with our API? The output of get connections has more information than just the session. Should we allow you to get at that information, even if the session is invalid?

pbirkhol-ni: Nevermind. It looks like you allow an empty session. So python does it right. I just have to fix LabVIEW to do something similar.

ni_measurementlink_service/session_management/_reservation.py
pbirkhol-ni: I commented on this below, but yes, currently we don't have a generic Get Connections for LabVIEW. When we do, I haven't decided whether instrument type id will be required or optional.

ni_measurementlink_service/session_management/_reservation.py
pbirkhol-ni: After looking at the tests, changing this will have some pretty serious ramifications on the tests. But I suspect refactoring them in python will be a lot easier than in LabVIEW.  :D

tests/unit/test_reservation.py
```diff
@@ -208,5 +215,402 @@ def test___heterogenous_session_infos___create_sessions___grouped_by_instrument_
...
+        "Pin3",
+        "Pin2",
+        "Pin1",
+        "Pin3",
+        "Pin2",
+    ]
+    assert [conn.site for conn in nifoo_connections] == [0, 0, 0, 1, 1, 1]
+    assert [conn.channel_name for conn in nifoo_connections] == ["0", "2", "1", "3", "5", "4"]
+    assert [conn.pin_or_relay_name for conn in nibar_connections] == ["Pin4", "Pin4"]
+    assert [conn.site for conn in nibar_connections] == [0, 1]
```
bkeryan, line 439: Building a list of named tuples would make these more readable, compared to the parallel lists.

ni_measurementlink_service/session_management/_reservation.py
bkeryan: > I suppose excluding -1 if there are no reserved system pins is reasonable, but it requires tracking extra state that I don't currently track.

The pin map context's site list doesn't include site -1, so this didn't work correctly anyway.

tests/unit/test_reservation.py
bkeryan: Moved to feature AB#2546523: Fully support querying connections for shared pins

BTW, I added an integration test for shared pins.

ni_measurementlink_service/session_management/_reservation.py
bkeryan: Appending to a list doesn't remove duplicates.

bkeryan: > ( note: that dict.keys() is O(n), so current implementation is O(n2)!! )

This is a cached property, so it only executes once.

ni_measurementlink_service/session_management/_reservation.py
bkeryan: This is the fallback order for reserve_all_registered_sessions.

pbirkhol-ni: I must be missing something. The way I read the code below, if I don't pass sites to get_connections(), the site order will get populated with the result of _reserved_sites(). Same with pins.

ni_measurementlink_service/session_management/_reservation.py
mshafer-NI: > Appending to a list doesn't remove duplicates.

If this is a reason to use a dict, then the comment should say this.

mshafer-NI: Also, if you REALLY REALLY care about order -> which should be taken, the first occurrence of a name or the last? (what you currently have will be the first):


```
>python
Python 3.9.6 (tags/v3.9.6:db3ff76, Jun 28 2021, 15:26:21) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> data = {}
>>> for x in "apple_apple":
...     data[x] = True
...
>>> data
{'a': True, 'p': True, 'l': True, 'e': True, '_': True}
```

ni_measurementlink_service/session_management/_reservation.py
bkeryan: This code now adds -1 to the end, if not already present.

tests/unit/test_reservation.py
bkeryan: I updated the code to error if the caller specifies pins, sites, or instrument type id and they don't match anything.

ni_measurementlink_service/session_management/_reservation.py
bkeryan: Instrument type id is now the last-level, "tiebreaker" sort.

ni_measurementlink_service/session_management/_reservation.py
bkeryan: _reserved_sites is used if you don't specify sites when calling get_connection(s). However, it doesn't always come from this `@cached_property` function.
- reserve_sessions initializes _reserved_sites based on the pin map context, so it doesn't use the cached property.
- reserve_all_registered_sessions does not initialize _reserved_sites, so it uses the cached property.

ni_measurementlink_service/session_management/_reservation.py
jasonmreding: FYI, the pin map service will return an invalid argument error if you try and pass a pin site of -1. It will return it as a site number in the output returned from the API but not as an input to the API.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -239,10 +425,83 @@ def create_sessions(
...
+                sessions matched the instrument type ID.
         """
         return self._create_sessions_core(session_constructor, instrument_type_id)
 
+    @requires_feature(SESSION_MANAGEMENT_2024Q1)
+    def get_connection(
+        self,
+        session_type: Type[TSession],
+        pin_or_relay_names: Union[str, Iterable[str], None] = None,
+        sites: Union[int, Iterable[int], None] = None,
```
bkeryan, line 437: The single-connection version might as well take a single pin/relay and a single site, since there can only be one result.

jasonmreding: That is consistent with the LV API.

ni_measurementlink_service/session_management/_reservation.py
pbirkhol-ni: Thanks. I got it now.

pbirkhol-ni: I think this was just my confusion about how `@cached_property` works. Upon looking at the docs, what you are doing now makes sense.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -183,6 +285,90 @@ def _create_sessions_core(
...
+
+        if pin_or_relay_names is not None:
+            missing_pins = set(requested_pins) - matching_pins
+            if missing_pins:
+                missing_pin_list = ", ".join(f"'{pin}'" for pin in sorted(missing_pins))
+                raise ValueError(
+                    f"No reserved connections matched pin or relay name(s) {missing_pin_list}."
+                )
+
+        if sites is not None:
```
pbirkhol-ni, line 359: We don't do this check in LabVIEW. My main concern is in that in the all sites case, do you really want an error if you don't get a connection for at least every site? Maybe the answer is yes. I'm not sure. It seemed to heavy handed to me.

pbirkhol-ni (approved):

ni_measurementlink_service/session_management/_reservation.py
bkeryan: In general, all sites should have the same pins. If you create a site that doesn't have connections for all of the DUT pins, the pin map editor displays a warning:

> Warning: An error might occur at run-time.
> The following pins in the pin map are not connected:
>
> DUTPin1 - Sites: 1

I could change the code to check for missing sites only when there are no results, but this will allow passing a non-existent site like `sites=[0, 1, 3873875837]`, as long as one of the other sites matches a connection.

I could change the code to check whether the requested sites are a subset of the reserved sites up-front. Perhaps this is better.

Note that this code only knows which pins and sites are reserved, not which pins and sites exist in the pin map. This means that the `SITE_ALL_SITES` constant actually matches against all reserved sites, not all sites. Perhaps I should rename it?

Cc: @jonathanmendez @jasonmreding

ni_measurementlink_service/session_management/_reservation.py
bkeryan: I admit that it's subtle and not obvious if you're not familiar with `@cached_property`. That's why I wrote the comments about "If __init__ doesn't initialize ..., the cached properties lazily initialize them" and "If ...  is not specified at construction time ...".

ni_measurementlink_service/session_management/_reservation.py
jasonmreding: I don't know that I have a strong opinion either way. For the most part, I think it won't really matter or have much impact on our users. There are perhaps some corner cases involving shared pins or system pins where this could cause an issue, but I suspect in most cases even that won't be an issue. 

I tend to view sites and instrument type ID inputs as equivalent in regards to this function so I would skew to keeping their behavior the same. So if one errors then I would also error for the other. If one doesn't throw an error, then I wouldn't throw an error for the other one either. In most cases, I suspect only throwing an error if there are no matching pins is sufficient and would cover no matching sites or instrument type ID as well. Is there any case where returning an empty list of connections without throwing an exception would be acceptable/preferable? Is the intent to also use this API to get at the "raw" session info without performing any filtering? If so, would it make sense to throw an error whenever there are no matching pins when any of the pins, instrument type ID, or sites were explicitly set? Or was this the intent all along, and we are just trying to provide a more targeted, helpful error message?

bkeryan: I want to avoid garbage-in, garbage out. If the user misspells a pin name or instrument type ID, I would prefer to return an error rather than silently ignore it. So I am adding pre-query checks to verify that the requested pins, sites, and instrument type ID match *something* in isolation (even if they do not match anything when combined).

This leaves two cases for post-query checks:
- You specified pin names, but we didn't find anything that matched them. If you got past the pre-query check, the pin names are valid, so this condition must mean that the combination of pin name and other criteria (site and instrument type ID) didn't match anything. So I think this could throw a slightly more descriptive error.
- For whatever reason, nothing matched. Perhaps you used reserve_sessions(), nothing was reserved, and you didn't specify any matching criteria. I guess it's okay for get_connections() to return an empty list here. get_connection() will still error.

> In most cases, I suspect only throwing an error if there are no matching pins is sufficient and would cover no matching sites or instrument type ID as well.

With the added pre-query checks, I agree. Any post-query check failure means that the combination of criteria failed to match.

> Is there any case where returning an empty list of connections without throwing an exception would be acceptable/preferable? 

Maybe the case where reserve_sessions() reserved nothing.

> Is the intent to also use this API to get at the "raw" session info without performing any filtering?

Yes, it can return all reserved connections.

> If so, would it make sense to throw an error whenever there are no matching pins when any of the pins, instrument type ID, or sites were explicitly set?

That seems plausible.

bkeryan: I updated the error handling as described in my last reply.

ni_measurementlink_service/session_management/_reservation.py
bkeryan: Update: I am now using `dict_keys` as an ordered set. It supports O(n) iteration and O(1) `x in set` (amortized best case).

```python
# The `dict_view` type is a set-like type. In Python 3.7 and later, dictionaries
# preserve insertion order.
#
# Note: the set difference operator does not preserve order.
def _to_ordered_set(values: Iterable[_T]) -> AbstractSet[_T]:
    return dict.fromkeys(values).keys()
```

bkeryan: > note: that dict.keys() is O(n)

BTW, in current versions of Python, `dict.keys()` returns a view object, which is O(1). Iterating over the keys is O(n).

mshafer-NI: > BTW, in current versions of Python, dict.keys() creates a view object, which is O(1). Iterating over the keys is O(n).

This is really what I meant. Thank you for phrasing better.

mshafer-NI (approved):
Thanks for addressing my comments

ni_measurementlink_service/session_management/_reservation.py
pbirkhol-ni: What you have now seems reasonable. I will create a work item so that LabVIEW can match this behavior.

dixonjoel (approved):

PR #443, opened 2023-10-13
examples: Run ni-python-styleguide separately for each example

bkeryan:

Add a dependency on `ni-python-styleguide` to each example. Running `ni-python-styleguide` from the main project treats the examples as part of the main project, which affects sorting of imports.

Update the main project's `ni-python-styleguide` and `black` configurations to exclude the examples.

Update `sample_measurement`'s `ni-python-styleguide` configuration to exclude Protobuf codegen.

Update the `check_examples` workflow to lint the examples separately.

Fixes https://github.com/ni/measurementlink-python/issues/374

PR build

### DIFF
--- a/.github/workflows/check_examples.yml
+++ b/.github/workflows/check_examples.yml
@@ -38,6 +38,15 @@ jobs:
             popd
             echo "::endgroup::"
           done
+      - name: Lint examples
+        run: |
+          for example in examples/*/; do
+            echo "::group::$example"
+            pushd $example
+            poetry run ni-python-styleguide lint
+            popd
+            echo "::endgroup::"
+          done
       - name: Mypy static analysis (examples, Linux)
         run: |
           for example in examples/*/; do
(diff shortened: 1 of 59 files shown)


dixonjoel:
To check my understanding, locally, when developing now, if we want to check any examples we need to change directory to the example and then run 'poetry run ni-python-styleguide lint (or fix)' there? And 'poetry run ni-python-styleguide lint' from the root will no longer lint the examples?

dixonjoel (approved):

bkeryan:
> To check my understanding, locally, when developing now, if we want to check any examples we need to change directory to the example and then run 'poetry run ni-python-styleguide lint (or fix)' there? And 'poetry run ni-python-styleguide lint' from the root will no longer lint the examples?

Yes, that's right.

I would still like to use PoeThePoet or some Poetry workspace plugin to automate running tasks on all of the examples, but this isn't required to fix the bug.

PR #448, opened 2023-10-13
[RFC] examples: Update one of the TestStand code module to use the new session management API

bkeryan:

Update one of the TestStand code modules to use the new session management API.

Uses same session creation code as for measurements, with simulation via `.env` file.

Allows deleting more example-specific "helpers" code.

Manually tested with TestStand 2021 SP1.

### DIFF
--- a/examples/nidcpower_source_dc_voltage/teststand_fixture.py
+++ b/examples/nidcpower_source_dc_voltage/teststand_fixture.py
@@ -2,9 +2,14 @@
 from typing import Any
 
 import ni_measurementlink_service as nims
-import nidcpower
 from _helpers import GrpcChannelPoolHelper, PinMapClient, TestStandSupport
-from _nidcpower_helpers import create_session
+from ni_measurementlink_service.discovery import DiscoveryClient
+from ni_measurementlink_service.session_management import (
+    INSTRUMENT_TYPE_NI_DCPOWER,
+    PinMapContext,
+    SessionInitializationBehavior,
+    SessionManagementClient,
+)
 
 
 def update_pin_map(pin_map_path: str, sequence_context: Any) -> str:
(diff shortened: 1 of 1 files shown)


examples/nidcpower_source_dc_voltage/teststand_fixture.py
```diff
@@ -39,48 +44,47 @@ def create_nidcpower_sessions(sequence_context: Any) -> None:
...
-                    session_info,
-                    grpc_device_channel,
-                    initialization_behavior=nidcpower.SessionInitializationBehavior.INITIALIZE_SERVER_SESSION,
-                )
+            # Initialize new sessions on the server, but do not close them.
+            # (Deliberately leak them.)
+            context_manager = reservation.create_nidcpower_sessions(
+                initialization_behavior=SessionInitializationBehavior.INITIALIZE_SERVER_SESSION
+            )
+            _ = context_manager.__enter__()
```
bkeryan, line 63: Do we want something less subtle/advanced here?

Ideas:
1. Add an `INITIALIZE_AND_DETACH` enum value to measurementlink-python and move this logic into measurementlink-python, like we are doing for VISA
2. Add an `INITIALIZE_AND_DETACH` enum value to nimi-python and nidaqmx-python, like we are doing for VISA
3. Add a detach_session() method to the measurementlink-python `SessionInformation` object
4. Add a detach() method to nimi-python and nidaqmx-python

examples/nidcpower_source_dc_voltage/teststand_fixture.py
```diff
@@ -39,48 +44,47 @@ def create_nidcpower_sessions(sequence_context: Any) -> None:
...
-                    grpc_device_channel,
-                    initialization_behavior=nidcpower.SessionInitializationBehavior.ATTACH_TO_SERVER_SESSION,
-                )
-                session.close()
+            # Attach to the sessions on the server and explicitly close them.
+            with reservation.create_nidcpower_sessions(
+                initialization_behavior=SessionInitializationBehavior.ATTACH_TO_SERVER_SESSION
+            ) as session_infos:
+                for session_info in session_infos:
+                    session_info.session.close()
```
bkeryan, line 89: I was a little surprised here that the `with` block didn't automatically close the sessions.

grpc-device returned `new_session_initialized=False`, so nimi-python's `__exit__` method didn't call `close()`. I don't think it checks for `initialization_behavior == AUTO` in `__exit__`.

examples/nidcpower_source_dc_voltage/teststand_fixture.py
```diff
@@ -39,48 +44,47 @@ def create_nidcpower_sessions(sequence_context: Any) -> None:
...
+        session_management_client = SessionManagementClient(
+            discovery_client=discovery_client, grpc_channel_pool=grpc_channel_pool
         )
         with session_management_client.reserve_all_registered_sessions(
-            instrument_type_id=nims.session_management.INSTRUMENT_TYPE_NI_DCPOWER,
+            instrument_type_id=INSTRUMENT_TYPE_NI_DCPOWER
         ) as reservation:
+            # If there are no registered sessions, there is nothing to clean up.
+            if not reservation.session_info:
+                return
```
bkeryan, line 80: This is necessary because create_nidcpower_sessions() raises an exception if there are no sessions to create. For measurements, we assume users always want to create one or more sessions, but for code module cleanup it's possible to have zero sessions.

We could potentially tweak the error handling or add an optional parameter to allow creating an empty list.

examples/nidcpower_source_dc_voltage/teststand_fixture.py
bkeryan: 5. Add `reservation.detach_sessions()`

jasonmreding: I have been thinking about this from the LV perspective as well, and I'm leaning more towards having a single enum that specifies both init and close behavior at the same time. It would also have an "Auto" option if that makes sense, where the auto option would calculate initialize behavior at run time and apply symmetrical cleanup behavior.

I think this makes sense because:
1. It makes sense for languages where cleanup can be provided by exit from a scope or context. This also applies to LV when the top level VI goes idle for refnums that have a cleanup proc.
2. Lowest common denominator. The VISA driver API is planning to provide an enum like this. If we don't offer a similar API to our wrapper APIs, then how will it integrate with driver APIs like VISA?

jasonmreding: I'm not opposed to pushing more of this capability directly into the driver APIs. However, I think we can also use the MeasurementLink layer to provide a more consistent API/experience across drivers as well.

bkeryan: The Python session management API defines a driver-independent version of this enum: https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/session_management/_types.py#L289

and maps it to each driver-specific enum:
https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/_drivers/_nidcpower.py#L18

Proposal:
- Change the MeasurementLink version of the enum to have 5 options (AUTO, INIT_AND_CLOSE, INIT_AND_DETACH, ATTACH_AND_CLOSE, ATTACH_AND_DETACH)
- Map these 5 options to the 3 options supported by nimi-python and nidaqmx-python
- Implement the 2 "_AND_CLOSE" options by explicitly calling close

examples/nidcpower_source_dc_voltage/teststand_fixture.py
jasonmreding: This also avoids making a bunch of grpc calls that are ultimately noops (i.e. unregister_sessions) so this is probably better example code regardless.

examples/nidcpower_source_dc_voltage/teststand_fixture.py
bkeryan: > I'm not opposed to pushing more of this capability directly into the driver APIs. However, I think we can also use the MeasurementLink layer to provide a more consistent API/experience across drivers as well.

I agree.

Also, we can do it in the MeasurementLink layer now without a driver API release.

jasonmreding: That's more or less what I was thinking for LV as well. It would mean bubbling that input up to all the driver specific Create Session APIs. The parameter would default to AUTO so it wouldn't have to be explicitly set as part of the measurement logic. FYI, @pbirkhol-ni , I think you had some opinions about this. If you have objections or concerns to this approach, now is a good time to voice them.

jasonmreding (approved):

examples/nidcpower_source_dc_voltage/teststand_fixture.py
bkeryan: Ok, I added AB#2552420

dixonjoel: > refer to VISA for exact naming

Do you have a link for this?

Can you explain each of the 5 options in terms of what is returned (it returns a ContextManager right?) and what state that context manager is in?

For example, does INIT_AND_CLOSE mean that we will close the session(s) when the ContextManager we return to you is exited? And INIT_AND_DETACH means that the session(s) will not be closed at all by us?

pbirkhol-ni: > If you have objections or concerns to this approach, now is a good time to voice them.

I don't have any major concerns. The enum option is better than any alternatives I have heard. @jasonmreding, can you add a user story for LabVIEW. This change would be hard, but will involve touching a lot of VIs, updating tests, and writing new tests.

dixonjoel (approved):

examples/nidcpower_source_dc_voltage/teststand_fixture.py
bkeryan: > > refer to VISA for exact naming
> 
> Do you have a link for this?

https://dev.azure.com/ni/DevCentral/_git/ni-central?path=%2Fdocs%2Fspecs%2Fvisa%2Fnivisa_grpc_client_support_for_labview%2Fvisa_grpc_client_support_for_labview_api.md&version=GBmain&_a=preview

> Can you explain each of the 5 options in terms of what is returned (it returns a ContextManager right?) and what state that context manager is in?
> 
> For example, does INIT_AND_CLOSE mean that we will close the session(s) when the ContextManager we return to you is exited? And INIT_AND_DETACH means that the session(s) will not be closed at all by us?

The context manager that we return is implemented as a generator in [_create_session(s)_core](https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/session_management/_reservation.py#L285).

I think we can implement this by updating closing_session(). This currently chooses between using the session as a context manager or returning contextlib.closing() to call close().

The updated behavior will be something like this (assuming the session is both a context manager and closable, which we can make a requirement):

- AUTO: use the session's context manager
- INIT_AND_CLOSE: use contextlib.closing
- INIT_AND_DETACH: use contextlib.nullcontext
- ATTACH_AND_CLOSE: use contextlib.closing
- ATTACH_AND_DETACH: use contextlib.nullcontext

jasonmreding: > can you add a user story for LabVIEW

This is what AB#2552373 is intended to capture. The API changes will impact both code modules and measurements so you're welcome to move ownership of the user story if you think it is better tracked somewhere else.

vigkre:
@bkeryan Can we go ahead and close this PR? 

#480, #481, #482 already covers the implementation required

Issue #370, opened 2023-09-08
ServerCallLogger incorrectly logs OK when call is canceled


bkeryan:
# Bug Report

The server call logger logs OK when the call is canceled.

## Repro or Code Sample

Run `nifgen_standard_waveform` and click the **Stop** button.

## Expected Behavior

Log message says "CANCELLED".

## Current Behavior

Log message says "OK":
```
PS D:\dev\measurementlink-python\examples\nifgen_standard_function> poetry run python .\measurement.py -v
2023-09-08 11:04:51,201 INFO: Measurement service hosted on port: 51284
2023-09-08 11:04:51,211 INFO: Successfully registered with discovery service.
Press enter to close the measurement service.
ed OK in 0.9954 ms
2023-09-08 11:04:56,595 INFO: Starting generation: pin_name=Pin1 waveform_type=Waveform.SINE frequency=1e+06 amplitude=2
2023-09-08 11:05:06,640 INFO: gRPC server call /ni.measurementlink.measurement.v2.MeasurementService/Measure responded OK in 10045.4617 ms
2023-09-08 11:05:09,758 INFO: Starting generation: pin_name=Pin1 waveform_type=Waveform.SINE frequency=1e+06 amplitude=2
2023-09-08 11:05:10,872 INFO: gRPC server call /ni.measurementlink.measurement.v2.MeasurementService/Measure responded OK in 1115.6107 ms
2023-09-08 11:05:28,375 INFO: Starting generation: pin_name=Pin1 waveform_type=Waveform.SINE frequency=1e+06 amplitude=2
2023-09-08 11:05:29,822 INFO: gRPC server call /ni.measurementlink.measurement.v2.MeasurementService/Measure responded OK in 1447.9786 ms
```

## Possible Solution

Pass exception to `close()` method in logging code.

## Your Environment

* OS & Device: Windows
* `ni-measurementlink-service` version: 1.2.0-dev2 from main
* MeasurementLink version: 23.8.0d578
* Python version: 3.9.13

[AB#2517125](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2517125)

bkeryan:
This is partially fixed by #371 

[...]

PR #462, opened 2023-10-17
ServerCallLogger incorrectly logs OK when call is canceled #370

Avinash2Suresh:

A `CANCELLED` log is displayed instead of `UNKNOWN` when the measurement is canceled, by raising a custom RpcError within the abort function.

Fixed the [bug #370 ](https://github.com/ni/measurementlink-python/issues/370)

The measurement was manually canceled by clicking the `Stop` button in `InstrumentStudio` and the log was confirmed.

### DIFF
--- a/examples/nifgen_standard_function/_nifgen_helpers.py
+++ b/examples/nifgen_standard_function/_nifgen_helpers.py
@@ -3,10 +3,9 @@
 from typing import Any, Dict, Optional
 
 import grpc
+import ni_measurementlink_service as nims
 import nifgen
 from _constants import USE_SIMULATION
-
-import ni_measurementlink_service as nims
 
 
 def create_session(
--- a/examples/nifgen_standard_function/_nifgen_helpers.py
+++ b/examples/nifgen_standard_function/_nifgen_helpers.py
@@ -3,9 +3,10 @@
 from typing import Any, Dict, Optional
 
 import grpc
-import ni_measurementlink_service as nims
 import nifgen
 from _constants import USE_SIMULATION
+
+import ni_measurementlink_service as nims
 
 
 def create_session(
--- a/ni_measurementlink_service/_internal/grpc_servicer.py
+++ b/ni_measurementlink_service/_internal/grpc_servicer.py
@@ -87,12 +87,9 @@ class MeasurementServiceContext:
             self._grpc_context.abort(code, details)
         except Exception:
             e = grpc.RpcError()
-            # Create a lambda function
-            custom_code = lambda: code
-            # Reference the custom lambda function to defer the assignment of the 'code' attribute
-            # This approach is used because direct assignment is restricted
-            e.code = custom_code
-            raise e
+            e.code = code
+            e.details = details
+            raise e     
 
 
 measurement_service_context: ContextVar[MeasurementServiceContext] = ContextVar(
--- a/ni_measurementlink_service/_internal/grpc_servicer.py
+++ b/ni_measurementlink_service/_internal/grpc_servicer.py
@@ -90,7 +90,7 @@ class MeasurementServiceContext:
             # This allows our logging interceptors to query the code/details.
             if type(e) is Exception:
                 raise CustomRpcError(code, details) from e
-            raise e
+            raise
 
 
 class CustomRpcError(grpc.RpcError):
--- a/ni_measurementlink_service/_internal/grpc_servicer.py
+++ b/ni_measurementlink_service/_internal/grpc_servicer.py
@@ -86,10 +86,23 @@ class MeasurementServiceContext:
         try:
             self._grpc_context.abort(code, details)
         except Exception:
-            e = grpc.RpcError()
-            e.code = code
-            e.details = details
-            raise e
+            raise CustomRpcError(code)
+
+
+class CustomRpcError(grpc.RpcError):
+    """A custom exception class for handling gRPC RPC errors.
+
+    gRPC's built-in RpcError is not directly configurable in Python, so this class
+    enables the creation of custom RPC errors with specific error codes.
+    """
+
+    def __init__(self, code):
+        """Initialize a CustomRpcError instance with a specific gRPC status code."""
+        self._code = code
+
+    def code(self):
+        """Get the gRPC status code associated with this custom RPC error."""
+        return self._code
 
 
 measurement_service_context: ContextVar[MeasurementServiceContext] = ContextVar(
--- a/ni_measurementlink_service/_internal/grpc_servicer.py
+++ b/ni_measurementlink_service/_internal/grpc_servicer.py
@@ -98,7 +98,7 @@ class CustomRpcError(grpc.RpcError):
     enables the creation of custom RPC errors with specific error codes.
     """
 
-    def __init__(self, code: grpc.StatusCode):
+    def __init__(self, code: grpc.StatusCode) -> None:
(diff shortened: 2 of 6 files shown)


examples/nifgen_standard_function/_nifgen_helpers.py
```diff
@@ -3,10 +3,11 @@
 from typing import Any, Dict, Optional
 
 import grpc
-import ni_measurementlink_service as nims
 import nifgen
 from _constants import USE_SIMULATION
 
+import ni_measurementlink_service as nims
```
dixonjoel, line 9: Why is this new import required with no other changes to the file? Similar for the next two files.

ni_measurementlink_service/_internal/grpc_servicer.py
```diff
@@ -83,7 +83,13 @@ def time_remaining(self) -> float:
 
     def abort(self, code: grpc.StatusCode, details: str) -> None:
         """Aborts the RPC."""
-        self._grpc_context.abort(code, details)
+        try:
+            self._grpc_context.abort(code, details)
+        except Exception:
+            e = grpc.RpcError()
+            e.code = code
```
dixonjoel, line 90: Some lint errors here. Reset me when these are fixed.

dixonjoel (changes requested):

examples/nifgen_standard_function/_nifgen_helpers.py
Avinash2Suresh: I initially reorganized these lines in response to linting errors but subsequently reverted the changes.

ni_measurementlink_service/_internal/grpc_servicer.py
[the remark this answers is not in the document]
Avinash2Suresh: Created a subclass named 'CustomRpcError'.

ni_measurementlink_service/_internal/grpc_servicer.py
```diff
@@ -83,7 +83,13 @@ def time_remaining(self) -> float:
 
     def abort(self, code: grpc.StatusCode, details: str) -> None:
         """Aborts the RPC."""
-        self._grpc_context.abort(code, details)
+        try:
+            self._grpc_context.abort(code, details)
+        except Exception:
```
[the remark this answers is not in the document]
Avinash2Suresh: Handled it using `if isinstance(e, Exception):`

ni_measurementlink_service/_internal/grpc_servicer.py
```diff
@@ -83,7 +83,28 @@ def time_remaining(self) -> float:
...
+
+
+class CustomRpcError(grpc.RpcError):
+    """A custom exception class for handling gRPC RPC errors.
+
+    gRPC's built-in RpcError is not directly configurable in Python, so this class
+    enables the creation of custom RPC errors with specific error codes.
+    """
+
+    def __init__(self, code: grpc.StatusCode):
```
vigkre, line 101:
```suggestion
    def __init__(self, code: grpc.StatusCode) -> None:
```

vigkre (approved):

ni_measurementlink_service/_internal/grpc_servicer.py
```diff
@@ -83,7 +83,28 @@ def time_remaining(self) -> float:
 
     def abort(self, code: grpc.StatusCode, details: str) -> None:
         """Aborts the RPC."""
-        self._grpc_context.abort(code, details)
+        try:
+            self._grpc_context.abort(code, details)
+        except Exception as e:
+            if isinstance(e, Exception):
+                raise CustomRpcError(code)
+            raise e
```
bkeryan, line 91:
```suggestion
        except Exception as e:
            # If gRPC raised an empty Exception, replace it with an RpcError so our logging interceptors can query the code/details.
            if e == Exception():
                raise CustomRpcError(code, details)
            raise
```

ni_measurementlink_service/_internal/grpc_servicer.py
```diff
@@ -83,7 +83,28 @@ def time_remaining(self) -> float:
...
+
+
+class CustomRpcError(grpc.RpcError):
+    """A custom exception class for handling gRPC RPC errors.
+
+    gRPC's built-in RpcError is not directly configurable in Python, so this class
+    enables the creation of custom RPC errors with specific error codes.
+    """
+
+    def __init__(self, code: grpc.StatusCode) -> None:
```
bkeryan, line 101: Support `details` too.

ni_measurementlink_service/_internal/grpc_servicer.py
```diff
@@ -83,7 +83,28 @@ def time_remaining(self) -> float:
...
+        """Initialize a CustomRpcError instance with a specific gRPC status code."""
+        self._code = code
+
+    def code(self) -> grpc.StatusCode:
+        """Get the gRPC status code associated with this custom RPC error."""
```
bkeryan, line 106: More words are not always better.
```suggestion
        """Initialize a CustomRpcError instance."""
        self._code = code

    def code(self) -> grpc.StatusCode:
        """Get the gRPC status code."""
```

bkeryan (changes requested):

ni_measurementlink_service/_internal/grpc_servicer.py
bkeryan: One more thing: use `from e` or `from None` to control how chained exceptions are displayed.

Using `from e` would display "The above exception was the direct cause..."
Using `from None` would just display the CustomRpcError.
I'm on the fence about which is better here.

Doing nothing would display "During handling of the above exception, another exception occurred: ...".

```python
raise CustomRpcError(code, details) from e
```

ni_measurementlink_service/_internal/grpc_servicer.py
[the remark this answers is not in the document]
Avinash2Suresh: - The comparison `if e == Exception():` fails because `Exception()` and `e` represent two distinct objects and are not considered equal.
- As a workaround, it's handled using `if type(e) is Exception:` to specifically check if the type of `e` is an instance of the `Exception` class.

ni_measurementlink_service/_internal/grpc_servicer.py
```diff
@@ -83,7 +83,35 @@ def time_remaining(self) -> float:
...
         """Aborts the RPC."""
-        self._grpc_context.abort(code, details)
+        try:
+            self._grpc_context.abort(code, details)
+        except Exception as e:
+            # If gRPC raises an empty exception, replace it with an RpcError.
+            # This allows our logging interceptors to query the code/details.
+            if type(e) is Exception:
+                raise CustomRpcError(code, details) from e
+            raise e
```
mshafer-NI, line 93: in Python -> if you manually raise an exception, it rasies with this starting as the stack trace. If you put an empty `raise` that will "re-raise" without altering the stack trace

```suggestion
            raise
```

bkeryan: Thank you, I included this in my suggestions but didn't take the time to explain it.

bkeryan (changes requested):

bkeryan (approved):

dixonjoel (approved):

bkeryan:
FYI, I don't think `context.cancel()` throws an exception, so it shouldn't need the same fix.

PR #466, opened 2023-10-18
examples: [NI-Scope] Support Initialize-then-detach and Attach-then-close session initialization behavior

Jayaseelan-James:

Update the NI-Scope teststand_fixture.py with the new session management APIs and update the `closing_session` API to support "initialize-and-detach" and "attach-and-close" session initialization behavior.

Add support to "initialize-and-detach" and "attach-and-close" behavior for session initialization as discussed in this [comment](https://github.com/ni/measurementlink-python/pull/448#discussion_r1358734658) in #448.

Ran the "NIScopeAcquireWaveform.seq" and the sequence was executed without any errors.

### DIFF
--- a/examples/niscope_acquire_waveform/teststand_niscope.py
+++ b/examples/niscope_acquire_waveform/teststand_niscope.py
@@ -30,7 +30,7 @@ def create_niscope_sessions(sequence_context: Any) -> None:
         with session_management_client.reserve_sessions(
             pin_map_context, instrument_type_id=INSTRUMENT_TYPE_NI_SCOPE
         ) as reservation:
-            with reservation.create_niscope_sessions(
+            with reservation.initialize_niscope_sessions(
                 initialization_behavior=SessionInitializationBehavior.INITIALIZE_SESSION_THEN_DETACH
             ):
                 pass
@@ -53,7 +53,7 @@ def destroy_niscope_sessions() -> None:
 
             session_management_client.unregister_sessions(reservation.session_info)
 
-            with reservation.create_niscope_sessions(
+            with reservation.initialize_niscope_sessions(
                 initialization_behavior=SessionInitializationBehavior.ATTACH_TO_SESSION_THEN_CLOSE
             ):
                 pass
--- a/examples/niscope_acquire_waveform/teststand_niscope.py
+++ b/examples/niscope_acquire_waveform/teststand_niscope.py
@@ -1,10 +1,14 @@
 """Functions to set up and tear down sessions of NI-Scope devices in NI TestStand."""
 from typing import Any
 
-import ni_measurementlink_service as nims
-import niscope
 from _helpers import GrpcChannelPoolHelper, TestStandSupport
-from _niscope_helpers import create_session
+from ni_measurementlink_service.discovery import DiscoveryClient
+from ni_measurementlink_service.session_management import (
+    INSTRUMENT_TYPE_NI_SCOPE,
+    PinMapContext,
+    SessionInitializationBehavior,
+    SessionManagementClient,
+)
 
 
 def create_niscope_sessions(sequence_context: Any) -> None:
@@ -15,28 +19,21 @@ def create_niscope_sessions(sequence_context: Any) -> None:
             (Dynamically typed.)
     """
     with GrpcChannelPoolHelper() as grpc_channel_pool:
-        session_management_client = nims.session_management.Client(
-            grpc_channel=grpc_channel_pool.session_management_channel
-        )
-
         teststand_support = TestStandSupport(sequence_context)
         pin_map_id = teststand_support.get_active_pin_map_id()
+        pin_map_context = PinMapContext(pin_map_id=pin_map_id, sites=None)
 
-        pin_map_context = nims.session_management.PinMapContext(pin_map_id=pin_map_id, sites=None)
-        grpc_device_channel = grpc_channel_pool.get_grpc_device_channel(
-            niscope.GRPC_SERVICE_INTERFACE_NAME
+        discovery_client = DiscoveryClient(grpc_channel_pool=grpc_channel_pool)
+        session_management_client = SessionManagementClient(
+            discovery_client=discovery_client, grpc_channel_pool=grpc_channel_pool
         )
         with session_management_client.reserve_sessions(
-            context=pin_map_context,
-            instrument_type_id=nims.session_management.INSTRUMENT_TYPE_NI_SCOPE,
+            pin_map_context, instrument_type_id=INSTRUMENT_TYPE_NI_SCOPE
         ) as reservation:
-            for session_info in reservation.session_info:
-                # Leave session open
-                _ = create_session(
-                    session_info,
-                    grpc_device_channel,
-                    initialization_behavior=niscope.SessionInitializationBehavior.INITIALIZE_SERVER_SESSION,
-                )
+            with reservation.create_niscope_sessions(
+                initialization_behavior=SessionInitializationBehavior.INITIALIZE_SESSION_NO_CLOSE
+            ):
+                pass
 
             session_management_client.register_sessions(reservation.session_info)
 
@@ -44,22 +41,19 @@ def create_niscope_sessions(sequence_context: Any) -> None:
 def destroy_niscope_sessions() -> None:
     """Destroy and unregister all NI-Scope sessions."""
     with GrpcChannelPoolHelper() as grpc_channel_pool:
-        session_management_client = nims.session_management.Client(
-            grpc_channel=grpc_channel_pool.session_management_channel
-        )
-        grpc_device_channel = grpc_channel_pool.get_grpc_device_channel(
-            niscope.GRPC_SERVICE_INTERFACE_NAME
+        discovery_client = DiscoveryClient(grpc_channel_pool=grpc_channel_pool)
+        session_management_client = SessionManagementClient(
+            discovery_client=discovery_client, grpc_channel_pool=grpc_channel_pool
         )
         with session_management_client.reserve_all_registered_sessions(
-            instrument_type_id=nims.session_management.INSTRUMENT_TYPE_NI_SCOPE,
+            instrument_type_id=INSTRUMENT_TYPE_NI_SCOPE,
         ) as reservation:
+            if not reservation.session_info:
+                return
+
             session_management_client.unregister_sessions(reservation.session_info)
 
-            for session_info in reservation.session_info:
-                session = create_session(
-                    session_info,
-                    grpc_device_channel,
-                    initialization_behavior=niscope.SessionInitializationBehavior.ATTACH_TO_SERVER_SESSION,
-                )
-
-                session.close()
+            with reservation.create_niscope_sessions(
+                initialization_behavior=SessionInitializationBehavior.ATTACH_TO_SESSION_AUTO_CLOSE
+            ):
+                pass
--- a/examples/niscope_acquire_waveform/teststand_niscope.py
+++ b/examples/niscope_acquire_waveform/teststand_niscope.py
@@ -31,7 +31,7 @@ def create_niscope_sessions(sequence_context: Any) -> None:
             pin_map_context, instrument_type_id=INSTRUMENT_TYPE_NI_SCOPE
         ) as reservation:
             with reservation.create_niscope_sessions(
-                initialization_behavior=SessionInitializationBehavior.INITIALIZE_SESSION_NO_CLOSE
+                initialization_behavior=SessionInitializationBehavior.INITIALIZE_SESSION_THEN_DETACH
             ):
                 pass
 
@@ -54,6 +54,6 @@ def destroy_niscope_sessions() -> None:
             session_management_client.unregister_sessions(reservation.session_info)
 
             with reservation.create_niscope_sessions(
-                initialization_behavior=SessionInitializationBehavior.ATTACH_TO_SESSION_AUTO_CLOSE
+                initialization_behavior=SessionInitializationBehavior.ATTACH_TO_SESSION_THEN_CLOSE
             ):
                 pass
--- a/ni_measurementlink_service/_drivers/__init__.py
+++ b/ni_measurementlink_service/_drivers/__init__.py
@@ -2,12 +2,17 @@
 from __future__ import annotations
 
 import contextlib
-from typing import ContextManager, TypeVar
+from typing import ContextManager, TypeVar, cast
+
+from ni_measurementlink_service.session_management._types import SessionInitializationBehavior
 
 TSession = TypeVar("TSession")
 
 
-def closing_session(session: TSession) -> ContextManager[TSession]:
+def closing_session(
+    session: TSession,
+    initialization_behavior: SessionInitializationBehavior = SessionInitializationBehavior.AUTO,
+) -> ContextManager[TSession]:
     """Create a context manager that closes the session.
 
     Args:
@@ -16,9 +21,15 @@ def closing_session(session: TSession) -> ContextManager[TSession]:
     Returns:
         A context manager that yields the session and closes it.
     """
-    if isinstance(session, contextlib.AbstractContextManager):
-        # Assume the session yields itself.
-        return session
+    if initialization_behavior == SessionInitializationBehavior.AUTO:
+        # yield the session's context manager.
+        return cast(ContextManager[TSession], session)
+    elif (
+        initialization_behavior
+        == SessionInitializationBehavior.INITIALIZE_SERVER_SESSION_THEN_DETACH
+        or initialization_behavior == SessionInitializationBehavior.ATTACH_TO_SERVER_SESSION
+    ):
+        return contextlib.nullcontext(session)
     elif hasattr(session, "close"):
         return contextlib.closing(session)
     else:
--- a/ni_measurementlink_service/_drivers/__init__.py
+++ b/ni_measurementlink_service/_drivers/__init__.py
@@ -20,8 +20,9 @@ def closing_session_with_ts_code_module_support(
     session: TSession,
 ) -> ContextManager[TSession]:
     """Create a context manager that closes the session.
-    
-    Emulates the behavior of INITIALIZE_SESSION_THEN_DETACH and ATTACH_TO_SESSION_THEN_CLOSE."""
+
+    Emulates the behavior of INITIALIZE_SESSION_THEN_DETACH and ATTACH_TO_SESSION_THEN_CLOSE.
+    """
     if not hasattr(session, "close"):
         raise TypeError("Session must have a close() method.")
 
@@ -33,7 +34,7 @@ def closing_session_with_ts_code_module_support(
         return contextlib.closing(session)
     else:
         raise ValueError(f"Invalid initialization behavior: '{initialization_behavior}'.")
-    
+
 
 def closing_session(session: TSession) -> ContextManager[TSession]:
     """A context manager that yields the session and closes it."""
--- a/ni_measurementlink_service/_drivers/__init__.py
+++ b/ni_measurementlink_service/_drivers/__init__.py
@@ -16,21 +16,16 @@ _EXISTING_BEHAVIORS = [
 
 
 def closing_session(
+    initialization_behavior: SessionInitializationBehavior,
     session: TSession,
-    initialization_behavior: SessionInitializationBehavior = SessionInitializationBehavior.AUTO,
-    can_emulate: bool = True,
 ) -> ContextManager[TSession]:
     """Create a context manager that closes the session.
 
     Args:
-        session: A driver session.
-
         initialization_behavior: Specifies whether to initialize a new
             session or attach to an existing session.
 
-        can_emulate: Specify True, if MeasurementLink must emulate the
-            INITIALIZE_SERVER_SESSION_THEN_DETACH and
-            ATTACH_TO_SERVER_SESSION_THEN_CLOSE behavior. Else, False.
+        session: A driver session.
 
     Returns:
         A context manager that yields the session and closes it.
@@ -45,11 +40,11 @@ def closing_session(
     elif initialization_behavior not in SessionInitializationBehavior:
         raise ValueError(f"Invalid initialization behavior: '{initialization_behavior}'.")
 
-    if not can_emulate or initialization_behavior in _EXISTING_BEHAVIORS:
+    if initialization_behavior in _EXISTING_BEHAVIORS:
         if not isinstance(session, contextlib.AbstractContextManager):
             raise TypeError("Session must be a context manager.")
         return session
 
-    if initialization_behavior == SessionInitializationBehavior.INITIALIZE_SESSION_NO_CLOSE:
+    if initialization_behavior == SessionInitializationBehavior.INITIALIZE_SESSION_THEN_DETACH:
         return contextlib.nullcontext(session)
     return contextlib.closing(session)
--- a/ni_measurementlink_service/_drivers/__init__.py
+++ b/ni_measurementlink_service/_drivers/__init__.py
@@ -15,36 +15,28 @@ _EXISTING_BEHAVIORS = [
 ]
 
 
-def closing_session(
+def closing_session_with_ts_code_module_support(
     initialization_behavior: SessionInitializationBehavior,
     session: TSession,
 ) -> ContextManager[TSession]:
     """Create a context manager that closes the session.
-
-    Args:
-        initialization_behavior: Specifies whether to initialize a new
-            session or attach to an existing session.
-
-        session: A driver session.
-
-    Returns:
-        A context manager that yields the session and closes it.
-
-    Raises:
-        TypeError: If the session is not a context manager and if it does not have a close() method.
-
-        ValueError: If the initialization behavior is invalid.
-    """
+    
+    Emulates the behavior of INITIALIZE_SESSION_THEN_DETACH and ATTACH_TO_SESSION_THEN_CLOSE."""
     if not hasattr(session, "close"):
         raise TypeError("Session must have a close() method.")
-    elif initialization_behavior not in SessionInitializationBehavior:
-        raise ValueError(f"Invalid initialization behavior: '{initialization_behavior}'.")
 
     if initialization_behavior in _EXISTING_BEHAVIORS:
-        if not isinstance(session, contextlib.AbstractContextManager):
-            raise TypeError("Session must be a context manager.")
-        return session
-
-    if initialization_behavior == SessionInitializationBehavior.INITIALIZE_SESSION_THEN_DETACH:
+        return closing_session(session)
+    elif initialization_behavior == SessionInitializationBehavior.INITIALIZE_SESSION_THEN_DETACH:
         return contextlib.nullcontext(session)
-    return contextlib.closing(session)
+    elif initialization_behavior == SessionInitializationBehavior.ATTACH_TO_SESSION_THEN_CLOSE:
+        return contextlib.closing(session)
+    else:
+        raise ValueError(f"Invalid initialization behavior: '{initialization_behavior}'.")
+    
+
+def closing_session(session: TSession) -> ContextManager[TSession]:
+    """A context manager that yields the session and closes it."""
+    if not isinstance(session, contextlib.AbstractContextManager):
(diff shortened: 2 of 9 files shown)


ni_measurementlink_service/_drivers/__init__.py
```diff
@@ -16,9 +20,12 @@ def closing_session(session: TSession) -> ContextManager[TSession]:
     Returns:
         A context manager that yields the session and closes it.
     """
-    if isinstance(session, contextlib.AbstractContextManager):
+    if isinstance(session, contextlib.AbstractContextManager) and initialization_behavior == SessionInitializationBehavior.AUTO:
```
Jayaseelan-James, line 23: As described in this HLD, the ideal behavior for AUTO is "to Init then Close if the instrument session is not existed on grpc-device server, attach then detach if the instrument session is existed on grpc-device server" but the current implementation doesn't fully incorporate this behavior.

I'll continue on exploring that part and will update the PR in the coming days. The other 4 enum behaviour is showcased for the niscope session with TestStand code module changes.

dixonjoel:
Did you also run 'poetry run pytest -v' locally to verify all the tests including those that require the discovery service to be installed?

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -302,7 +303,7 @@ def _create_session_core(
             )
 
         session_info = session_infos[0]
-        with closing_session(session_constructor(session_info)) as session:
+        with closing_session(session_constructor(session_info), initialization_behavior) as session:
```
jasonmreding, line 306: Doesn't the initialization parameter also need to be bubbled out through the "public" APIs for creating sessions?

ni_measurementlink_service/session_management/_types.py
```diff
@@ -320,3 +320,21 @@ class SessionInitializationBehavior(IntEnum):
     Note: When using the Session as a context manager and the context exits, it
     will detach from the server session and leave it open.
     """
+
+    INITIALIZE_SERVER_SESSION_THEN_DETACH = 3
+    """
+    Require the NI gRPC Device Server to initialize a new session with the
```
jasonmreding, line 326: If this is meant to work with third party drivers, we might want to remove references to "gRPC Device Server" from the help. Similarly, we should probably update the documentation to explain the "close behavior" only applies when used as a context manager. Otherwise, the user still has to call the driver close API directly with whatever behavior is desired. I don't recall if we were going to offer a generic `close_sessions` API directly from the reservation class like there is in LV. If there is going to be one, then that API won't have an input for the close behavior. Instead, the behavior specified when creating the session will be applied.

jasonmreding: I found the asymmetry between the enum value names a bit weird, but I see that you are just following what VISA has planned here. We'll have to decide how much we value consistency vs. symmetry here (or get VISA to change their plans to use symmetrical names.

ni_measurementlink_service/_drivers/__init__.py
jasonmreding: I might be missing something, but it seems like what you have here should more or less work. I don't know if this is also used by third party drivers or just NI drivers. 

My only other comment is this code makes certain assumptions about the driver session object's behavior. For instance, the context manager implemented by the session for VISA could likely be used as is all of the time. Similarly, I could see how some drivers might have a `Detach` API that should be called rather than just not calling anything. If you wanted to allow for that, I think you would have to add adapter code in each driver's `SessionConstructor` class rather than trying to share the code. I guess you could still call a shared function from the driver specific constructor if there are common implementation patterns used by various drivers.

ni_measurementlink_service/session_management/_types.py
Jayaseelan-James: > I don't recall if we were going to offer a generic close_sessions API directly from the reservation class like there is in LV.

In Python, the `create_foo_session(s)` API will return the session as a context manager whose exit block will take care of the Session's close behavior internally. Hence, we won't be exposing any generic or driver-specific `close_sessions` API.

dixonjoel (changes requested):
Overall I like the direction. I'd like to agree on the enum names and see some more test combinations in test_drivers that show how these things work with context managers and that the _DETACH ones don't close it when exiting the context.

ni_measurementlink_service/_drivers/__init__.py
```diff
@@ -16,9 +21,15 @@ def closing_session(session: TSession) -> ContextManager[TSession]:
...
-    if isinstance(session, contextlib.AbstractContextManager):
-        # Assume the session yields itself.
-        return session
+    if initialization_behavior == SessionInitializationBehavior.AUTO:
+        # yield the session's context manager.
+        return cast(ContextManager[TSession], session)
+    elif (
+        initialization_behavior
+        == SessionInitializationBehavior.INITIALIZE_SERVER_SESSION_THEN_DETACH
+        or initialization_behavior == SessionInitializationBehavior.ATTACH_TO_SERVER_SESSION
```
bkeryan, line 30: You can make these more concise by looking up the enum value in a list. Define a "constant" list at file scope so you don't have to construct a new list every time. 

```
_DETACH_BEHAVIORS = [
    SessionInitializationBehavior.ATTACH_TO_SERVER_SESSION,
    SessionInitializationBehavior.INITIALIZE_SERVER_SESSION_THEN_DETACH,
]
_CLOSE_BEHAVIORS = [
    SessionInitializationBehavior.INITIALIZE_SERVER_SESSION,
    SessionInitializationBehavior.ATTACH_TO_SERVER_SESSION_THEN_CLOSE,
]
...

    elif initialization_behavior in _DETACH_BEHAVIORS:
        ...
    elif initialization_behavior in _CLOSE_BEHAVIORS:
        ...
```

ni_measurementlink_service/session_management/_reservation.py
bkeryan: It is not in the generic API.

```
    @requires_feature(SESSION_MANAGEMENT_2024Q1)
    def create_session(
        self,
        session_constructor: Callable[[SessionInformation], TSession],
        instrument_type_id: str,
    ) -> ContextManager[TypedSessionInformation[TSession]]:
```

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -286,6 +286,7 @@ def _create_session_core(
         self,
         session_constructor: Callable[[SessionInformation], TSession],
         instrument_type_id: str,
+        initialization_behavior: SessionInitializationBehavior = SessionInitializationBehavior.AUTO,
```
bkeryan, line 289: Post-RFC, make this parameter required in the `_core` methods to ensure that we don't accidentally leave it out of any of the callers.

ni_measurementlink_service/session_management/_types.py
bkeryan: This enum started out as a copy of the nimi-python one that we can share between drivers. That's why the docs refer to NI gRPC Device Server.

I don't mind changing the docs and enum values to be more general, if it makes sense.

```
ATTACH_TO_SERVER_SESSION = 3, # (This will attach and close)
``` 
I don't like this because it makes the atypical behavior (mixing attach and close) implicit.

The typical behavior is init/close or attach/detach.

The asymmetry calls out the fact that these two cases are atypical: init/detach or attach/close.

```
INITIALIZE_SESSION_AUTO_CLOSE
```
These make it harder to distinguish between the typical and atypical cases.

tests/unit/test_drivers.py
```diff
@@ -1,9 +1,13 @@
 from ni_measurementlink_service._drivers import closing_session
+from ni_measurementlink_service.session_management._types import SessionInitializationBehavior
 from tests.utilities import fake_driver
 
 
 def test___closable_session___with_closing_session___session_closed() -> None:
-    with closing_session(fake_driver.ClosableSession("Dev1")) as session:
+    with closing_session(
+        fake_driver.ClosableSession("Dev1"),
+        SessionInitializationBehavior.INITIALIZE_SERVER_SESSION,
```
bkeryan, line 9: Post-RFC, please add test cases for all of the init behaviors.

ni_measurementlink_service/_drivers/__init__.py
```diff
@@ -16,9 +21,15 @@ def closing_session(session: TSession) -> ContextManager[TSession]:
...
+    ):
+        return contextlib.nullcontext(session)
     elif hasattr(session, "close"):
         return contextlib.closing(session)
```
bkeryan, line 34: One more thought: for nimi-python and nidaqmx-python, nullcontext and closing are only needed for the two new behaviors. The three existing behaviors are already implemented by the context manager code in nimi-python and nidaqmx-python.

However, using nullcontext and closing for the ATTACH_SERVER_SESSION and INITIALIZE_SERVER_SESSION may eliminate some subtlety for user-provided sessions. Maybe this is a good thing.

bkeryan (changes requested):

ni_measurementlink_service/_drivers/__init__.py
bkeryan, line 1: On further thought: we are emulating behavior that should be implemented in the driver API, so we also need to handle the case where the driver API natively supports this behavior. I think this may require another parameter to opt in/out of this emulation, in the `_core` and generic APIs.

The immediate motivation for this is VISA. We are not currently planning to develop a PyVISA gRPC plugin, so VISA sessions will use the VISA DLL and VISA gRPC passport (native VISA gRPC plugin). This means that they have native resources that must be cleaned up, even when detaching from the server session. 

The client-side VISA session stack will look something like this:
- Instrument driver session (Python)
- PyVISA session (Python)
- VISA session (native)
- VISA gRPC passport session (native)
- gRPC client (native)

This doesn't currently affect nimi-python and nidaqmx-python because they are "pure Python" and use a Python gRPC channel that is managed by the application. It's okay to leak a nimi-python or nidaqmx-python session as long as you clean up the underlying gRPC channel object.

If we add the asymmetrical init+detach and attach+close enum values to nimi-python and nidaqmx-python later, we will want to disable this emulation for those APIs as well.

bkeryan (changes requested):

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: I've added them now.

ni_measurementlink_service/session_management/_types.py
Jayaseelan-James: For now, I've updated the atypical enum values alone with Joel's suggestions. If everyone agrees with the naming, I'll update the existing enum values with the approved suggestions.

examples/niscope_acquire_waveform/teststand_fixture.py
```diff
@@ -38,51 +42,42 @@ def create_niscope_sessions(sequence_context: Any) -> None:
...
-                _ = create_session(
-                    session_info,
-                    grpc_device_channel,
-                    initialization_behavior=niscope.SessionInitializationBehavior.INITIALIZE_SERVER_SESSION,
-                )
-
+            # Initialize new sessions on the server, but do not close them.
+            with reservation.create_niscope_sessions(
+                initialization_behavior=SessionInitializationBehavior.INITIALIZE_SERVER_SESSION_THEN_DETACH
+            ) as _ :
```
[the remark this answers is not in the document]
Jayaseelan-James: Yes, Joel. I've removed the `as _` now.

ni_measurementlink_service/_drivers/__init__.py
```diff
@@ -16,9 +21,15 @@ def closing_session(session: TSession) -> ContextManager[TSession]:
     Returns:
         A context manager that yields the session and closes it.
     """
-    if isinstance(session, contextlib.AbstractContextManager):
-        # Assume the session yields itself.
-        return session
+    if initialization_behavior == SessionInitializationBehavior.AUTO:
+        # yield the session's context manager.
+        return cast(ContextManager[TSession], session)
```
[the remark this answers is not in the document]
Jayaseelan-James: Thanks for the detailed explanation, Brad. I've updated the code module to reflect the behavior that you stated. Please review and provide your thoughts on the same.

ni_measurementlink_service/_drivers/__init__.py
Jayaseelan-James: Thanks for the suggestion. For the revised logic, I've used the _EXISTING_BEHAVIORS constant that holds the 3 typical initialization behaviors.

``` python
_EXISTING_BEHAVIORS = [
    SessionInitializationBehavior.AUTO,
    SessionInitializationBehavior.ATTACH_TO_SERVER_SESSION,
    SessionInitializationBehavior.INITIALIZE_SERVER_SESSION,
]
```

ni_measurementlink_service/_drivers/__init__.py
Jayaseelan-James: I've added a `can_emulate` boolean for the `_core` and generic APIs with the default value as "False" since we expect the drivers to natively support this behavior, but in the `closing_session` API the default value is set to "True" since we don't have driver support for these new behaviors at the moment (this is only to refrain from passing "True" from all the nimi-python and nidaqmx-python drivers).

> This means that they have native resources that must be cleaned up, even when detaching from the server session.

By this, do you mean, that when detaching from the server session MeasurmentLink must clean up the native resources? 
As far as I understand, and at the moment, we're only calling `contextlib.nullcontext` for the new detach behavior, what should I additionally do to clean the native resources?

dixonjoel (changes requested):
The initialization behaviors code looks good. Setting Request Changes so you can reset me when the tests are ready and / or it's no longer an RFC.

ni_measurementlink_service/_drivers/__init__.py
```diff
@@ -4,25 +4,52 @@
...
-        return contextlib.closing(session)
-    else:
-        raise TypeError(
-            f"Invalid session type '{type(session)}'. A session must be a context manager and/or "
-            "have a close() method."
-        )
+
+    if initialization_behavior == SessionInitializationBehavior.INITIALIZE_SESSION_NO_CLOSE:
+        return contextlib.nullcontext(session)
+    return contextlib.closing(session)
```
bkeryan, line 55: use elif

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -286,6 +286,8 @@ def _create_session_core(
...
+        can_emulate: bool = True,
```
bkeryan, line 290: Replace with a callable as described in the other thread.

ni_measurementlink_service/_drivers/__init__.py
bkeryan: Also, let's try leaving `closing_function` out of the generic API, since VISA will use the default behavior.

bkeryan (changes requested):

ni_measurementlink_service/_drivers/__init__.py
Jayaseelan-James: If I add an `elif` case, then "mypy" insists on adding an else case but logically all the enum values are covered until or within the elif case itself (if I add one), so adding an else case seemed redundant. That was the reason to return "contextlib.closing(session)" directly. 

So, would you still recommend having an `elif` case and then within the `else` block, return the "session" object or another `return contextlib.closing(session)` statement?

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: Updated.

ni_measurementlink_service/_drivers/__init__.py
Jayaseelan-James: I've updated the code modules with the `Callable` parameter as per your suggestion.

tests/utilities/fake_driver.py
```diff
@@ -17,34 +17,11 @@
 
 GRPC_SERVICE_INTERFACE_NAME = "nifake_grpc.NiFake"
 
-# The GrpcSessionOptions classes in nimi-python and nidaqmx-python have an api_key field.
```
Jayaseelan-James, line 20: Team, I've removed the below constant (`_API_KEY`) and classes (`SessionInitializationBehavior` & `GrpcSessionOptions`) as they're not used anywhere. 

Please let me know if these were intended for any future purpose.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -1305,7 +1319,10 @@ def create_niscope_session(
             options,
             initialization_behavior,
         )
-        return self._create_session_core(session_constructor, INSTRUMENT_TYPE_NI_SCOPE)
+        closing_function = functools.partial(closing_session, initialization_behavior)
```
Jayaseelan-James, line 1322: In this PR, only the NI-Scope specific APIs are updated i.e., the `create_niscope_session` and `create_niscope_sessions` driver APIs in _reservation.py and the `create_niscope_sessions` and `destroy_niscope_sessions` in the [teststand_niscope.py](https://github.com/ni/measurementlink-python/blob/users/jay/enum_support_for_detach_close/examples/niscope_acquire_waveform/teststand_niscope.py). 

All the other driver API and TestStand code module updates will be done as part of [AB#2552333](https://dev.azure.com/ni/DevCentral/_workitems/edit/2552333).

vigkre (approved):

tests/utilities/fake_driver.py
bkeryan: They are intended to model a gRPC-aware driver API. Put them back.

dixonjoel (approved):

tests/utilities/fake_driver.py
```diff
@@ -2,13 +2,13 @@
...
 
 import sys
-from enum import Enum, IntEnum
+from enum import Enum
 from types import TracebackType
 from typing import TYPE_CHECKING, Any, ContextManager, Dict, Optional, Type
 
-if TYPE_CHECKING:
-    import grpc
+from ni_measurementlink_service.session_management._types import SessionInitializationBehavior
```
bkeryan, line 9: Driver APIs do not depend on ni_measurementlink_service.

Driver APIs define their own SessionInitializationBehavior enums.

The MeasurementLink specific glue for each driver API maps the MeasurementLink enum to the driver enum, but this doesn't belong in the driver API.

tests/utilities/fake_driver.py
```diff
@@ -145,4 +122,8 @@ def __exit__(
         traceback: Optional[TracebackType],
     ) -> None:
         """Exit the session's runtime context."""
-        self.close()
+        if (
+            "initialization_behavior" not in self.options
```
bkeryan, line 126: initialization_behavior should be a separate constructor parameter like in the MI drivers, not part of the IVI options dictionary.

ni_measurementlink_service/_drivers/__init__.py
bkeryan: If you validate initialization_behavior based on the cases covered by the function, it should work out. This is better because it does the right thing if we add an enum value without updating this function.

```python
    if initialization_behavior in _EXISTING_BEHAVIORS:
        if not isinstance(session, contextlib.AbstractContextManager):
            raise TypeError("Session must be a context manager.")
        return session
    elif initialization_behavior == SessionInitializationBehavior.INITIALIZE_SESSION_THEN_DETACH:
        return contextlib.nullcontext(session)
    elif initialization_behavior == SessionInitializationBehavior.ATTACH_SESSION_THEN_CLOSE:
        return contextlib.closing(session)
    else:
        raise ValueError(f"Invalid initialization behavior: '{initialization_behavior}'.")
```

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -281,11 +282,17 @@ def _get_matching_session_infos(self, instrument_type_id: str) -> List[SessionIn
             if instrument_type_id and instrument_type_id == info.instrument_type_id
         ]
 
+    def _default_closing(self, session: TSession) -> ContextManager[TSession]:
```
bkeryan, line 285: This should be a free function, not a method. That's why my code snippet didn't indent this function or give it a "self" parameter.

ni_measurementlink_service/session_management/_types.py
```diff
@@ -305,18 +305,32 @@ class SessionInitializationBehavior(IntEnum):
...
 
     Note: When using the Session as a context manager and the context exits, it
     will detach from the server session and leave it open.
     """
+
+    INITIALIZE_SESSION_THEN_DETACH = 3
+    """
+    Initialize a new session with the specified name.
+
+    Note: When using the Session as a context manager and the context exits, it
```
bkeryan, line 326: For the two new enum values, please highlight the asymmetry instead of hiding it.

```
"""
Initialize a new session. When exiting the context manager, detach instead of closing.

Note: This initialization behavior is intended for TestStand code modules used in ProcessSetup.
"""
```

Likewise for the other one.

bkeryan (changes requested):

ni_measurementlink_service/_drivers/__init__.py
```diff
@@ -4,25 +4,47 @@
...
+
         session: A driver session.
 
     Returns:
         A context manager that yields the session and closes it.
+
+    Raises:
+        TypeError: If the session is not a context manager and if it does not have a close() method.
+
+        ValueError: If the initialization behavior is invalid.
```
bkeryan, line 36: This is an internal function, so its docstring doesn't need all this detail.

What it does need is to say that this function emulates the behavior of INITIALIZE_SESSION_THEN_DETACH and ATTACH_SESSION_THEN_CLOSE.

ni_measurementlink_service/session_management/_reservation.py
bkeryan: On further thought, it would be better to move this function to _drivers/__init__.py, rename it to `closing_session`, and rename the other function to something like `closing_session_with_ts_code_module_support`.

ni_measurementlink_service/_drivers/__init__.py
```diff
@@ -4,25 +4,47 @@
...
-        # Assume the session yields itself.
+    if not hasattr(session, "close"):
+        raise TypeError("Session must have a close() method.")
+    elif initialization_behavior not in SessionInitializationBehavior:
+        raise ValueError(f"Invalid initialization behavior: '{initialization_behavior}'.")
+
+    if initialization_behavior in _EXISTING_BEHAVIORS:
+        if not isinstance(session, contextlib.AbstractContextManager):
+            raise TypeError("Session must be a context manager.")
         return session
```
bkeryan, line 46: Once you move the default closing function into this file, you can call it here to make it clearer that this function extends the default function's behavior.

examples/niscope_acquire_waveform/teststand_niscope.py
```diff
@@ -15,51 +19,41 @@ def create_niscope_sessions(sequence_context: Any) -> None:
...
+        pin_map_context = PinMapContext(pin_map_id=pin_map_id, sites=None)
 
-        pin_map_context = nims.session_management.PinMapContext(pin_map_id=pin_map_id, sites=None)
-        grpc_device_channel = grpc_channel_pool.get_grpc_device_channel(
-            niscope.GRPC_SERVICE_INTERFACE_NAME
+        discovery_client = DiscoveryClient(grpc_channel_pool=grpc_channel_pool)
+        session_management_client = SessionManagementClient(
+            discovery_client=discovery_client, grpc_channel_pool=grpc_channel_pool
         )
         with session_management_client.reserve_sessions(
```
bkeryan, line 30: You need to merge main again and retrofit for the changes in #472

ni_measurementlink_service/session_management/_types.py
```diff
@@ -309,18 +309,34 @@ class SessionInitializationBehavior(IntEnum):
...
     will detach from the server session and leave it open.
     """
+
+    INITIALIZE_SESSION_THEN_DETACH = 3
+    """
+    Initialize a new session. 
+    
+    When exiting the context manager, detach instead of closing.
+
+    Note: This initialization behavior is intended for TestStand code modules used in ProcessSetup.
```
bkeryan, line 332:
```suggestion
    Note: This initialization behavior is intended for TestStand code modules used in ``Setup`` steps or ``ProcessSetup`` callback sequences.
```

ni_measurementlink_service/session_management/_types.py
```diff
@@ -309,18 +309,34 @@ class SessionInitializationBehavior(IntEnum):
...
+    """
+
+    ATTACH_TO_SESSION_THEN_CLOSE = 4
+    """
+    Attach to an existing session. 
+    
+    When exiting the context manager, automatically close the server session.
+
+    Note: This initialization behavior is intended for TestStand code modules used in Cleanup.
```
bkeryan, line 341:
```suggestion
    Note: This initialization behavior is intended for TestStand code modules used in ``Cleanup`` steps or ``ProcessCleanup`` callback sequences.
```

tests/unit/test_drivers.py
```diff
@@ -1,26 +1,87 @@
...
-def test___closable_session___with_closing_session___session_closed() -> None:
-    with closing_session(fake_driver.ClosableSession("Dev1")) as session:
-        assert isinstance(session, fake_driver.ClosableSession)
+@pytest.mark.parametrize(
+    "initialization_behavior",
+    [
+        (SessionInitializationBehavior.AUTO),
+        (SessionInitializationBehavior.INITIALIZE_SERVER_SESSION),
+        (SessionInitializationBehavior.ATTACH_TO_SESSION_THEN_CLOSE),
+    ],
```
bkeryan, line 14: These parentheses have no effect. Constructing a one-element tuple requires a comma, but you don't need a tuple for a single pytest parameter.

This applies to the other test functions too.

```suggestion
    "initialization_behavior",
    [
        SessionInitializationBehavior.AUTO,
        SessionInitializationBehavior.INITIALIZE_SERVER_SESSION,
        SessionInitializationBehavior.ATTACH_TO_SESSION_THEN_CLOSE,
    ],
```

bkeryan (approved):
Approved with suggestions

bkeryan:
> Approved with suggestions

I spoke too soon. Your most recent changes include an incorrect merge.

tests/utilities/fake_driver.py
```diff
@@ -21,12 +23,18 @@
 _API_KEY = "00000000-0000-0000-0000-000000000000"
 
 
-class SessionInitializationBehavior(IntEnum):
-    """Specifies whether to initialize a new session or attach to an existing session."""
+_CLOSE_BEHAVIORS = [
+    SessionInitializationBehavior.AUTO,
+    SessionInitializationBehavior.INITIALIZE_SERVER_SESSION,
+    SessionInitializationBehavior.ATTACH_TO_SESSION_THEN_CLOSE,
+]
```
bkeryan, line 30: Revert this. fake_driver.py should include its own copy of the enum and only one copy of _CLOSE_BEHAVIORS.

bkeryan (changes requested):

bkeryan (approved):

ni_measurementlink_service/session_management/_types.py
bkeryan: > If this is meant to work with third party drivers, we might want to remove references to "gRPC Device Server" from the help. Similarly, we should probably update the documentation to explain the "close behavior" only applies when used as a context manager.

@jasonmreding I think this is resolved.

bkeryan:
> @Avinash2Suresh
[Revert "chore(deps): lock file maintenance (](https://github.com/ni/measurementlink-python/pull/466/commits/071b3b8ef756f6d5a6054afceefb0000888c4c20)https://github.com/ni/measurementlink-python/pull/477[)"](https://github.com/ni/measurementlink-python/pull/466/commits/071b3b8ef756f6d5a6054afceefb0000888c4c20) 
[071b3b8](https://github.com/ni/measurementlink-python/pull/466/commits/071b3b8ef756f6d5a6054afceefb0000888c4c20)
This reverts commit https://github.com/ni/measurementlink-python/commit/4937c11888cfc575f9a361be4bc0cfd125e61598.

@Avinash2Suresh GitHub didn't show this revert in the diff, so I thought this PR would leave the poetry.lock files alone. However, when I merged your PR, it reverted the poetry.lock files. 

Since it's just lock files, I'll just have Renovate update them again. If this had reverted other developers' changes, it might make sense to revert your PR so you can redo the merge correctly.

PR #486, opened 2023-10-28
Speed up GitHub workflows

bkeryan:

- Caching improvements
  - Cache in-project virtualenvs. All of our caching code was still using `~/.cache/pypoetry/virtualenvs`.
  - I originally tried using actions/setup-python's built-in caching, but I ran into two problems:
    - It requires installing Poetry before Python (e.g. with pipx), and when I did this, tests used the wrong Python version. (pipx is slightly faster than Gr1N/setup-poetry, though.)
    - Its cache key setup is limited to simple use cases and can't handle optional extras or Poetry dependency groups.
  - Use restore/save for workflows that run `poetry install` multiple times with different extras and/or dependency groups.
  - Cache example `poetry.lock` files. These lock files are not checked into Git, but generating them takes over 5 minutes, so they are worth caching.
  - For system tests, cache the `.tox` directory as well. Caching on the self-hosted runner has significant overhead, but it seems to be an overall win compared to installing from PyPI for each Python version.
  - Remove caching from `Publish_NIMS.yml`. We don't run this workflow very often and it's better to avoid a possible source of failure.
- Disable unit tests on macOS
  - Running tests on macOS often takes longer than on Linux and Windows
  - Caching virtualenvs on macOS was returning errors about invalid/corrupted virtualenvs. Maybe this was because I was skipping `poetry install` on cache hits at the time, but I don't want to spend any more time on it.

Speed up builds (when there is a cache hit):
- Check NIMG: 1 min -> 30 sec
- Check NIMS: 2 min 30 sec -> 1 min 15 sec
- Check examples: 12-15 min -> 2 min 30 sec
- Run unit tests (Linux): 3 min -> 2 min
- Run unit tests (Windows): 4-6 min -> 3-4 min
- Run system tests: 21-23 min -> 15-16 min (not counting time to scale up new runner instance)
- Total run time: 2 hr -> 1 hr 
- Total duration: 40 min -> 22 min

Ran PR workflow.

### DIFF
--- a/.github/workflows/run_system_tests.yml
+++ b/.github/workflows/run_system_tests.yml
@@ -22,6 +22,8 @@ jobs:
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
+      - name: Add gzip to path so cache can use tar -z
+        run: echo "C:\Program Files\Git\usr\bin" >> $GITHUB_PATH
       - name: Cache virtualenvs
         uses: actions/cache@v3
         id: cache
--- a/.github/workflows/run_system_tests.yml
+++ b/.github/workflows/run_system_tests.yml
@@ -31,8 +31,7 @@ jobs:
             .tox
           key: run-system-tests-${{ runner.os }}-${{ matrix.configuration }}-${{ hashFiles('poetry.lock') }}
       - name: Install dependencies
-        if: steps.cache.outputs.cache-hit != 'true'
-        run: poetry install
+        run: poetry install -v
       - name: Run system tests
         run: poetry run tox
       - name: Upload test results
--- a/.github/workflows/run_system_tests.yml
+++ b/.github/workflows/run_system_tests.yml
@@ -24,12 +24,14 @@ jobs:
         uses: actions/checkout@v4
       - name: Cache virtualenvs
         uses: actions/cache@v3
+        id: cache
         with:
           path: |
             .venv
             .tox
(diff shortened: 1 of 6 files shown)


.github/workflows/run_system_tests.yml
```diff
@@ -22,7 +22,16 @@ jobs:
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
+      - name: Cache virtualenvs
```
bkeryan, line 25: Caching isn't working on measlinkbot:
```
Post job cleanup.
"C:\Program Files\Git\usr\bin\tar.exe" --posix -cf cache.tgz --exclude cache.tgz -P -C C:/actions-runner/[1](https://github.com/ni/measurementlink-python/actions/runs/6675067827/job/18142674125#step:11:1)/_work/measurementlink-python/measurementlink-python --files-from manifest.txt --force-local -z
/bin/sh: line 1: gzip: command not found
/usr/bin/tar: cache.tgz: Cannot write: Broken pipe
/usr/bin/tar: Child returned status 1[2](https://github.com/ni/measurementlink-python/actions/runs/6675067827/job/18142674125#step:11:2)7
/usr/bin/tar: Error is not recoverable: exiting now
Warning: Failed to save: "C:\Program failed with error: The process 'C:\Program Files\Git\usr\bin\tar.exe' failed with exit code 2
```

bkeryan: Fixed by changing measlinkbot to include the UNIX tools when installing Git and also install zstandard.

dixonjoel (approved):

PR #491, opened 2023-10-30
examples: Update NI-DCPower readme to describe the .env simulation process

Avinash2Suresh:

- Updated the `NI-DCPower` example's ReadMe file to describe the simulation process.
- Moved `.env.simulation` and `.env.sample` to `examples` directory.

Explains the process of enabling the simulation in the ReadMe file.

### DIFF
--- a/examples/nidcpower_source_dc_voltage/README.md
+++ b/examples/nidcpower_source_dc_voltage/README.md
@@ -27,12 +27,20 @@ This is a MeasurementLink example that sources and measures a DC voltage with an
 
 - NI-DCPower
 
+By default, this example uses a physical instrument. To use a simulated instrument, follow the steps below:
+- Make sure to create a `.env` file in the current directory of the measurement file or any of its parent directories.
+- Add the following environment variables to the `.env` file to set up the simulation:
+
+  ```
+  MEASUREMENTLINK_NIDCPOWER_SIMULATE=1 
+  MEASUREMENTLINK_NIDCPOWER_BOARD_TYPE=PXIe
+  MEASUREMENTLINK_NIDCPOWER_MODEL=4141
+  ```
+
 ### Required Hardware
 
 This example requires an NI SMU that is supported by NI-DCPower (e.g. PXIe-4141).
 
-By default, this example uses a simulated instrument. To use a physical instrument, edit
-`_constants.py` to specify `USE_SIMULATION = False`.
 
 > **Note**
 > The multi-site pin map, `NIDCPowerSourceDCVoltageMultiSite.pinmap`, requires an
--- a/examples/nidcpower_source_dc_voltage/README.md
+++ b/examples/nidcpower_source_dc_voltage/README.md
@@ -27,20 +27,19 @@ This is a MeasurementLink example that sources and measures a DC voltage with an
 
 - NI-DCPower
 
-By default, this example uses a physical instrument. To use a simulated instrument, follow the steps below:
-- Make sure to create a `.env` file in the current directory of the measurement file or any of its parent directories.
-- Add the following environment variables to the `.env` file to set up the simulation:
+### Required Hardware
+
+This example requires an NI SMU that is supported by NI-DCPower (e.g. PXIe-4141).
+
+By default, this example uses a physical instrument or a simulated device created in NI MAX. To automatically create a simulated device when running the measurement or TestStand sequence, follow the steps below:
+- Create a `.env` file in the measurement service's directory or one of its parent directories (such as the root of your Git repository or `C:\ProgramData\National Instruments\MeasurementLink\Services` for statically registered measurement services).
+- Add the following options to the `.env` file to enable simulation via the driver's option string:
 
   ```
   MEASUREMENTLINK_NIDCPOWER_SIMULATE=1 
   MEASUREMENTLINK_NIDCPOWER_BOARD_TYPE=PXIe
   MEASUREMENTLINK_NIDCPOWER_MODEL=4141
   ```
-
-### Required Hardware
-
-This example requires an NI SMU that is supported by NI-DCPower (e.g. PXIe-4141).
-
 
 > **Note**
 > The multi-site pin map, `NIDCPowerSourceDCVoltageMultiSite.pinmap`, requires an
--- a/examples/.env.sample
+++ b/examples/.env.sample
@@ -2,8 +2,8 @@
 #
 # To use it:
 # - Copy this file to your service's directory or one of its parent directories
-#   (such as the root of your Git repository or `C:\ProgramData\National
-#   Instruments\MeasurementLink\Services`).
+#   (such as the root of your Git repository or `C:\ProgramData\National 
+#   Instruments\MeasurementLink\Services` for statically registered measurement services).
 # - Rename it to `.env`.
 # - Uncomment and edit the options you want to change.
 # - Restart any affected services.
--- a/examples/.env.simulation
+++ b/examples/.env.simulation
@@ -4,8 +4,8 @@
 #
 # To use it:
 # - Copy this file to your service's directory or one of its parent directories
-#   (such as the root of your Git repository or `C:\ProgramData\National
-#   Instruments\MeasurementLink\Services`).
+#   (such as the root of your Git repository or `C:\ProgramData\National 
+#   Instruments\MeasurementLink\Services` for statically registered measurement services).
 # - Rename it to `.env`.
 # - Uncomment and edit the options you want to change.
 # - Restart any affected services.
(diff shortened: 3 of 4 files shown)


Avinash2Suresh:
Once this PR is approved, I'll raise a PR for all the other examples' readme.

examples/nidcpower_source_dc_voltage/README.md
```diff
@@ -27,12 +27,20 @@ This is a MeasurementLink example that sources and measures a DC voltage with an
 
 - NI-DCPower
 
+By default, this example uses a physical instrument. To use a simulated instrument, follow the steps below:
+- Make sure to create a `.env` file in the current directory of the measurement file or any of its parent directories.
+- Add the following environment variables to the `.env` file to set up the simulation:
```
bkeryan, line 32:
```suggestion
By default, this example uses a physical instrument. To use a simulated instrument, do the following:
- Create a `.env` file in the measurement service's directory or one of its parent directories (such as the root of your Git repository or `C:\ProgramData\National Instruments\MeasurementLink\Services`).
- Add the following options to the `.env` file to enable simulation via the driver's option string:
```

bkeryan (approved):
Approved with suggestions

examples/nidcpower_source_dc_voltage/README.md
dixonjoel: To make this even easier, why don't we just include a '.env' file in the root directory of each example with our preferred defaults and then tell the user, e.g. to just change
  MEASUREMENTLINK_NIDCPOWER_SIMULATE=0 to 1

and optionally change these settings if they want a different board type or model.
  MEASUREMENTLINK_NIDCPOWER_BOARD_TYPE=PXIe
  MEASUREMENTLINK_NIDCPOWER_MODEL=4141

bkeryan: `.env` is owned by the user and is in our `.gitignore` file.

I want to encourage users to share a `.env` file between multiple services.

How about this?
- Move `.env.simulation` to the `examples/` directory.
- Add `> **Note** To enable simulation for all of the examples, copy `.env.simulation` to `.env` in the examples directory.`

dixonjoel: A related question I haven't been able to find an answer for: If you have multiple .env files in the hierarchy, what is the precedent of the variables in them? E.g. if I have a .env in my example directory and another one in the repo root, both with values for MEASUREMENTLINK_NIDCPOWER_SIMULATE, which one takes precedence?

bkeryan: BTW, enabling simulation with `.env` at the root of this Git repo makes some of the tests fail, so it may be better to put `.env.sample` in the `examples/` directory too. My only concern is that the `examples/` directory makes it less visible than keeping it in the root of the Git repo.

bkeryan: > A related question I haven't been able to find an answer for: If you have multiple .env files in the hierarchy, what is the precedent of the variables in them? E.g. if I have a .env in my example directory and another one in the repo root, both with values for MEASUREMENTLINK_NIDCPOWER_SIMULATE, which one takes precedence?

Only the first `.env` file is used. There is no merging of `.env` files.

Avinash2Suresh:
> Create a `.env` file in the measurement service's directory or one of its parent directories (such as the root of your Git repository or **`C:\ProgramData\National Instruments\MeasurementLink\Services`**).

When I place the `.env` file in the Git repository, everything works as expected. However, when I try to do the same by putting the `.env` file in the `C:\ProgramData\National Instruments\MeasurementLink\Services` folder, I encounter an error that says, "Device was not recognized." Is there anything I might be overlooking in this situation?

CC: @bkeryan

examples/nidcpower_source_dc_voltage/README.md
Avinash2Suresh: > BTW, enabling simulation with .env at the root of this Git repo makes some of the tests fail, so it may be better to put .env.sample in the examples/ directory too.

In case of test case failures, I would also suggest moving the `.env.sample` file to the `examples/` directory. Simultaneously, we can include the below note within the README file located in the examples directory to provide clear instructions:

> **Note** To enable simulation for all of the modular instrument and VISA examples, copy `.env.simulation` to `.env` in the examples directory.

This way, users can easily grasp how to enable simulation and consolidate the configuration into a single .env file, as the instructions are localized within the examples folder for clarity.

Avinash2Suresh:
@bkeryan If we relocate the `.env.simulation` file to the `examples/` directory, it's better to also move the `.env.sample` file to the same directory. Keeping these files in different directories may confuse. Please let me know if you have different thoughts.

dixonjoel:
> > Create a `.env` file in the measurement service's directory or one of its parent directories (such as the root of your Git repository or **`C:\ProgramData\National Instruments\MeasurementLink\Services`**).
> 
> When I place the `.env` file in the Git repository, everything works as expected. However, when I try to do the same by putting the `.env` file in the `C:\ProgramData\National Instruments\MeasurementLink\Services` folder, I encounter an error that says, "Device was not recognized." Is there anything I might be overlooking in this situation?
> 
> CC: @bkeryan

'Device was not recognized' sounds like it's not picking up your .env file at all and trying to use hardware from NI MAX? Can you describe your process? E.g.
1. Copy the DCPower folder to c:\programdata...
2. Cd over to that directory in powershell or VS Code terminal
3. Run poetry install
4. Run poetry run python .\measurement.py

My understanding is it would use the current working directory to start looking for the .env file. If, e.g. your working directory is c:\dev\measurementlink-python and then you run, e.g. poetry run python c:\programdata...\measurement.py, it may not find the .env.

bkeryan:
> > Create a `.env` file in the measurement service's directory or one of its parent directories (such as the root of your Git repository or **`C:\ProgramData\National Instruments\MeasurementLink\Services`**).
> 
> When I place the `.env` file in the Git repository, everything works as expected. However, when I try to do the same by putting the `.env` file in the `C:\ProgramData\National Instruments\MeasurementLink\Services` folder, I encounter an error that says, "Device was not recognized." Is there anything I might be overlooking in this situation?
> 
> CC: @bkeryan

Use `C:\ProgramData\National Instruments\MeasurementLink\Services` when your measurement service is in `C:\ProgramData\National Instruments\MeasurementLink\Services`.

How about adding the text "for statically registered measurement services"?

> (such as the root of your Git repository or `C:\ProgramData\National Instruments\MeasurementLink\Services` for statically registered measurement services)

examples/nidcpower_source_dc_voltage/README.md
bkeryan: Sounds good. Adding a note to the README resolves my concern about discoverability.

examples/nidcpower_source_dc_voltage/README.md
```diff
@@ -27,12 +27,20 @@ This is a MeasurementLink example that sources and measures a DC voltage with an
 
 - NI-DCPower
 
+By default, this example uses a physical instrument. To use a simulated instrument, follow the steps below:
```
bkeryan, line 30: Move this content back to "Required Hardware". Don't move it to "Required Driver Software".

bkeryan, line 30: Let's clarify the 1st line:
```suggestion
By default, this example uses a physical instrument or a simulated device created in NI MAX. To automatically create a simulated device when running the measurement or TestStand sequence, follow the steps below:
```

Avinash2Suresh:
> > > Create a `.env` file in the measurement service's directory or one of its parent directories (such as the root of your Git repository or **`C:\ProgramData\National Instruments\MeasurementLink\Services`**).
> > 
> > 
> > When I place the `.env` file in the Git repository, everything works as expected. However, when I try to do the same by putting the `.env` file in the `C:\ProgramData\National Instruments\MeasurementLink\Services` folder, I encounter an error that says, "Device was not recognized." Is there anything I might be overlooking in this situation?
> > CC: @bkeryan
> 
> Use `C:\ProgramData\National Instruments\MeasurementLink\Services` when your measurement service is in `C:\ProgramData\National Instruments\MeasurementLink\Services`.
> 
> How about adding the text "for statically registered measurement services"?
> 
> > (such as the root of your Git repository or `C:\ProgramData\National Instruments\MeasurementLink\Services` for statically registered measurement services)

Yes, this gives much better context.

examples/nidcpower_source_dc_voltage/README.md
Avinash2Suresh: If we relocate the `.env.simulation` file to the `examples/` directory, it's better to also move the .env.sample file to the same directory. Keeping these files in different directories may confuse.

Please let me know if you have different thoughts.

Avinash2Suresh: Moved both `.env.simulation` and `.env.sample` to the `examples/` directory.

dixonjoel (approved):

vigkre (approved):

PR #499, opened 2023-11-02
[PR1] Move .env.simulation to .env in run_system_tests workflow

MounikaBattu17:

- Driver acceptance tests needs `.env` file in **root directory** to simulate driver instruments, so moved `examples/.env.simulation` to `.env` in pipeline such that it runs acceptance tests without any errors.
- The configuration in `.env` overrides the session parameter default values in unit tests that doesn't use simulation mode, which fails the tests. So to resolve this issue, passed in the default values when calling initialize session in unit tests.

To run acceptance tests that uses simulated instrument drivers in pipeline.

Manually ran `poetry run pytest`.

### DIFF
--- a/.github/workflows/run_system_tests.yml
+++ b/.github/workflows/run_system_tests.yml
@@ -22,6 +22,8 @@ jobs:
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
+      - name: Rename .env.simulation to .env
+        run: mv .env.simulation .env
       - name: Cache virtualenvs
         uses: actions/cache@v3
         id: cache
--- a/.github/workflows/run_system_tests.yml
+++ b/.github/workflows/run_system_tests.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Check out repo
         uses: actions/checkout@v4
       - name: Rename .env.simulation to .env
-        run: mv .env.simulation .env
+        run: mv examples/.env.simulation .env
       - name: Cache virtualenvs
         uses: actions/cache@v3
         id: cache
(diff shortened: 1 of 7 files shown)


keer1111 (approved):

tests/unit/_drivers/test_nidcpower.py
```diff
@@ -44,7 +44,7 @@ def test___single_session_info___initialize_nidcpower_session___session_created(
     session = create_mock_nidcpower_session()
     session_new.side_effect = [session]
 
-    with reservation.initialize_nidcpower_session() as session_info:
+    with reservation.initialize_nidcpower_session(options={}) as session_info:
```
MounikaBattu17, line 47: The default values are overridden in `initialize_nidcpower_session` since we are using `.env`, so passing default values explicitly here will ensure `.env` values are not used.

vigkre (approved):

MounikaBattu17:
Now, we see that having `.env` file in root folder doesn't affect the tests, there is a suggestion [PR](https://github.com/ni/measurementlink-python/pull/500) to move `.env.simulation` & `.env.sample` to root foldet itself. Please provide your comments on this.

cc: @bkeryan @dixonjoel

dixonjoel (approved):

.github/workflows/run_system_tests.yml
```diff
@@ -22,6 +22,8 @@ jobs:
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
+      - name: Rename .env.simulation to .env
+        run: mv examples/.env.simulation .env
```
bkeryan, line 26: Moving this file dirties the Git checkout. You should copy it instead.

bkeryan (approved):
I guess this is ok. You should copy the .env file instead of moving it, though.

.github/workflows/run_system_tests.yml
MounikaBattu17: Updated `mv` to `cp` to copy files instead of moving.

PR #500, opened 2023-11-02
[PR2] Move .env.simulation and .env.sample to root directory

MounikaBattu17:

This is a suggestion PR for this [comment](https://github.com/ni/measurementlink-python/pull/491#discussion_r1376427538). 
- Acceptance tests needs `.env` file to be present in root directory to run tests in simulation mode.
- This movement makes some of unit tests to fail, this issue has been addressed in [PR](https://github.com/ni/measurementlink-python/pull/499#discussion_r1379860260).
- Hence, moving the `.env` file from examples folder back to root folder doesn't fail any tests and also makes it more discoverable to the user.
- With that note, I have update `run_system_tests `workflow and `examples/README.md`.

Makes `.env` file more discoverable to the user.

Manually ran `poetry ran pytest` & ensured all tests passed in simulation.


dixonjoel:
This seems ok to me. It satisfies the discoverability concerns and also fixes the tests. We'll see what @bkeryan says. I'll wait to review until it's targetted to main since I think some of the files will get reverted.

bkeryan:
Moving the file to the root of the Git repo only makes it more discoverable for users who have checked out the Git repo. However, we are recommending that users download the example zip file instead:
https://github.com/ni/measurementlink-python/releases/download/1.3.0-dev0/measurementlink-python-examples-1.3.0-dev0.zip

Let's leave `.env.sample` and `.env.simulation` in the examples/ subdirectory so they continue to be included in the example zip file.

PR #503, opened 2023-11-06
tests: Add integration and acceptance tests for NI-Digital and NI-Switch drivers session management APIs

Jayaseelan-James:

Includes `nidigitial_measurement.py` and `niswitch_measurement.py` test service modules and adds integration and acceptance tests for nidigital and niswitch driver-specific session management APIs.

To add "nidigital" and "niswitch" driver-specific tests to the repository. [#AB2570088](https://dev.azure.com/ni/DevCentral/_workitems/edit/2570088)

Ran mypy and pytest.


vigkre (approved):

keer1111 (approved):

tests/acceptance/test_niswitch_measurement.py
```diff
@@ -0,0 +1,96 @@
...
+)
+from tests.utilities import niswitch_measurement
+from tests.utilities.pin_map_client import PinMapClient
+
+_SITE = 0
+
+
+def test___single_session___measure___creates_single_session(
+    pin_map_context: PinMapContext,
+    stub_v2: MeasurementServiceStub,
```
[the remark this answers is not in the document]
bkeryan: No, it's the stub_v2 fixture in conftest.py. Pytest matches up fixtures by name. Line 66 defines the measurement_service fixture not the stub_v2 fixture.

https://github.com/ni/measurementlink-python/blob/main/tests/conftest.py#L55

tests/utilities/nidigital_measurement/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+                    [session_info.resource_name],
+                    [session_info.channel_list],
+                    [connection.channel_name],
+                    passing_sites,
+                    failing_sites,
+                )
+
+
+def _burst_spi_pattern(
+    session_infos: Sequence[nims.session_management.TypedSessionInformation[nidigital.Session]],
```
bkeryan, line 68: [style] `nims.session_management.TypedSessionInformation` is overly long. Use `from ... import ...` instead.

tests/acceptance/test_nidigital_measurement.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+    ]
+
+
+def test___multiple_sessions___measure___creates_multiple_sessions(
+    pin_map_id: str,
+    stub_v2: MeasurementServiceStub,
+) -> None:
+    pin_map_context = PinMapContext(pin_map_id=pin_map_id, sites=[0, 1])
+    configurations = Configurations(pin_names=["CS"], multi_session=True)
```
bkeryan, line 58: This measurement bursts a pattern on the DUT's SPI bus pins, which uses all 4 SPI bus pins, not just CS. It doesn't make sense to reserve only CS. Please reserve all 4 pins.

This doesn't apply to the integration tests, which are testing lower-level functions and not running the measurement.

tests/utilities/nidigital_measurement/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+) -> Tuple:
+    test_assets_directory = (
+        service_directory.parent.parent / "assets" / "acceptance" / "session_management"
+    )
```
bkeryan, line 72: There is no benefit in treating the NI-Digital files (digipat, digilevels, digitiming, specs) as test assets. Move them into the service directory and you won't have to duplicate them.

tests/assets/integration/session_management/PinLevels.digilevels
bkeryan, line 1: The integration test doesn't use the NI-Digital files (digipat, digilevels, specs, digitiming). Only nidigital_measurement needs them.

bkeryan (changes requested):

tests/utilities/nidigital_measurement/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+        selected_sites = session.sites[selected_sites_string]
+        session.load_pin_map(pin_map_context.pin_map_id)
+        session.load_specifications_levels_and_timing(
+            str(_resolve_relative_path(test_assets_directory, specifications_file_path)),
+            str(_resolve_relative_path(test_assets_directory, levels_file_path)),
+            str(_resolve_relative_path(test_assets_directory, timing_file_path)),
+        )
+        session.load_pattern(
+            str(_resolve_relative_path(test_assets_directory, pattern_file_path)),
+        )
```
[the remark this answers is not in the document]
Jayaseelan-James: Yes, I did notice this `if` condition in the actual measurement, but as these were test measurements I assumed that the sessions would always be created newly so I didn't consider adding them. Now, I've wrapped them within the if case.

tests/utilities/nidigital_measurement/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+
+        levels_file_name = pathlib.Path(levels_file_path).stem
+        timing_file_name = pathlib.Path(timing_file_path).stem
+        selected_sites.apply_levels_and_timing(levels_file_name, timing_file_name)
+        selected_sites.burst_pattern(start_label="SPI_Pattern")
```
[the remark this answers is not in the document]
Jayaseelan-James: I sincerely apologise for missing this. Yes, I do remember this comment of yours in the nidcpower test measurement itself, I did add separate `for` loops in all the other test measurements, truly sorry for missing it here. :(

tests/acceptance/test_nidigital_measurement.py
Jayaseelan-James: Understood, updated the test to pass all the 4 pins.

tests/utilities/nidigital_measurement/__init__.py
Jayaseelan-James: Agreed. Moved them to the service directory and removed this tests_assets_directory variable.

tests/assets/integration/session_management/PinLevels.digilevels
Jayaseelan-James: That's right. Thank you for pointing it out. I've removed them.

tests/utilities/nidigital_measurement/__init__.py
```diff
@@ -38,12 +39,20 @@ def measure(
                 connections = reservation.get_nidigital_connections(pin_names)
                 assert all([session_info is not None for session_info in session_infos])
                 passing_sites, failing_sites = _burst_spi_pattern(session_infos)
+                site0_connections, site1_connections = [], []
+                # Assumption: All pins connected to site0 belongs to one session and
```
bkeryan, line 43: If you are trying to group the channels by session, then why not do exactly that?

Grouping the channels by site would be almost the same, but the key function would use `conn.site` instead of `conn.session_info.session_name`. However, the comment suggests you aren't really trying to group by site.

Yes, `sorted` is stable.

```python
from itertools import groupby

                key_func = lambda conn: conn.session_info.session_name
                connections_by_session = [
                    list(g) for _, g in groupby(sorted(connections, key=key_func), key=key_func)
                ]

                return (
                    ...
                    [
                        ", ".join(conn.channel_name for conn in conns)
                        for conns in connections_by_session
                    ],
                    ...
                )

File: tests/utilities/nidigital_measurement/__init__.py [hunk 8]
```diff
@@ -38,12 +39,20 @@ def measure(
                 connections = reservation.get_nidigital_connections(pin_names)
                 assert all([session_info is not None for session_info in session_infos])
                 passing_sites, failing_sites = _burst_spi_pattern(session_infos)
+                site0_connections, site1_connections = [], []
```
bkeryan, line 42: FYI, these are channel names, not connections, but my other suggestion removes these variables.

bkeryan (approved):
Approved with suggestion: group channels by session name instead of hacking around it.

tests/utilities/niswitch_measurement/NISwitchMeasurement.serviceconfig
```diff
@@ -0,0 +1,19 @@
+{
+  "services": [
+    {
+      "displayName": "NI-Switch Measurement (Py)",
```
bkeryan, line 4: The correct capitalization is "NI-SWITCH": https://www.ni.com/en/support/documentation/release-notes/product.ni-switch.html

tests/utilities/nidigital_measurement/__init__.py
Jayaseelan-James: Yeah, Thank you. I've updated the construction logic as per your suggestion.

tests/utilities/nidigital_measurement/__init__.py
Jayaseelan-James: Yes, removed them.

tests/utilities/nidigital_measurement/__init__.py
bkeryan: Ok, this is fine.

> maybe they missed handling this special case.

They explicitly chose not to handle this case: https://github.com/python/mypy/blob/bc591c756a453bb6a78a31e734b1f0aa475e90e0/mypy/renaming.py#L431

```python
    If there's a reference to a variable defined in 'with' outside the
    statement, or if there's any trickiness around variable visibility
    (e.g. function definitions), we give up and won't perform renaming.
```

PR #504, opened 2023-11-06
tests: Add integration and acceptance tests for NI-DMM and NI-DAQmx drivers session management APIs 

Avinash2Suresh:

- Includes `nidmm_measurement.py` and `nidaqmx_measurement.py` test service modules and adds integration and acceptance tests for nidmm and nidaqmx driver-specific session management APIs.
- Updated the return type of `create_nidaqmx_tasks` from `nidcpower.Session` to `nidaqmx.Task` since it was previously returning the wrong type."

To add "nidmm" and "nidaqmx" driver-specific tests to the repository. [#AB2564984](https://dev.azure.com/ni/DevCentral/_workitems/edit/2564984)

Ran mypy and pytest.


tests/utilities/nidmm_measurement/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+measurement_service = nims.MeasurementService(
+    service_config_path=service_directory / "NIDMMMeasurement.serviceconfig",
+    version="0.1.0.0",
+    ui_file_paths=[
+        service_directory,
+    ],
+)
+
+
+class Function(Enum):
```
vigkre, line 22: Do we really need to create this class? I see this is only used in https://github.com/ni/measurementlink-python/pull/504/files#diff-a97403d3637222c25a83c1fb125bed1727aeb8be60e2b9c8e99846c331d748a0R95

tests/acceptance/test_nidaqmx_measurement.py
```diff
@@ -0,0 +1,107 @@
...
+
+def test___single_session___measure___returns_measured_values(
+    pin_map_context: PinMapContext,
+    stub_v2: MeasurementServiceStub,
+) -> None:
+    pin_names = ["Pin1"]
+    configurations = Configurations(pin_names=pin_names, multi_session=False)
+
+    outputs = _measure(stub_v2, pin_map_context, configurations)
+    assert outputs.voltage_values == [-0.14969328897976622, 0.019074068422498244]
```
vigkre, line 34: We can have a line break between `act` and `assert` statements

tests/integration/session_management/test_nidaqmx_reservation.py
```diff
@@ -0,0 +1,96 @@
...
+    PinMapContext,
+    SessionManagementClient,
+)
+from tests.utilities.connection_subset import ConnectionSubset, get_connection_subset
+from tests.utilities.pin_map_client import PinMapClient
+
+_SITE = 0
+
+
+def test___single_session_reserved___initialize_nidaqmx_session___creates_single_session(
```
vigkre, line 16:
```suggestion
def test___single_session_reserved___create_nidaqmx_task___creates_task(
```

tests/integration/session_management/test_nidaqmx_reservation.py
```diff
@@ -0,0 +1,96 @@
...
+        assert all([session_info.session is not None for session_info in session_infos])
+        assert all(
+            [
+                session_info.session_name == expected_resource
+                for session_info, expected_resource in zip(session_infos, nidaqmx_resource)
+            ]
+        )
+
+
+def test___session_created___get_nidaqmx_connection___returns_connection(
```
vigkre, line 54:
```suggestion
def test___task_created___get_nidaqmx_connection___returns_connection(
```

vigkre (changes requested):
Please reset my review once the requested changes are addressed

tests/utilities/nidaqmx_measurement/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def _read_voltage_values(
+    session_infos: Sequence[nims.session_management.TypedSessionInformation[nidaqmx.Task]],
+) -> Iterable[float]:
+    sample_rate = 1000.0
+    number_of_samples = 2
+
+    voltage_values = []
+    for session_info in session_infos:
```
bkeryan, line 71: As with NI-Digital, wrapping the entire multi-session measurement in one for-loop is a mistake because it prevents parallelizing across sessions. Call task.start(), task.read(), and task.stop() in separate for-loops. 

Yes, task.start() and task.stop() are new. Explicitly starting and stopping the tasks allows you to run multiple measurements in parallel.

tests/utilities/nidmm_measurement/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+                    [session_info.resource_name],
+                    [session_info.channel_list],
+                    [connection.channel_name],
+                    [list(signals_out_of_range)[0]],
+                    [list(absolute_resolutions)[0]],
+                )
+
+
+def _get_dmm_readings(
+    session_infos: Sequence[nims.session_management.TypedSessionInformation[nidmm.Session]],
```
bkeryan, line 71: From other review: `from ... import ...`

tests/utilities/nidmm_measurement/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+) -> Tuple[Iterable[bool], Iterable[float]]:
+    nidmm_function = nidmm.Function(nidmm.Function.DC_VOLTS.value)
+    range = 10.0
+    resolution_digits = 5.5
+
+    signals_out_of_range, absolute_resolutions = [], []
+    for session_info in session_infos:
+        session = session_info.session
+        session.configure_measurement_digits(nidmm_function, range, resolution_digits)
+        measured_value = session.read()
```
bkeryan, line 81: Put configure and read in separate for-loops to perform measurements closer together in time. (It won't parallelize anything, but it's a better example of multi-session DMM measurement technique.)

tests/utilities/nidmm_measurement/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+                assert session_info.session is not None
+                signals_out_of_range, absolute_resolutions = _get_dmm_readings([session_info])
+
+                return (
+                    [session_info.resource_name],
+                    [session_info.resource_name],
+                    [session_info.channel_list],
+                    [connection.channel_name],
+                    [list(signals_out_of_range)[0]],
+                    [list(absolute_resolutions)[0]],
```
bkeryan, line 66: Type hint these as Sequence or List instead of Iterable so you don't need to build a temp list.

I guess you could do that for pin_names too...

tests/assets/acceptance/session_management/1Daqmx2ChannelGroup2Pin1Site.pinmap
```diff
@@ -0,0 +1,22 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<PinMap xmlns="http://www.ni.com/TestStand/SemiconductorModule/PinMap.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" schemaVersion="1.1">
+	<Instruments>
+		<NIDAQmxTask name="Dev1a" taskType="AnalogInput" channelList="Dev1/ai0, Dev1/ai1" />
+		<NIDAQmxTask name="Dev1b" taskType="AnalogInput" channelList="Dev1/ai2, Dev1/ai3" />
```
bkeryan, line 5: This will not work because X Series MIOs only support running one AI task at a time. If you want multiple AI tasks with X Series, you need multiple devices (e.g. Dev1 and Dev2). 

It only *seemed* to work because you were running the tasks one at a time.

This will require updating the simulated device configuration on the self-hosted GitHub runner. See the measlinkbot verison of nisimdev_simulate.sls in ni-central.

tests/assets/acceptance/session_management/1Daqmx2ChannelGroup2Pin1Site.pinmap
bkeryan, line 1: Rename `1Daqmx` to `1Mio` to indicate the type of device (similar to how we used `1Smu` instead of `1Dcpower`)

MIO = multi-function I/O

tests/acceptance/test_nidmm_measurement.py
```diff
@@ -0,0 +1,111 @@
...
+    pin_map_context: PinMapContext,
+    stub_v2: MeasurementServiceStub,
+) -> None:
+    pin_names = ["Pin1"]
+    configurations = Configurations(pin_names=pin_names, multi_session=False)
+
+    outputs = _measure(stub_v2, pin_map_context, configurations)
+
+    assert outputs.signals_out_of_range == [False]
+    assert outputs.absolute_resolutions == [4.9999999999999996e-05]
```
bkeryan, line 36: pytest.approx

bkeryan (changes requested):

tests/acceptance/test_nidaqmx_measurement.py
```diff
@@ -0,0 +1,112 @@
...
+def test___single_session___measure___returns_measured_values(
+    pin_map_context: PinMapContext,
+    stub_v2: MeasurementServiceStub,
+) -> None:
+    pin_names = ["Pin1"]
+    configurations = Configurations(pin_names=pin_names, multi_session=False)
+
+    outputs = _measure(stub_v2, pin_map_context, configurations)
+
+    assert math.isclose(outputs.voltage_values[0], -0.14969328897976622, rel_tol=1e-9)
```
bkeryan, line 36:
```
================================== FAILURES ===================================
__________ test___single_session___measure___returns_measured_values __________

pin_map_context = pin_map_id: "C:\\actions-runner\\1\\_work\\measurementlink-python\\measurementlink-python\\tests\\assets\\acceptance\\session_management\\1Daqmx2ChannelGroup2Pin1Site.pinmap"
sites: 0

stub_v2 = <ni_measurementlink_service._internal.stubs.ni.measurementlink.measurement.v2.measurement_service_pb2_grpc.MeasurementServiceStub object at 0x06E78CD0>

    def test___single_session___measure___returns_measured_values(
        pin_map_context: PinMapContext,
        stub_v2: MeasurementServiceStub,
    ) -> None:
        pin_names = ["Pin1"]
        configurations = Configurations(pin_names=pin_names, multi_session=False)
    
        outputs = _measure(stub_v2, pin_map_context, configurations)
    
>       assert math.isclose(outputs.voltage_values[0], -0.14969328897976622, rel_tol=1e-9)
E       assert False
E        +  where False = <built-in function isclose>(0.019074068422498244, -0.14969328897976622, rel_tol=1e-09)
E        +    where <built-in function isclose> = math.isclose

tests\acceptance\test_nidaqmx_measurement.py:36: AssertionError
```

This is far too specific. Simulated DAQmx devices return a sine wave with random noise. Real X Series devices like the one in my test machine can return any value between -10.5 and +10.5.

tests/utilities/nidaqmx_measurement/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+        if not session_info.session_exists:
+            task.ai_channels.add_ai_voltage_chan(session_info.channel_list)
+
+        task.timing.cfg_samp_clk_timing(
+            rate=sample_rate,
+            samps_per_chan=number_of_samples,
+        )
+
+    for session_info in session_infos:
+        session_info.session.start()
```
dixonjoel, line 85: Any reason we can't just attach this .start as the last line of the previous for loop? Similar for the .stop.

dixonjoel (approved):
Approved with suggestions.

tests/utilities/nidaqmx_measurement/__init__.py
bkeryan: It's better to perform [task state transitions](https://www.ni.com/docs/en-US/bundle/ni-daqmx/page/mxcncpts/taskstatemodel.html) at the same time for all tasks.

This measurement doesn't use a start trigger, so putting task creation, configuration, and start in the same loop would maximize the amount of time skew between tasks. Putting start in a separate loop minimizes the amount of time skew.

If this measurement used a start trigger, it would be import to start the trigger receiver tasks before starting the trigger sender task. Also, you might need to configure the trigger sender task first, then query the trigger source terminal, before configuring the trigger receiver tasks.

Performing task state transitions in separate loops also groups them in the NI I/O Trace output, which may make debugging easier.

tests/utilities/nidaqmx_measurement/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+
+    for session_info in session_infos:
+        task = session_info.session
+        timeout = min(measurement_service.context.time_remaining, 10.0)
+        voltage_value = task.read(number_of_samples, timeout)
```
bkeryan, line 90: This code is misleading. `voltage_value` sounds like a single floating-point value, but it's actually a list of N values, where N = number_of_samples = 2.
```
[0.26760104780524063, 0.26760104780524063]
```
As a result, the single-session test cases return 2 values (both from Pin1) and the multi-session test cases return 4 values (2 from Pin1, 2 from Pin2).

If you want one sample, then read 1 sample and use append() instead of extend(). You can pass 2 samples to cfg_samp_clk_timing() and pass 1 sample to read().

If you are intentionally trying to return 2+2 samples in a single flat list, then change the variable names to be less misleading.

bkeryan (changes requested):

ni_measurementlink_service/session_management/_reservation.py
bkeryan, line 1: The PR title uses "NI-Daqmx". The correct capitalization is "NI-DAQmx". If you used the wrong capitalization anywhere else, please fix it there too.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -622,7 +622,7 @@ def create_nidaqmx_task(
     def create_nidaqmx_tasks(
         self,
         initialization_behavior: SessionInitializationBehavior = SessionInitializationBehavior.AUTO,
-    ) -> ContextManager[Sequence[TypedSessionInformation[nidcpower.Session]]]:
```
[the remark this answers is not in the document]
Avinash2Suresh: Yes, when creating a multi-session measurement, I caught this issue and fixed it.

vigkre (approved):

tests/utilities/nidaqmx_measurement/__init__.py
Jayaseelan-James: Updated it to get a single sample and appended it to the list.

tests/acceptance/test_nidmm_measurement.py
```diff
@@ -0,0 +1,111 @@
...
+    pin_map_context: PinMapContext,
+    stub_v2: MeasurementServiceStub,
+) -> None:
+    pin_names = ["Pin1"]
+    configurations = Configurations(pin_names=pin_names, multi_session=False)
+
+    outputs = _measure(stub_v2, pin_map_context, configurations)
+
+    assert outputs.signals_out_of_range == [False]
+    assert outputs.absolute_resolutions == pytest.approx([4.9999999999999996e-05])
```
bkeryan, line 36:
```suggestion
    assert outputs.absolute_resolutions == pytest.approx([5.0e-05])
```

bkeryan (approved):
Approved with one more minor suggestion

Issue #534, opened 2023-12-04
Remove the UI Progress Updates example


pbirkhol-ni:
# Bug Report

The Game of Life example shows how to do continuous updates. There is no need to have it shown in two examples. Some things to consider:
- The Game of Life doesn't do cancellation. Consider adding it.
- Mention in the example Readme that the Game of Life example shows how to do UI progress updates.

[AB#2594787](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2594787)

bkeryan:
Let's also remove sample_streaming_measurement. It was added to support automated tests, but now the tests have a fork of this code.

PR #536, opened 2023-12-05
Remove examples and move example feature to Game of Life

pbirkhol-ni:

#534 

- Remove the `ui_progress_updates` example since UI progress updates are now demonstrated in the `game_of_life` example.
- Remove the `sample_streaming_measurement` example. This was originally meant to be used for tests, but that never ended up happening. Now it is just a maintenance burden that doesn't add any value over the other examples.
- Add cancellation support to the `game_of_life` example. Also, fix the example to wait for Update Interval - Actual Update Time. This makes the update interval more accurate since it can take > 50msec to update the board.

Ran the `game_of_life` example on my machine.


examples/game_of_life/_helpers.py
pbirkhol-ni, line 1: The changes in this file were from black/ni-style-guide. I don't make any real changes. I'm not sure why black decided to reformat this, though.

bkeryan:
> Remove the sample_streaming_measurement example. This was originally meant to be used for tests, but that never ended up happening.

It did happen, but then we forked it so the tests are self-contained.

examples/game_of_life/_helpers.py
bkeryan: Did you run black directly or through ni-python-styleguide? The example projects don't yet configure [tool.black] line-length = 100, so running black directly produces different results than running it through ni-python-styleguide. 

The weird part is that ni-python-styleguide lint doesn't fail with either format. Perhaps flake8-black & black --check are tolerant of some formatting differences (at least when adding/removing a trailing comma).

Cc: @mshafer-NI

bkeryan (changes requested):

dixonjoel (approved):

examples/game_of_life/_helpers.py
pbirkhol-ni: I did run black directly first. Then I ran ni-style-guide fix. I will revert this file and only run ni-style-guide.

examples/game_of_life/measurement.py
```diff
@@ -32,9 +34,13 @@ def measure(
     width: int, height: int, update_interval: int, max_generations: int
 ) -> Generator[Tuple[xydata_pb2.DoubleXYData, int], None, None]:
     """Streaming measurement that returns Conway's Game of Life grid as DoubleXYData."""
+    cancellation_event = threading.Event()
+    measurement_service.context.add_cancel_callback(cancellation_event.set)
     grid = _initialize_grid_with_seeded_data(width, height)
+    update_interval_sec = update_interval / 1000
```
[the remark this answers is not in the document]
pbirkhol-ni: It's easy enough to change so I made the variable `update_interval_msec` and then rebinded the measui control.

bkeryan (approved):

PR #553, opened 2023-12-13
Remove "@" character prefix in the Update Pin Map step in all the example TestStand sequences

samchris007:

Removes the unnecessary "@" prefix in the Pin Map Path textbox for all the examples' sequence files. [AB#2603557](https://dev.azure.com/ni/DevCentral/_workitems/edit/2603557)

The "@" character will be automatically added when the user chooses a file path using the Browse button, but all the updated examples' sequence file (#538, #539, #540, #541) contains this "@" character which is redundant.

Verified all the updated examples' sequence files with TestStand 2021 SP1.


vigkre (approved):

Jayaseelan-James:
FYI,
The NI-Digital example's TestStand sequence doesn't require this refactor since it uses the existing `PinMapPath` variable within the MeasurementLink FileGlobals container.

Issue #35, opened 2022-03-24
Research alternatives to google.protobuf.internal encoder and decoder


Vasantha-Kumar-J:
Research on using the encoder and decoder alternatives from non internal locations so that our framework will not have risk when there is a breaking change with the APIs from the google.protobuf.internal.

[AB#2267388](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2267388)

bkeryan:
Can we create descriptor objects directly? No.

descriptor_test.py:
```python
from google.protobuf.descriptor import Descriptor, FieldDescriptor

int32_in = FieldDescriptor(
    name="int32_in",
    full_name="Configurations.int32_in",
    index=0,
    number=1,
    type=FieldDescriptor.TYPE_INT32,
    cpp_type=FieldDescriptor.CPPTYPE_INT32,
    label=FieldDescriptor.LABEL_OPTIONAL,
    default_value=123,
    message_type=None,
    enum_type=None,
    containing_type=None,
    is_extension=False,
    extension_scope=None,
)
configurations = Descriptor(
    name="Configurations",
    full_name="Configurations",
    filename=None,
    containing_type=None,
    fields=[int32_in],
    nested_types=[],
    enum_types=[],
    extensions=[],
)
```

```
PS D:\dev\measurementlink-python> poetry run python .\descriptor_test.py
Traceback (most recent call last):
  File "D:\dev\measurementlink-python\descriptor_test.py", line 3, in <module>
    int32_in = FieldDescriptor(
  File "D:\dev\measurementlink-python\.venv\lib\site-packages\google\protobuf\descriptor.py", line 561, in __new__
    _message.Message._CheckCalledFromGeneratedFile()
TypeError: Descriptors cannot not be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).

More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
```

Setting the environment variable allows creating descriptor objects directly, but this feature is deprecated.
```
[...]

PR #554, opened 2023-12-13
[releases/1.2] service: Fix tag deserialization for large messages

bkeryan:

Cherry-pick #552 into releases/1.2.

Description:

> Tags are i32 varints, but we were only deserializing the first byte.
> 
> Simplify some related details:
> - Decoding a varint returns the updated position, so we don't need to use encoder._TagSize()
> - We can directly construct a memory view once instead of constructing a BytesIO for each loop iteration.
> 
> > **Note**
> >
> > This reuses an internal+underscored protobuf function, `google.protobuf.internal.decoder._DecodeSignedVarint32`. This code was already using a different internal+underscored protobuf function, `google.protobuf.internal.encoder._TagSize`.
> >
> > For xydata support, we forked the unsigned `_DecodeVarint`, but I didn't do that here. 
> >
> > In the long run, I think we should generate descriptors so that we can stop using protobuf internals and let protobuf do the heavy lifting: #35 

Fixes [[AB#2602618](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2602618): Protobuf Error on Measurement run [MeasurementLink UI Editor/InstrumentStudio]](https://dev.azure.com/ni/DevCentral/_workitems/edit/2602618)

Release the fix in version 1.2.1.

Ran `poetry run pytest -v` in both releases/1.2 and main branches.
Manually tested the service attached to the bug report with my `users/bkeryan/fix-field-id-1.2` branch (targeting `releases/1.2`).


dixonjoel:
Are we also going to add a 1.2.0.1 bug fix release?

dixonjoel (approved):

bkeryan:
> Are we also going to add a 1.2.0.1 bug fix release?

Released Python packages have 3 version number fields to work with, so 1.2.1 would be the bug fix release. You can also do 1.2.0-post2 but that's typically for non-code changes such as packaging problems.

We can build a 1.2.1-dev0 for testing beforehand, though.

PR #575, opened 2023-12-29
fix: use os.path to avoid os errors for invalid file paths

Jayaseelan-James:

Use [os.path.exists(path)](https://docs.python.org/3/library/os.path.html#os.path.exists) instead of `Path.exists()` to swallow OS errors for invalid file paths for older Python versions. [AB#2606153](https://dev.azure.com/ni/DevCentral/_workitems/edit/2606153)

When the `.env` file isn't present in any of the measurement.py parent directories and if TestStand adaptors use any Python version <3.8.10 or <3.9.5, an OSError is thrown for invalid file paths. Using `os.path.exists(path)` resolves this error.

Ran mypy and manually ran the "nidcpower_source_dc_voltage" example's sequence file in TestStand 2021 SP1 with Python versions 3.8.9, 3.8.10, 3.9.4, 3.9.5, 3.9.13 and 3.10.11.

### DIFF
--- a/ni_measurementlink_service/_dotenvpath.py
+++ b/ni_measurementlink_service/_dotenvpath.py
@@ -1,4 +1,5 @@
 import inspect
+import os
 import sys
 import traceback
 from pathlib import Path, PurePath
@@ -44,7 +45,7 @@ def _get_caller_path() -> Optional[Path]:
     for frame, _ in traceback.walk_stack(inspect.currentframe()):
         if frame.f_code.co_filename:
             module_path = Path(frame.f_code.co_filename)
-            if module_path.exists() and not _is_relative_to(module_path, nims_path):
+            if os.path.exists(module_path) and not _is_relative_to(module_path, nims_path):
                 return module_path
 
     return None
--- a/ni_measurementlink_service/_dotenvpath.py
+++ b/ni_measurementlink_service/_dotenvpath.py
@@ -49,6 +49,7 @@ def _get_caller_path() -> Optional[Path]:
 
     return None
 
+
 # Path.exists() throws OSError when the path has invalid file characters.
 # https://github.com/python/cpython/issues/79487
 if sys.version_info >= (3, 10):
--- a/ni_measurementlink_service/_dotenvpath.py
+++ b/ni_measurementlink_service/_dotenvpath.py
@@ -1,5 +1,4 @@
 import inspect
-import os
 import sys
 import traceback
 from pathlib import Path, PurePath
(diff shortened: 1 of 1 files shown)


vigkre (approved):

ni_measurementlink_service/_dotenvpath.py
```diff
@@ -44,7 +45,7 @@ def _get_caller_path() -> Optional[Path]:
     for frame, _ in traceback.walk_stack(inspect.currentframe()):
         if frame.f_code.co_filename:
             module_path = Path(frame.f_code.co_filename)
-            if module_path.exists() and not _is_relative_to(module_path, nims_path):
+            if os.path.exists(module_path) and not _is_relative_to(module_path, nims_path):
```
mshafer-NI, line 48: It's annoying when subtle differences like this happen between Python versions.

@bkeryan, should we keep this like this? (I believe the Path object will be converted back to a string - non-optimal but not a huge hit to performance unless called many times) 
or make a helper method like we did for `_is_relative_to` that provides a try/except (although, that require a program stack operations and will also cost us a few clock cycles), 
or directly reference the str from above (probably by storing off `frame.f_code.co_filename`)?

dixonjoel (approved):

ni_measurementlink_service/_dotenvpath.py
bkeryan, line 48: Add a comment about the workaround. Mixing pathlib and os.path is messy and someone is likely to revert this workaround by accident when cleaning up the code.

Please identify which part of this is the workaround, the problem it's working around (Path.exists raises OSError for invalid paths like "<frozen importlib_.bootstrap>"), and the affected Python versions (<3.8.10, <3.9.5). A link to the bug is also helpful.

Some of this may be better to document in code.
- Adding a helper function would clearly identify that this is a workaround for an issue with Path.exists.
- Checking sys.version_info would clearly document the affected versions, and tools like pyupgrade may be able to automatically delete this code when we drop support for 3.8 and 3.9. (We will still have to remove the helper function manually, though.)

I have something like this in mind:
```python
if (3, 8, 0) <= sys.version_info < (3, 8, 10) or (3, 9, 0) <= sys.version_info < (3, 9, 5):
    def _exists(path: PurePath) -> bool:
        # do workaround
else:
    def _exists(path: PurePath) -> bool:
        return path.exists()
```

We can also change _is_relative_to to define the function conditionally like this.

bkeryan: I don't like mixing pathlib and os.path, so I would prefer doing one of the following:
- Rewrite this entire line to use os.path. Is there an os.path one-liner equivalent to is_relative_to?
- Add a helper method and use try/except to ignore OSError when e.errno == errno.EINVAL.

bkeryan (changes requested):

ni_measurementlink_service/_dotenvpath.py
bkeryan: I guess a simpler version check of `if sys.version_info >= (3, 10):` is also ok and will give us coverage of both code paths.

Jayaseelan-James: Thanks for the comment. I've added `_exists` method for the appropriate Python versions and also updated the `_is_relative_to` accordingly.

Jayaseelan-James: > Is there an os.path one-liner equivalent to is_relative_to?

I did a little web search but couldn't find any equivalent methods.

> Add a helper method and use try/except to ignore OSError when e.errno == errno.EINVAL.

I used the `sys.version` approach as described in the other thread since I felt it was comparatively more expressive.

ni_measurementlink_service/_dotenvpath.py
```diff
@@ -44,16 +44,35 @@ def _get_caller_path() -> Optional[Path]:
...
     return None
 
 
-def _is_relative_to(path: PurePath, other: PurePath) -> bool:
-    if sys.version_info >= (3, 9):
+# Path.exists() throws OSError when the path has invalid file characters.
+# https://github.com/python/cpython/issues/79487
+if sys.version_info >= (3, 10):
+
+    def _exists(path: Path) -> bool:
```
Jayaseelan-James, line 57: Used `Path` instead of `PurePath` because mypy throwed an error: "PurePath" has no attribute "exists".

bkeryan: Right, `exists` does need to access the filesystem :)

bkeryan (approved):

ni_measurementlink_service/_dotenvpath.py
bkeryan: Ok, the updated code is fine with me. I'm not concerned about the performance aspect because this should only be used for one-time initialization code and it's not used for Python 3.10+.

PR #595, opened 2024-01-24
nims: update reserve API's response to accommodate multiplexer session info

Jayaseelan-James:

- Update pin_map_service.proto and session_management_service.proto files to hold the multiplexer data.
- Add `MultiplexerSessionInformation` to parse the multiplexer info from grpc type to python-specific type.

To include support for multiplexers in MeasurementLink, this will be the entry point where the user will now see the multiplexer's name and route (if present) in the reservation object for the corresponding pin(s).

- Added new tests to verify the presence of multiplexer data in the reservation object.
- Ran lint and mypy locally.


tests/integration/session_management/test_reservation.py
```diff
@@ -158,12 +158,7 @@ def test___sessions_reserved___get_connections_by_instrument_type___connections_
         ]
 
 
-# AB#2546523: Fully support querying connections for shared pins
-#
-# When you reserve sessions for all sites, the session management service
-# currently only returns shared pins for the 1st site with which they are
-# associated.
-def test___sessions_reserved_with_shared_pins_all_sites___get_connections___connections_returned_for_first_matching_site(
+def test___sessions_reserved_with_shared_pins_all_sites___get_connections___returns_connections_for_all_sites(
```
Jayaseelan-James, line 161: The recent changes in the Pin map framework and follow-up updates in the session management and pin map service have enabled the reserve API to return _all_ the pins connected on an instrument's channel.

Jayaseelan-James:
I just got to know that the pipeline uses MeasurementLink v24.0.0.433-0+d433. That's the reason behind the _shared pins_ test case failure. 
@bkeryan, @dixonjoel I would like to know if I can consider updating the MeasurementLink version in `measlinkbot` or should I revert the "_shared pins_" test case update now and take that up later?

tests/unit/test_session_management.py
```diff
@@ -595,6 +595,164 @@ def f(typed_session_info: TypedSessionInformation[object]) -> None:
...
+    assert [info.session_name for info in reservation.multiplexer_session_info] == [
+        f"MyMultiplexer{i}" for i in range(multiplexer_session_count)
+    ]
+    assert [info.multiplexer_type_id for info in reservation.multiplexer_session_info] == [
+        f"MyMultiplexerType{i}" for i in range(multiplexer_session_count)
+    ]
+
+
+@pytest.mark.parametrize("multiplexer_session_count", [1, 2])
+def test___single_session_with_shared_pins_and__varying_multiplexer_count___reserve_session___returns_session_and_multiplexer_infos_for_all_sites(
```
SowmyaRavee, line 657: test___single_session_with_shared_pins_and_varying_multiplexer_count

tests/unit/test_session_management.py
```diff
@@ -595,6 +595,164 @@ def f(typed_session_info: TypedSessionInformation[object]) -> None:
...
+    assert [info.session_name for info in reservation.multiplexer_session_info] == [
+        f"MyMultiplexer{i}" for i in range(multiplexer_session_count)
+    ]
+    assert [info.multiplexer_type_id for info in reservation.multiplexer_session_info] == [
+        f"MyMultiplexerType{i}" for i in range(multiplexer_session_count)
+    ]
+
+
+@pytest.mark.parametrize("multiplexer_session_count", [1, 2])
+def test___multiple_sessions__with_shared_pins_and_varying_multiplexer_count___reserve_session___returns_session_and_multiplexer_infos_for_all_sites(
```
SowmyaRavee, line 685: sessions_with_shared_pins

SowmyaRavee (approved):

dixonjoel:
@Jayaseelan-James waiting for tests to pass. Other than that looks good. Ping me when the tests are passing.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -261,6 +272,13 @@ def _connection_cache(self) -> Dict[_ConnectionKey, Connection]:
                 cache[key] = value
         return cache
 
+    @property
+    def multiplexer_session_info(self) -> Sequence[MultiplexerSessionInformation]:
+        """Multiplexer session information object."""
+        if not self._multiplexer_session_info:
+            raise ValueError("No multiplexer session(s) exist.")
```
bkeryan, line 279: This error seems overly harsh. It makes it harder for the service to ask whether there is a multiplexer or not.

The corresponding `session_info` property doesn't have this error. The multi-session version returns an empty list when there are no sessions.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -190,17 +191,27 @@ def __init__(
...
+        multiplexer_session_info: Optional[
+            Sequence[session_management_service_pb2.MultiplexerSessionInformation]
+        ] = None,
         reserved_pin_or_relay_names: Union[str, Iterable[str], None] = None,
         reserved_sites: Optional[Iterable[int]] = None,
     ) -> None:
         """Initialize reservation object."""
         super().__init__(session_management_client)
-        self._grpc_session_info = session_info  # needed for unreserve
+        self._grpc_session_info = session_info  # required for unreserve
```
bkeryan, line 202: This seems like [splitting hairs](https://idioms.thefreedictionary.com/splitting+hairs) but whatever, I guess "required" is ok too.

bkeryan (changes requested):

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: Sure. Removed the check and updated the test to verify if an empty list is returned.

Jayaseelan-James:
> Go ahead and update measlinkbot to the latest prerelease version of MeasurementLink, but please do some manual testing to ensure that the other tests still pass with the latest released version of MeasurementLink. We don't want the mux changes to break compatibility with older versions of the services.

> BTW, you can temporarily mark tests as expected to fail by adding @pytest.mark.xfail(reason="Requires MeasurementLink 2024Q2 or later")

I expect a good amount of changes in ASW to support multiplexer registration in the upcoming weeks, thus to avoid consecutive version updates in `measlinkbot`, I've temporarily added the `@pytest.mark.xfail` decorator to the test. 

Also, I verified all the Python tests locally with MeasurementLink **v24.3.0.93-0+d93**. 👍🏼

dixonjoel (approved):

bkeryan (approved):

PR #605, opened 2024-01-31
nims: add get connection(s) with multiplexer APIs to accommodate mux info

Jayaseelan-James:

Adds `get_foo_connection(s)` wrapper APIs in the `BaseReservation` to accommodate the multiplexer info in the returning typed connection object.
[AB#2623289](https://dev.azure.com/ni/DevCentral/_workitems/edit/2623289)

Now with the reserve API containing the multiplexer info. This PR extends the support by adding multiplexer data to the connection object.

Included automated tests that verify the functionalities of the newly added APIs.


Jayaseelan-James:
Team, I would like to get your thoughts on the possible approaches to append multiplexer info to the connection object and how it can be retrieved.

Approach 1 [_Recommended_]: _This PR contains changes for this approach_
- In addition to the existing `get_nifoo_connection(s)`, add separate APIs to get connections that have multiplexer for each respective `get_nifoo_connection(s)` with the name `get_nifoo_connection(s)_with_multiplexer`.
- ✔Pros: 
    - Users get type hinting out of the box.
- ❌Cons:
    - Users need to pass in the multiplexer session type.
    - Users need to use separate APIs to get connections depending on the presence of a multiplexer.

ℹ️ Note: By this approach, an error will be thrown for the following cases:
- The user attempts to get a connection that has a multiplexer by passing in an _incorrect_ multiplexer session type.
- The user tries to get multiple connections and one or more connections have multiplexers of _different_ session types other than the type specified in the parameter.

Approach 2:
- Extend the TypedConnection structure so that the user can continue using the same API with the trade-off that they lose type hinting as the session info and sessions will be object type.
- ✔Pros:
    - Users can use the same API to get connections regardless of the multiplexer's presence.
    - Users need not pass multiplexer session type as a parameter to the API.
- ❌Cons: 
    - Users don't get type hinting out of the box. It would be their responsibility to typecast the objects and use them accordingly.

Approach 3 [_Not Recommended_]:
- Add type-specific properties in the Connection object. By this, a separate property will be added for each multiplexer type in the Connection object.
- ✔Pros:
    - Users can use the same API to get connections regardless of the multiplexer's presence.
    - Users get type hinting out of the box.
- ❌Cons:
    - Users must remember and use different properties for different multiplexer types.
    - This approach is not scalable as a property must be added each time when we support a new multiplexer type.

examples/nidcpower_source_dc_voltage/measurement.py
Jayaseelan-James, line 1: I've temporarily updated this example to showcase how the API could be utilized in a measurement.

ni_measurementlink_service/session_management/_types.py
```diff
@@ -287,11 +321,27 @@ class Connection(NamedTuple):
     session_info: SessionInformation
     """The instrument session information."""
 
+    multiplexer_resource_name: Optional[str]
```
bkeryan, line 324: `Optional[str]` only affects typing and does not make the constructor paraemter optional. To do that, add a default value, like `multiplexer_resource_name: Optional[str] = None`.

bkeryan (changes requested):

ni_measurementlink_service/session_management/_types.py
```diff
@@ -337,6 +387,42 @@ def session(self) -> TSession_co:
         ...
 
 
+class TypedConnectionWithMultiplexer(
+    TypedConnection, Protocol, Generic[TSession_co, TMultiplexerSession_co]
```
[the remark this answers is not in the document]
Jayaseelan-James: Thank you, I wasn't aware of that. I have updated it now.

ni_measurementlink_service/session_management/_types.py
Jayaseelan-James: Yes, I've removed "Optional" as the connection construction occurs with the `multiplexer_resource_name` and `multiplexer_route`.

I've retained `multiplexer_session_info` as `Optional[MultiplexerSessionInformation]` so that it can hold _None_.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -858,6 +870,35 @@ def get_nidcpower_connection(
...
 
+    def get_nidcpower_connection_with_multiplexer(
+        self,
+        multiplexer_session_type: Type[TMultiplexerSession],
+        pin_name: Optional[str] = None,
+        site: Optional[int] = None,
+    ) -> TypedConnectionWithMultiplexer[nidcpower.Session, TMultiplexerSession]:
+        """Get the NI-DCPower connection matching the specified criteria.
+
+        Args:
```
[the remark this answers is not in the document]
Jayaseelan-James: I've included the necessary documentation for all the newly added APIs.

ni_measurementlink_service/session_management/_types.py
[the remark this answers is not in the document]
Jayaseelan-James: I did a quick check [here](https://onlinegdb.com/XQdGEIcRw) and it looks like the Protocol accepts the Optional[str] for str and vice versa during run time. 

And we can see [here](https://mypy-play.net/?mypy=latest&python=3.8&gist=996ae51d7e2a837328ec3d1ec8047b09), mypy check fails.

tests/unit/test_reservation.py
```diff
@@ -815,6 +1189,7 @@ def _create_grpc_session_infos_with_system_pins() -> (
 
 
 _T = TypeVar("_T")
+_TMultiplexer = TypeVar("_TMultiplexer")
 
 
 class _ConnectionSubset(NamedTuple):
```
Jayaseelan-James, line 1195: Given that this PR is already overwhelmed with numerous changes, I'm planning to remove this class and instead reuse the ConnectionSubset within utilities in a separate PR.

tests/unit/test_reservation.py
Jayaseelan-James, line 1: FYI I'll also be adding a few more tests post the inclusion of the `initialize_multiplexer_session(s)` API.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -887,6 +1099,45 @@ def get_nidcpower_connections(
...
+    def get_nidcpower_connections_with_multiplexer(
+        self,
+        multiplexer_session_type: Type[TMultiplexerSession],
+        pin_names: Union[str, Iterable[str], None] = None,
+        sites: Union[int, Iterable[int], None] = None,
+    ) -> Sequence[TypedConnectionWithMultiplexer[nidcpower.Session, TMultiplexerSession]]:
+        """Get all NI-DCPower connections matching the specified criteria.
+
+        Args:
+            multiplexer_session_type: The multiplexer(s) session type.
```
SowmyaRavee, line 1111: multiplexer_session_type: The multiplexer session type.

Jayaseelan-James: I included the additional `(s)` to denote that there could be multiple multiplexers but of the same session type.

tests/unit/test_reservation.py
```diff
@@ -649,6 +657,256 @@ def test___no_reservation_order___get_connections___connections_returned_in_defa
...
+
+        assert connection.pin_or_relay_name == "Pin1"
+        assert connection.site == 2
+        assert connection.channel_name == "3"
+        assert connection.multiplexer_resource_name == "Mux0"
+        assert connection.multiplexer_route == "route1"
+        assert connection.multiplexer_session is None
+
+
+def test___no_connection___get_connection_with_multiplexer_with_multiplexer_session_type_object___value_error_raised(
```
SowmyaRavee, line 716: The test name is so confusing. Can we update?

Jayaseelan-James: How about `test___no_connection___get_connection_with_multiplexer__raises_value_error`?

SowmyaRavee (approved):

tests/unit/test_reservation.py
SowmyaRavee: I could see a similar test above with different criteria

Jayaseelan-James: Yes, that has a different precondition which is there are connections but no multiplexer session isn't created.
In this case, there's no connection at all present for querying.

SowmyaRavee: Sure then we can proceed with your suggestion

dixonjoel (approved):
Approved for RFC and approach. Please reset me if you make this a regular PR.

Jayaseelan-James:
@dixonjoel The current implementation represents the fair changes. I apologize for neglecting to update the PR title. 😞 I will address the initialize_multiplexer_sessions API additions in a separate PR.

Therefore, I'm resetting your vote. Sorry about the confusion.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -311,6 +319,18 @@ def _get_matching_session_infos(self, instrument_type_id: str) -> List[SessionIn
             info for info in self._session_info if instrument_type_id == info.instrument_type_id
         ]
 
+    def _try_get_matching_multiplexer_session_info(
```
bkeryan, line 322: We have been reserving the prefix "try_" for methods that implement the .NET Try pattern, which involves returning a bool indicating whether the operation succeeded.

In this case, I'd rather remove "try_" and call this "_get_matching_multiplexer_session_info"

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -311,6 +319,18 @@ def _get_matching_session_infos(self, instrument_type_id: str) -> List[SessionIn
...
+        self, multiplexer_resource_name: str
+    ) -> Union[MultiplexerSessionInformation, None]:
```
bkeryan, line 324:
```suggestion
    ) -> Optional[MultiplexerSessionInformation]:
```

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -448,6 +470,8 @@ def _get_connections_core(
                         session = self._session_cache.get(value.session_info.session_name)
                         value = value._with_session(session)
                         value._check_runtime_type(session_type)
+                        if multiplexer_session_type:
```
bkeryan, line 473:
```suggestion
                        if multiplexer_session_type is not None:
```

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -608,6 +674,50 @@ def get_connections(
...
+        instrument_type_id: Optional[str] = None,
+    ) -> Sequence[TypedConnectionWithMultiplexer[TSession, TMultiplexerSession]]:
+        """Get all connections matching the specified criteria.
+
+        This is a generic method that supports any instrument driver.
+
+        Args:
+            session_type: The instrument session type.
+
+            multiplexer_session_type: The multiplexer(s) session type.
```
bkeryan, line 692: "multiplexer" does not have to be plural to modify "session type".

```suggestion
            multiplexer_session_type: The multiplexer session type.
```

Replace this everywhere.

tests/unit/test_reservation.py
```diff
@@ -542,7 +550,7 @@ def test___session_not_created___get_connection_with_session_type_object___conne
     assert connection.session is None
 
 
-def test___no_connections___get_connection_with_session_type_object___value_error_raised(
+def test___no_connection___get_connection_with_session_type_object___value_error_raised(
```
bkeryan, line 553: "no connections" is fine. It doesn't have to match get_connection. It's not a big deal though.

https://englishlessonsbrighton.co.uk/followed-singular-plural-noun/

tests/unit/test_reservation.py
```diff
@@ -649,6 +657,254 @@ def test___no_reservation_order___get_connections___connections_returned_in_defa
...
+            _ = reservation.get_connection_with_multiplexer(object, fake_driver.Session)
+
+        assert "Incorrect type for multiplexer session 'MyMultiplexer0'." in exc_info.value.args[0]
+        assert (
+            "Expected <class 'tests.utilities.fake_driver.Session'>, got <class 'NoneType'>."
+            in exc_info.value.args[0]
+        )
+
+
+def test___multiplexer_session_not_created___get_connection_with_multiplexer_using_multiplexer_session_type_as_object___returns_connection(
```
bkeryan, line 688: This is kind of wordy. I like the wording "using object type" of the plural version better.

```suggestion
def test___multiplexer_session_not_created___get_connection_with_multiplexer_using_object_type___returns_connection(
```

tests/unit/test_reservation.py
```diff
@@ -649,6 +657,254 @@ def test___no_reservation_order___get_connections___connections_returned_in_defa
...
+        _ConnectionSubset("Pin3", 0, "Dev1", "2", "Mux0", "route3"),
+        _ConnectionSubset("Pin4", 0, "Dev3", "6", "Mux0", "route7"),
+        _ConnectionSubset("Pin1", 1, "Dev1", "3", "Mux0", "route4"),
+        _ConnectionSubset("Pin2", 1, "Dev2", "4", "Mux0", "route5"),
+        _ConnectionSubset("Pin3", 1, "Dev2", "5", "Mux0", "route6"),
+        _ConnectionSubset("Pin4", 1, "Dev3", "7", "Mux0", "route8"),
+    ]
+
+
+def test____partial_multiplexed_connections___get_connections_with_multiplexer_using_session_type___returns_all_specified_pin_connections(
```
bkeryan, line 835:
```suggestion
def test___partial_multiplexed_connections___get_connections_with_multiplexer_using_session_type___returns_all_specified_pin_connections(
```

tests/unit/test_reservation.py
```diff
@@ -649,6 +657,254 @@ def test___no_reservation_order___get_connections___connections_returned_in_defa
...
+        channel="1",
+        multiplexer_resource_name="Mux0",
+        multiplexer_route="route2",
+    )
+    grpc_session_infos[0].channel_mappings.add(pin_or_relay_name="Pin4", site=0, channel="3")
+    reservation = MultiSessionReservation(
+        session_management_client, grpc_session_infos, grpc_multiplexer_session_infos
+    )
+
+    # TODO: Update the multiplexer_session_type to fake_driver.Session once the
```
bkeryan, line 860: FYI: You should add a 2nd fake driver for multiplexers so that these tests aren't passing the same type for both parameters.

tests/unit/test_reservation.py
bkeryan, line 1: This is adding a lot of tests. Consider moving them to `test_reservation_multiplexer.py`, where "multiplexer" is the flavor or differentiator described in https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/20934/Regression-Test-Conventions?anchor=component-and-unit-tests

bkeryan (changes requested):

tests/unit/test_reservation.py
Jayaseelan-James: Updated.

tests/unit/test_reservation.py
Jayaseelan-James: With the decoupling of multiplexer-related tests into a separate file. Now, I've added a `ni_fake_multiplexer_driver`.

tests/unit/test_reservation.py
Jayaseelan-James: Agreed. I've moved the new tests to `.tests\\unit\test_reservation_multiplexer.py`.

tests/unit/test_reservation.py
Jayaseelan-James: After moving the new tests to `test_reservation_multiplexer.py`. I've accommodated this reusable change in this PR.

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: Thank you for the suggestion.

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: Sure.

dixonjoel (approved):

bkeryan (approved):

Jayaseelan-James:
I verified all the tests locally using Python _3.9.13-win32_. Therefore, I'm skipping the `system tests (py32)`.

PR #610, opened 2024-02-07
nims: add initialize session(s) apis for generic and niswitch specific multiplexers

Jayaseelan-James:

This PR contains changes enabling users to initialize generic and niswitch driver-specific multiplexers using the reservation object.
[AB#2623284](https://dev.azure.com/ni/DevCentral/_workitems/edit/2623284)

Following the addition of the `get_nifoo_connection(s)_with_multiplexer` APIs, this PR holds changes that expose new generic and niswitch driver-specific multiplexer session initialization APIs.

Included necessary unit and integration tests.


Jayaseelan-James:
@bkeryan @dixonjoel I've marked this as a "draft" PR as I'm yet to add _integration_ and _acceptance_ tests for the newly added APIs. While I add them, I would like to get your comments on the implementation.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -527,6 +715,42 @@ def initialize_session(
         """
         return self._initialize_session_core(session_constructor, instrument_type_id)
 
+    def initialize_multiplexer_session(
+        self,
+        session_constructor: Callable[[MultiplexerSessionInformation], TMultiplexerSession],
+        multiplexer_resource_name: Optional[str] = None,
```
Jayaseelan-James, line 721: We decided to include and have both `multiplexer_resource_name(s)` and `multiplexer_type_id` as optional parameters for the following reasons:
- Multiplexer type id is an optional user-defined identifier for multiplexers in the pin map (editor).
- The framework doesn't know about the multiplexer type. Therefore, when there are multiple multiplexers in the reservation, we rely on users to provide appropriate values to filter multiplexers for initialization.

tests/unit/test_reservation_multiplexer.py
```diff
@@ -30,6 +31,578 @@
...
+            object, fake_multiplexer_driver.Session
+        )
+
+        assert get_connection_subset_with_multiplexer(connection) == ConnectionSubset(
+            "Pin1", 2, "Dev0", "3", "Mux0", "route0"
+        )
+        assert connection.multiplexer_session_info == session_info
+
+
+def test___multiple_connections___get_connections_with_multiplexer___raises_value_error(
```
SowmyaRavee, line 570:
```suggestion
def test___multiple_connections___get_connection_with_multiplexer___raises_value_error(
```

tests/integration/session_management/test_niswitch_multiplexer_reservation .py
```diff
@@ -0,0 +1,137 @@
...
+        multiplexer_session_info = stack.enter_context(
+            reservation.initialize_niswitch_multiplexer_session()
+        )
+
+        assert multiplexer_session_info.session is not None
+        assert multiplexer_session_info.session_name == "Multiplexer2"
+
+
+@pytest.mark.xfail(reason="Requires MeasurementLink 2024Q2 or later.")
+def test___reserved_single_session_with_multiple_multiplexer___initialize_niswitch_multiplexer_sessions___creates_multiple_multiplexer_sessions(
```
SowmyaRavee, line 40:
```suggestion
def test___reserved_sessions_with_multiple_multiplexer___initialize_niswitch_multiplexer_sessions___creates_multiple_multiplexer_sessions(
```

Jayaseelan-James: Thanks for pointing it out.

SowmyaRavee (approved):

ni_measurementlink_service/session_management/_reservation.py
dixonjoel, line 1: This file is getting pretty big. Can we consider breaking it up somehow? Maybe by reservation vs multiplexer reservation?

tests/integration/session_management/test_niswitch_multiplexer_reservation .py
```diff
@@ -0,0 +1,137 @@
...
+from tests.utilities.connection_subset import (
+    ConnectionSubset,
+    get_connection_subset_with_multiplexer,
+)
+from tests.utilities.pin_map_client import PinMapClient
+
+_SITE = 0
+
+
+@pytest.mark.xfail(reason="Requires MeasurementLink 2024Q2 or later.")
```
dixonjoel, line 20: Remind me why we're marking these new tests as failing again? When will we be able to run them on the build machines?

ni_measurementlink_service/session_management/_reservation.py
dixonjoel: It wouldn't make sense to call this with both `multiplexer_resource_name` and `multiplexer_type_id` being None would it? I guess you'd always get a `ValueError` in that case?

Would it make sense to call this with `multiplexer_resource_name` being None and `multiplexer_type_id` being a valid identifier? In that case, it would give you session(s) for all multiplexers of that type?

I don't think there's any action required here. Just trying to understand the parameter usage.

dixonjoel (changes requested):

ni_measurementlink_service/_drivers/_niswitch_multiplexer.py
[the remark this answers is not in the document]
Jayaseelan-James: >  If so, would the existing Switch example still work if you updated _niswitch.py instead of adding this driver?

Yes, functionally it would work as we only access common properties. Now that I reread it, I guess if I make the `session_info` parameter to accept both _SessionInformation_ and _MultiplexerSessionInformation_ in _niswitch.py's `__call__` method, the type checking would be satisfied.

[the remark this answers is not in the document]
Jayaseelan-James: Thanks for the comment. I've discarded the newly added one and have reused the existing `_niswitch.py` :)

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -527,6 +715,42 @@ def initialize_session(
...
+            TypeError: If the argument types are incorrect.
+
+            ValueError: If no multiplexer sessions are available or
+                too many multiplexer sessions are available.
+        """
+        return self._initialize_multiplexer_session_core(
+            session_constructor, multiplexer_resource_name, multiplexer_type_id
+        )
+
     @requires_feature(SESSION_MANAGEMENT_2024Q1)
```
Jayaseelan-James, line 754: @bkeryan I would like to know if we can remove these decorators.

ni_measurementlink_service/session_management/_reservation.py
dixonjoel: Sure, I'm fine with a separate PR

dixonjoel (approved):

ni_measurementlink_service/session_management/_reservation.py
bkeryan: yes, claim https://github.com/ni/measurementlink-python/issues/516 when you do it

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -527,6 +715,42 @@ def initialize_session(
         """
         return self._initialize_session_core(session_constructor, instrument_type_id)
 
+    def initialize_multiplexer_session(
```
bkeryan, line 718: Please add a feature toggle for any new APIs that you are introducing, so that they won't be enabled until the feature is complete. This includes the APIs you have already added.

You can do this in a separate PR, but please do it before the 1st 1.4 prerelease build.

tests/integration/session_management/test_niswitch_multiplexer_reservation .py
bkeryan, line 1: Don't use spaces in Python filenames

ni_measurementlink_service/session_management/_reservation.py
bkeryan: Agreed, a separate PR is fine

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -208,6 +220,7 @@ def __init__(
             SessionInformation._from_grpc_v1(info) for info in self._grpc_session_info
         ]
         self._session_cache: Dict[str, object] = {}
+        self._multiplexer_session_cache: Dict[str, object] = {}
```
bkeryan, line 223: Ok, keeping multiplexer sessions in a separate cache is fine. That ensures that they can only be accessed with multiplexer APIs, not niswitch APIs intended for relay drivers.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -527,6 +715,42 @@ def initialize_session(
         """
         return self._initialize_session_core(session_constructor, instrument_type_id)
 
+    def initialize_multiplexer_session(
+        self,
+        session_constructor: Callable[[MultiplexerSessionInformation], TMultiplexerSession],
+        multiplexer_resource_name: Optional[str] = None,
+        multiplexer_type_id: Optional[str] = None,
```
bkeryan, line 722: Thanks for explaining why you added a multiplexer_resource_name parameter.
 
I think you should design these API to encourage using multiplexer_type_id and discourage using multiplexer_resource_name by doing both of the following:
- Swap the parameters so that multiplexer_type_id can be passed as a positional parameter.
- Make multiplexer_resource_name a keyword-only parameter so that clients that want it must ask for it by name.

bkeryan (changes requested):

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: Sure. Since this PR already holds numerous changes. I'll take that under this [work item](https://dev.azure.com/ni/DevCentral/_workitems/edit/2649584) right after closing this PR.

tests/integration/session_management/test_niswitch_multiplexer_reservation .py
Jayaseelan-James: Sorry. That seems to be typo from my end.

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: Sure. This will also enable us to achieve some level of consistency with the other existing APIs.

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: Sure, I'll track it as part of this [tech debt](https://dev.azure.com/ni/DevCentral/_workitems/edit/2649594).

ni_measurementlink_service/session_management/_reservation.py
bkeryan: On further thought, I am not convinced by this.

Pin-aware measurement services generally should not have a priori knowledge of resource names. They should use the session management and pin map services to interpret the pin map and tell them which resources to use.

Reserving pins should tell the measurement service about the multiplexers used by those pins, not all of the multiplexers in the pin map. Is that correct? 

When the pin map has two different types of multiplexer, the correct solution is to distinguish them using the type id, not to hardcode information about the resource names in the services.

The one use case I can forsee for specifying a resource name when reserving from a properly defined pin map is to set different topologies for different NI-SWITCH sessions in the TestStand ProcessSetup step. It may be reasonable for ProcessSetup to have a priori knowledge of resource names.

jonathanmendez: > Pin-aware measurement services generally should not have a priori knowledge of resource names. They should use the session management and pin map services to interpret the pin map and tell them which resources to use.

I agree with this assertion.

> Reserving pins should tell the measurement service about the multiplexers used by those pins, not all of the multiplexers in the pin map. Is that correct?

I believe this is also correct. If you do a session reservation without specifying pins, you should get the multiplexer info for all pin connections in the pin map.

> The one use case I can forsee for specifying a resource name when reserving from a properly defined pin map is to set different topologies for different NI-SWITCH sessions in the TestStand ProcessSetup step. It may be reasonable for ProcessSetup to have a priori knowledge of resource names. I don't know whether type ids should take topology into account.

I'm not familiar enough with setting up switch topologies to know if this would require a priori knowledge of the resource names. If you query for "all pins" to get all the multiplexers used in pin connections, would that be enough to do this kind of setup?

bkeryan: > I'm not familiar enough with setting up switch topologies to know if this would require a priori knowledge of the resource names. If you query for "all pins" to get all the multiplexers used in pin connections, would that be enough to do this kind of setup?

Not quite, because one call to the API creates multiple sessions.

Here's an example: you have two multiplexers using the NI-SWITCH API named PXI1Slot2 and PXI1Slot3. One should be configured for "2501/1-Wire 48x1 Mux" and the other should be configured for a different topology, like "2501/2-Wire 24x1 Mux" or "2503/1-Wire 48x1 Mux".

Your ProcessSetup code module uses `reservation.initialize_niswitch_multiplexer_sessions`, similar to https://github.com/ni/measurementlink-python/blob/main/examples/niswitch_control_relays/teststand_niswitch.py#L36 . This initializes multiple sessions in one call.

By default, the `topology` parameter is set to "Configured Topology", which means to use the topology that is configured in NI-MAX.

If you didn't want to configure the topology in NI-MAX, then you would have to specify the `topology` parameter when you call `initialize_niswitch_multiplexer_sessions`. However, one call creates two sessions, so if you want to specify multiple topologies, you must split up this call (or use the generic `initialize_multiplexer_sessions` to specify a session constructor that looks up the right topology).

This is only important if you have multiple topologies but you don't configure them in NI MAX. Perhaps we should go with [you're not gonna need it](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and encourage users to configure their topologies in NI MAX?

Jayaseelan-James: @bkeryan, I've updated the changes by removing the "_resource_name_" parameter. Let me know if anything more is required from my end.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -535,21 +496,19 @@ def _initialize_multiplexer_sessions_core(
...
         )
-
-        # If the user specified both the matching criteria, validate that matches were found.
-        self._validate_matched_multiplexer_sessions_with_resource_names(
-            multiplexer_session_infos, _to_iterable(multiplexer_resource_names)
-        )
+        if len(multiplexer_session_infos) == 0:
+            raise ValueError(
+                f"No multiplexer sessions matched type ID '{multiplexer_type_id}'. "
+                "Expected single session, got 0 sessions."
```
SowmyaRavee, line 510: `Expected single or multiple sessions, got 0 sessions.`

Jayaseelan-James: This check is redundant as this [check](https://github.com/ni/measurementlink-python/blob/7aaba2b3744655be0eded07cc5b29d1f03d285ab/ni_measurementlink_service/session_management/_reservation.py#L388-L390) will validate that case. Hence, I've removed it now. Thanks for pointing it out.

A unit [test](https://github.com/ni/measurementlink-python/blob/7aaba2b3744655be0eded07cc5b29d1f03d285ab/tests/unit/test_reservation_multiplexer.py#L171) is in place showcasing that behavior.

ni_measurementlink_service/session_management/_reservation.py
bkeryan: Looks good, thanks

bkeryan (approved):

Jayaseelan-James:
I verified all the tests locally using Python _3.9.13-win32_. Therefore, I'm skipping the system tests (py32).

PR #621, opened 2024-02-21
refactor: decouple multiplexer initialize APIs from BaseReservation to a seperate class

Jayaseelan-James:

Moves all the multiplexer-related properties and the initialize APIs to a separate class.

The multiplexer initialize APIs must be decoupled from the `BaseReservation` to enable independent multiplexer initialization in TestStand fixtures.

Updated the existing tests.

### DIFF
--- a/ni_measurementlink_service/session_management/_reservation.py
+++ b/ni_measurementlink_service/session_management/_reservation.py
@@ -199,6 +199,349 @@ class _BaseSessionContainer(abc.ABC):
         return self._session_management_client._grpc_channel_pool
 
 
+class Multiplexer(_BaseSessionContainer):
+    """Manages multiplexer session information."""
+
+    def __init__(
+        self,
+        session_management_client: SessionManagementClient,
+        multiplexer_session_info: Optional[
+            Sequence[session_management_service_pb2.MultiplexerSessionInformation]
+        ],
+    ) -> None:
+        """Initialize multiplexer object."""
+        super().__init__(session_management_client)
+        self._multiplexer_session_cache: Dict[str, object] = {}
+        self._multiplexer_session_info: Sequence[MultiplexerSessionInformation] = []
+
+        self._multiplexer_session_info = []
+        if multiplexer_session_info is not None:
+            self._multiplexer_session_info = [
+                MultiplexerSessionInformation._from_grpc_v1(info)
+                for info in multiplexer_session_info
+            ]
+
+    @cached_property
+    def _multiplexer_type_ids(self) -> AbstractSet[str]:
+        return _to_ordered_set(
+            sorted(info.multiplexer_type_id for info in self._multiplexer_session_info)
+        )
+
+    @property
+    @requires_feature(MULTIPLEXER_SUPPORT_2024Q2)
+    def multiplexer_session_info(self) -> Sequence[MultiplexerSessionInformation]:
+        """Multiplexer session information object."""
+        if not self._multiplexer_session_cache:
+            return self._multiplexer_session_info
+
+        return [
+            info._with_session(self._multiplexer_session_cache.get(info.session_name))
+            for info in self._multiplexer_session_info
+        ]
+
+    def _get_multiplexer_session_info_for_resource_name(
+        self, multiplexer_resource_name: str
+    ) -> Optional[MultiplexerSessionInformation]:
+        return next(
+            (
+                info
+                for info in self._multiplexer_session_info
+                if info.resource_name == multiplexer_resource_name
+            ),
+            None,
+        )
+
+    def _get_multiplexer_session_infos_for_type_id(
+        self, multiplexer_type_id: str
+    ) -> List[MultiplexerSessionInformation]:
+        return [
+            info
+            for info in self._multiplexer_session_info
+            if info.multiplexer_type_id == multiplexer_type_id
+        ]
+
+    def _validate_and_get_matching_multiplexer_session_infos(
+        self,
+        multiplexer_type_ids: Iterable[str],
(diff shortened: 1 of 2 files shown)


SowmyaRavee (approved):

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -199,6 +199,362 @@ def _grpc_channel_pool(self) -> GrpcChannelPool:
...
+        self,
+        session_management_client: SessionManagementClient,
+        multiplexer_session_info: Optional[
+            Sequence[session_management_service_pb2.MultiplexerSessionInformation]
+        ],
+    ) -> None:
+        """Initialize multiplexer object."""
+        super().__init__(session_management_client)
+        self._multiplexer_session_cache: Dict[str, object] = {}
+        self._multiplexer_session_info: Sequence[MultiplexerSessionInformation] = []
```
bkeryan, line 215: `Sequence[T]` is an immutable sequence, kind of like `IEnumerable<T>` in .NET. For a member variable, I would prefer using the actual type (which is `List[T]`). But see the next comment first.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -199,6 +199,362 @@ def _grpc_channel_pool(self) -> GrpcChannelPool:
...
+
+        self._multiplexer_session_info = []
+        if multiplexer_session_info is not None:
+            self._multiplexer_session_info = [
```
bkeryan, line 219: This assigns `self._multiplexer_session_info` three times.

If you rewrite it as if/else, then it will only assign the field once. Also, you probably won't need to explicitly specify a type hint anymore, because type checkers can infer the list element type from the first assignment.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -286,31 +637,24 @@ def _connection_cache(self) -> Dict[_ConnectionKey, Connection]:
...
-        return _to_ordered_set(
-            sorted(info.multiplexer_type_id for info in self._multiplexer_session_info)
-        )
+    @property
+    def _multiplexer_session_cache(self) -> Dict[str, object]:
+        return self._multiplexer_session_handler._multiplexer_session_cache
 
     @property
-    @requires_feature(MULTIPLEXER_SUPPORT_2024Q2)
     def multiplexer_session_info(self) -> Sequence[MultiplexerSessionInformation]:
```
bkeryan, line 655: Keep the feature toggles

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -199,6 +199,362 @@ def _grpc_channel_pool(self) -> GrpcChannelPool:
         return self._session_management_client._grpc_channel_pool
 
 
+class MultiplexerSessionHandler(_BaseSessionContainer):
+    """Manages multiplexer session information."""
+
+    def __init__(
```
bkeryan, line 205: Are you planning to make the `MultiplexerSessionHandler` class public? If so, please make its constructor enforce the feature toggle as well. If the feature toggle decorator doesn't work on `__init__` methods, then I guess you can call `THE_FEATURE_TOGGLE._raise_if_disabled()`. (Consider making this method public from the feature toggle class if you do.)

BTW, even if users can't construct instances of the class directly, if you return it to the user then you should make it or a corresponding ABC or Protocol public so that users can write type hints for their code.

bkeryan (changes requested):

ni_measurementlink_service/session_management/_reservation.py
bkeryan: The last part is addressed in the next PR

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -199,6 +199,362 @@ def _grpc_channel_pool(self) -> GrpcChannelPool:
         return self._session_management_client._grpc_channel_pool
 
 
+class MultiplexerSessionHandler(_BaseSessionContainer):
```
bkeryan, line 202: I'm not 100% sold on the word "handler" because it's commonly associated with event callbacks.

"accessor" or "container" would make more sense to me.

@dixonjoel What do you think?

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: Sure. I've made the `MultiplexerSessionContainer` (previously `MultiplexerSessionHandler`) class public.

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: I went ahead with `MultiplexerSessionContainer` as it sounds better. Please let me know if that's fine.

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: Sure, I've updated with an if/else.

vigkre (approved):

ni_measurementlink_service/session_management/_reservation.py
bkeryan: The feature toggle decorator also serves as documentation. Please keep it on all of the new methods.

bkeryan (approved):
Approved with suggestions: restore the feature toggles on all entry points for documentation purposes

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: Got it. I'll have them back.

PR #639, opened 2024-03-06
GetConnection(s) to Support Pin Groups

Avinash2Suresh:

- Flatten the requested pin or relay names in the `_get_connections_core` API.
- Added automated tests.

- [AB#2655409](https://dev.azure.com/ni/DevCentral/_workitems/edit/2655409): obtain pin/relay connections using pin/relay groups.

Existing and new tests passes.


ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -816,13 +816,19 @@ def _get_connections_core(
             instrument_type_id, self._reserved_instrument_type_ids
         )
 
+        requested_pin_or_relay_names = _to_ordered_set(
```
Avinash2Suresh, line 819: With this update, the `get_connections` API now ignores duplicate connections. For instance, when requesting connections for ['Pin1', 'Pin1'], the current implementation returns only one connection for Pin1. In the previous version, the get-connection(s) API would return two connections with the same Pin1 data. 

Please confirm whether there is a valid reason for returning duplicate data, or if the updated change to ignore duplicates is acceptable.

Cc: @bkeryan, @dixonjoel

tests/integration/session_management/test_reservation.py
```diff
@@ -254,6 +254,94 @@ def test___sessions_reserved_by_nested_relay_group___get_connections_by_relays__
         ]
 
 
+def test___sessions_reserved___get_connections_by_pin_group___connections_returned(
```
Jayaseelan-James, line 257: As per the Python [Regression Test Convention](https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/20934/Regression-Test-Conventions?anchor=facts-and-theories), the `Then` clause should be something like `..._returns_connections`. 
I see that there are already inconsistencies b/w `connections_returned` and `returns_connections` but I think we should prefer and move forward with the right naming approach and it was also discussed sometime earlier.

Avinash2Suresh:
As this PR has a dependency on the previous one [PR-636](https://github.com/ni/measurementlink-python/pull/636), I've raised it on top of that. I have locally tested these changes, and all the tests are passing. Kindly review this PR. I will retarget it to the main branch once the other PR is approved and merged.

Cc: @dixonjoel, @bkeryan

tests/integration/session_management/test_reservation.py
Jayaseelan-James: https://github.com/ni/measurementlink-python/pull/490#discussion_r1378807667

ni_measurementlink_service/session_management/_reservation.py
SowmyaRavee, line 819:
```suggestion
        resolved_pin_or_relay_names = _to_ordered_set(
```
Please share your thoughts

SowmyaRavee (approved):

ni_measurementlink_service/session_management/_reservation.py
Jayaseelan-James: I have shared opinions on this. In one way, we could take it like they're still pins or relays (at the lowest level) that the user has reserved. Thinking in this particular context, it is the requested pin or relay names that is resolved now. 

Maybe `requested_pin_or_relay_names` sounds good in that way.

ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -816,13 +816,19 @@ def _get_connections_core(
...
+            self._to_resolved_pin_or_relay_names(requested_pins)
```
Jayaseelan-James: `requested_pins` is only used here. Can we move this construction right below it to improve readability?

ni_measurementlink_service/session_management/_reservation.py
dixonjoel: I don't know of any valid reason.

ni_measurementlink_service/session_management/_reservation.py
Avinash2Suresh: The logics were grouped in such a way that converting the pins, sites, and instruments to an iterable is done within the same block. Moving this above might compromise readability. Therefore, I would prefer to stick with the current arrangement.

dixonjoel (approved):

bkeryan (changes requested):

tests/unit/test_reservation.py
```diff
@@ -261,6 +261,54 @@ def test___session_reserved_using_pin_group___get_connection_by_pin___returns_co
...
+
+
+def test___session_reserved___get_connections_by_pin_group___returns_connections(
+    session_management_client: Mock,
+) -> None:
+    with ExitStack() as stack:
+        grpc_session_infos = create_nifake_session_infos(1)
+        grpc_session_infos[0].channel_mappings.add(pin_or_relay_name="Pin1", site=2, channel="3")
+        grpc_session_infos[0].channel_mappings.add(pin_or_relay_name="Pin2", site=2, channel="2")
+        group_mappings: Dict[str, Iterable[str]] = {"PinGroup1": ["Pin1", "Pin2"]}
```
bkeryan, line 293: The fact that you added this type hint suggests that `Dict[str, Iterable[str]]` is wrong. Look more closely at the error message that caused you to add this.

https://mypy-play.net/?mypy=latest&python=3.12&gist=98c21f5d04cf0f7c6be3ee362f5c5462

```
import typing

x = {"a": ["b", "c"]}
reveal_type(x)

def f(x: typing.Dict[str, typing.Iterable[str]]) -> None:
    pass

f(x)
```

```
main.py:4: note: Revealed type is "builtins.dict[builtins.str, builtins.list[builtins.str]]"
main.py:9: error: Argument 1 to "f" has incompatible type "dict[str, list[str]]"; expected "dict[str, Iterable[str]]"  [arg-type]
main.py:9: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
main.py:9: note: Consider using "Mapping" instead, which is covariant in the value type
Found 1 error in 1 file (checked 1 source file)
```

Since `dict` is invariant, you can use either:
- `Mapping[str, Iterable[str]]`: works because `Mapping`'s value type is covariant
- `Dict[str, List[str]]`: works because it exactly matches the types we are using

bkeryan: BTW, mutable collections are probably invariant because otherwise they would have to be covariant for reads and contravariant for writes. Otherwise, you could upcast `Dict[str, List[str]]` to `Mapping[str, Iterable[str]]` and insert an `Iterable[str]` that isn't a `List[str]`, such as `Tuple[str, ...]`, `str`, or a generator.

ni_measurementlink_service/session_management/_reservation.py
Avinash2Suresh: We're removing the duplicates from the resolved requested pin list. Therefore, the duplicate pins will be removed before framing the connection response.

Avinash2Suresh: Updated.

bkeryan: > We're removing the duplicates from the resolved requested pin list.

This is because of pin groups, right? Because they may overlap?

Avinash2Suresh: > This is because of pin groups, right? Because they may overlap?

yes.


> I'm not asking about what you're doing. I'm asking about why you're doing it.

I apologize. As you mentioned, to avoid overlaps, we are removing duplicates.

bkeryan: ok, thanks

tests/unit/test_reservation.py
bkeryan: Delete the type hint from the test and fix the type hint in the BaseReservation class:

```
class BaseReservation(_BaseSessionContainer):
    """Manages session reservation."""
    def __init__(
        self,
        session_management_client: SessionManagementClient,
        session_info: Sequence[session_management_service_pb2.SessionInformation],
        multiplexer_session_info: Optional[
            Sequence[session_management_service_pb2.MultiplexerSessionInformation]
        ] = None,
        pin_or_relay_group_mappings: Optional[Dict[str, Iterable[str]]] = None,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        reserved_pin_or_relay_names: Union[str, Iterable[str], None] = None,
```

Avinash2Suresh: Thanks for pointing it out. I've now updated.

bkeryan (approved):

PR #643, opened 2024-03-07
fix: remove feature toggle from MultiplexerSessionContainer's init method

Jayaseelan-James:

Removes the feature toggle from the `MultiplexerSessionContainer`'s "__init__" method.
[AB#2673383](https://dev.azure.com/ni/DevCentral/_workitems/edit/2673383)

If the examples that don't use multiplexer update their NIMS dependency to `1.4.0-dev0`, a `FeatureNotSupportedError` is thrown.

- Manually verified,
  - the "ni_dcpower_source_dc_voltage" measurement _with_ and _without_ enabling the feature toggle (**no** errors were thrown).
  - the "ni_dcpower_source_dc_voltage_with_multiplexer" _with_ and _without_ enabling the feature toggle and the `FeatureNotSupportedError` was thrown when the feature toggle wasn't enabled. 

- Existing tests should still pass.

### DIFF
--- a/ni_measurementlink_service/session_management/_reservation.py
+++ b/ni_measurementlink_service/session_management/_reservation.py
@@ -203,7 +203,6 @@ class _BaseSessionContainer(abc.ABC):
 class MultiplexerSessionContainer(_BaseSessionContainer):
     """Manages multiplexer session information."""
 
-    @requires_feature(MULTIPLEXER_SUPPORT_2024Q2)
     def __init__(
         self,
         session_management_client: SessionManagementClient,


Jayaseelan-James:
@bkeryan I went ahead with removing the feature toggle from the `__init__` method instead of conditionally constructing the `MultiplexerSessionContainer` object because having the [_multiplexer_session_container](https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/session_management/_reservation.py#L584C14-L584C44) field has an optional entity would force us to do a None check in all of its references (currently 7). 

Since all the publicly exposed properties and methods in `MultiplexerSessionContainer` are encapsulated with the feature toggle, I hope removing it from the `__init__` doesn't incur any damage.

bkeryan (approved):

PR #663, opened 2024-03-18
Move to Poetry version 1.8.2

dixonjoel:

Poetry recently released 1.8. One feature we've wanted is the 'non-package' pyproject.toml file. This update adds the 'package-mode' to specify if a pyproject.toml is for building a package or not. See https://python-poetry.org/blog/announcing-poetry-1.8.0/

Update the poetry version so we can build non-package projects.

Relying on automated tests.

### DIFF
--- a/examples/game_of_life/pyproject.toml
+++ b/examples/game_of_life/pyproject.toml
@@ -1,6 +1,7 @@
 [tool.poetry]
 name = "game_of_life"
 version = "1.0.0"
+package-mode = false
 description = "MeasurementLink example that displays Conway's Game of Life in a graph."
 authors = ["National Instruments"]
 
--- a/.github/workflows/Publish_NIMS.yml
+++ b/.github/workflows/Publish_NIMS.yml
@@ -9,7 +9,7 @@ concurrency: publish_to_pypi
 
 env:
   # Versions are also listed in PR.yml
-  POETRY_VERSION: 1.2.2
+  POETRY_VERSION: 1.8.2
   PYTHON_VERSION: 3.9
 
 jobs:
--- a/.github/workflows/check_examples.yml
+++ b/.github/workflows/check_examples.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
 
 env:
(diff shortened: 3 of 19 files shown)


examples/game_of_life/pyproject.toml
bkeryan, line 1: README.md has many references to "Managing Measurement as Python Package (Project)"

https://github.com/ni/measurementlink-python?tab=readme-ov-file#appendix-managing-measurement-as-python-package-project

bkeryan (approved):
Code/workflow changes look fine, but README.md needs more work

examples/game_of_life/pyproject.toml
dixonjoel: I think you're saying that wherever we talk about managing a measurement as a 'package', whether it is with poetry, a venv, or with a system level measurementlink-service package, we should stop referring to the measurement as a measurement package since we're not building it into a package, is that right?

So maybe just refer to it as a 'Python measurement'? E.g. `Create and Manage Python Measurement using Poetry` instead of `Create and Manage Python Measurement Package using Poetry`?

dixonjoel: Or `Create and Manage Python Measurement as a Poetry package`?

dixonjoel: Also, we refer to 'Python project' multiple times. Is this really a 'Poetry project'? Or is that synonymous in your mind?

bkeryan: > I guess I'm still a bit confused about our guidance for measurement services.
> 
> But is there utility in a measurement service also being a package? Doesn't it make it easier to transport / install? Would we rather guide users to set up their measurements as packages initially for easy distribution?

Talk to @nick-beer offline since his team is working on making measurements redistributable.

> Our examples are not set up to be packages and so we couldn't easily build them and distribute them as a python package. 

Right, they have the wrong layout. Redistributable Python packages should follow https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/ and not dump everything in one directory.

PR #682, opened 2024-04-04
Add timeouts while destroying sessions for the measurements when running in TestStand

samchris007:

Adds timeouts while destroying sessions for the measurements in TestStand

Resolves [User Story 2676827](https://dev.azure.com/ni/DevCentral/_workitems/edit/2676827): Add timeout as 10s when destroying sessions in the TestStand sequences

Manually verified by running and aborting the measurement steps in the example sequences.

### DIFF
--- a/examples/nidaqmx_analog_input/teststand_nidaqmx.py
+++ b/examples/nidaqmx_analog_input/teststand_nidaqmx.py
@@ -49,8 +49,8 @@ def destroy_nidaqmx_tasks() -> None:
         session_management_client = SessionManagementClient(
             discovery_client=discovery_client, grpc_channel_pool=grpc_channel_pool
         )
-        # Timeouts are added when destroying sessions to ensure that if a cancellation is performed for the measurement,
-        # session is closed within an appropriate timeframe.
+        # Timeout is added when destroying sessions to ensure that if a cancellation is performed
+        # for the measurement, session is closed within an appropriate timeframe.
         with session_management_client.reserve_all_registered_sessions(
             instrument_type_id=INSTRUMENT_TYPE_NI_DAQMX,
             timeout=60,
--- a/examples/nidaqmx_analog_input/teststand_nidaqmx.py
+++ b/examples/nidaqmx_analog_input/teststand_nidaqmx.py
@@ -53,7 +53,7 @@ def destroy_nidaqmx_tasks() -> None:
         # for the measurement, session is closed within an appropriate timeframe.
         with session_management_client.reserve_all_registered_sessions(
             instrument_type_id=INSTRUMENT_TYPE_NI_DAQMX,
-            timeout=60,
+            timeout=10,
         ) as reservation:
             if not reservation.session_info:
                 return
--- a/examples/nidaqmx_analog_input/teststand_nidaqmx.py
+++ b/examples/nidaqmx_analog_input/teststand_nidaqmx.py
@@ -49,8 +49,11 @@ def destroy_nidaqmx_tasks() -> None:
         session_management_client = SessionManagementClient(
             discovery_client=discovery_client, grpc_channel_pool=grpc_channel_pool
         )
+        # Timeouts are added when destroying sessions to ensure that if a cancellation is performed for the measurement,
+        # session is closed within an appropriate timeframe.
         with session_management_client.reserve_all_registered_sessions(
             instrument_type_id=INSTRUMENT_TYPE_NI_DAQMX,
+            timeout=60,
         ) as reservation:
             if not reservation.session_info:
                 return
--- a/examples/nidcpower_source_dc_voltage/teststand_nidcpower.py
+++ b/examples/nidcpower_source_dc_voltage/teststand_nidcpower.py
@@ -47,8 +47,11 @@ def destroy_nidcpower_sessions() -> None:
         session_management_client = SessionManagementClient(
             discovery_client=discovery_client, grpc_channel_pool=grpc_channel_pool
         )
+        # Timeout is added when destroying sessions to ensure that if a cancellation is performed
+        # for the measurement, session is closed within an appropriate timeframe.
         with session_management_client.reserve_all_registered_sessions(
             instrument_type_id=INSTRUMENT_TYPE_NI_DCPOWER,
+            timeout=10,
         ) as reservation:
             if not reservation.session_info:
                 return
--- a/examples/nidcpower_source_dc_voltage_with_multiplexer/teststand_nidcpower.py
+++ b/examples/nidcpower_source_dc_voltage_with_multiplexer/teststand_nidcpower.py
@@ -47,8 +47,11 @@ def destroy_nidcpower_sessions() -> None:
         session_management_client = SessionManagementClient(
             discovery_client=discovery_client, grpc_channel_pool=grpc_channel_pool
         )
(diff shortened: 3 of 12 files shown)


samchris007:
FYI, currently, we see that these measurements (i.e., nidaqmx_analog_input, nifgen_standard_function) are long-running. Since other measurements take considerably less time to complete the execution (where cancellations are not applicable), we have included a timeout for only those long-running measurements. Please share your thoughts on this.

examples/nidaqmx_analog_input/teststand_nidaqmx.py
dixonjoel, line 1: I know you said these measurements are 'long-running'. But is there anything fundamentally different for these or do they just tend to take a long time to cancel? Isn't the time to cancel for all the examples indeterminate?

What happens if you don't have these timeouts? Do you get errors when cancelling? And only for these measurements?

SowmyaRavee: TestStand performs Cleanup on aborting the sequence execution. Since there is some taken by the measurement to cancel, the cleanup step doesn't wait until then and starts to destroy the instrument sessions. The timeout would let the destroy session wait until the measurement releases the instrument session

Jayaseelan-James: I believe the other existing examples as we ship don't consume that much time (at least with simulation) to have users cancel it during execution.

We could still add a timeout to all the other TestStand fixtures with the comment stating the reason so that even if the user updates the example to be long-running they need not worry about adding a timeout. 

Please add your thoughts on this team.

examples/nidaqmx_analog_input/teststand_nidaqmx.py
```diff
@@ -49,8 +49,11 @@ def destroy_nidaqmx_tasks() -> None:
         session_management_client = SessionManagementClient(
             discovery_client=discovery_client, grpc_channel_pool=grpc_channel_pool
         )
+        # Timeout is added when destroying sessions to ensure that if a cancellation is performed
+        # for the measurement, session is closed within an appropriate timeframe.
         with session_management_client.reserve_all_registered_sessions(
             instrument_type_id=INSTRUMENT_TYPE_NI_DAQMX,
+            timeout=60,
```
Jayaseelan-James, line 56: Also, I would like to know if we could reduce the timeout to 30 seconds or even 10 seconds if that is sufficient. I knew that earlier we had a 60-second timeout but at least I'm unaware of the decision behind it.

examples/nidaqmx_analog_input/teststand_nidaqmx.py
dixonjoel: I would favor putting it in all the examples if there's a chance they could take long enough to hit Run / Stop before the measurement finishes. But I'll defer to @bkeryan if he has other thoughts.

dixonjoel (approved):
Approved with suggestions.

examples/nidaqmx_analog_input/teststand_nidaqmx.py
jasonmreding: I think this is a combination of the fact that cancellation over gRPC is asynchronous between the client and server, and the measurement services themselves will have implementation specific latency at which they will notice a cancellation request. Since cancellation between the client and server is async, that means the abort request for our step type in TestStand will effectively be processed immediately. Technically, there is 100 ms polling loop for abort in the step type. However, once the abort is recognized, the step type will effectively exit immediately and move onto the cleanup steps for the sequence. These cleanup steps will race with cancellation being recognized and processed from the measurement service.

Given this, I agree it would be good to apply some minimum timeout to all of our examples to avoid the cleanup steps failing due to the async nature of cancellation. I would think something on the order of magnitude of a few seconds worst case should be sufficient to overcome the base line grpc latency for the async cancellation. For example services where the service polls at some fixed rate when checking for cancellation, the timeout should be adjusted accordingly to take into account this additional latency. Glancing at the examples for DAQmx and FGen, it looks like they are registering a cancel callback with the underlying Python gRPC libraries. I have no idea how responsive the underlying libraries are to cancellation, but I would imagine it is pretty responsive. The measurement logic which responds to the cancel callback also appears to be pretty much immediate, so I would be curious what times you were observing since from my reading of the code, it looks like it should be reasonably fast.

BTW, the LV examples will have the same problem, and we should update them accordingly.

samchris007: Sure. Thanks for the input. I have updated the PR to include a timeout of 10s in all examples. I hope the timeout added when destroying sessions is enough to close the sessions within the timeframe. I will also update the LV examples accordingly.

SowmyaRavee (approved):

jasonmreding (approved):

examples/nidaqmx_analog_input/teststand_nidaqmx.py
jasonmreding: Should we consider changing the default timeout for `reserve_all_registered_sessions` to 10 instead? I think the only place we really expect this API to get used is when cleaning up sessions from within code modules. That would make it less likely for somebody to fall into this trap in the future. I have a slight preference for this approach, but I can live with what you have here if we're uncomfortable changing the behavior of the API or if others don't think it is appropriate or don't like having a different default timeout between this method and the regular reserve method.

bkeryan: We previously discussed changing the timeout on https://dev.azure.com/ni/DevCentral/_git/ASW/pullrequest/542140 , on multiple threads.

At the time, we didn't support cancellation in TestStand, so there was no need for anything other than a timeout of 0.

examples/nidaqmx_analog_input/teststand_nidaqmx.py
```diff
@@ -49,8 +49,11 @@ def destroy_nidaqmx_tasks() -> None:
         session_management_client = SessionManagementClient(
             discovery_client=discovery_client, grpc_channel_pool=grpc_channel_pool
         )
+        # Timeout is added when destroying sessions to ensure that if a cancellation is performed
+        # for the measurement, session is closed within an appropriate timeframe.
```
bkeryan, line 53: "Timeout is added" is appropriate for a PR description or changelog saying what you changed. For comments in the code, it's better to focus on the "why" of whatever you're trying to explain, and leave "what I changed" out of it.

Also, "... session is closed within an appropriate timeframe" makes it sound like the alternative is an inappropriate timeframe (e.g. a hang) but that's not true.

Proposed wording: "Use a timeout when destroying sessions in order to wait for the measurement to complete or be canceled."

bkeryan (changes requested):

examples/nidaqmx_analog_input/teststand_nidaqmx.py
samchris007: Okay. As suggested, I have moved the timeout to have 10s for `reserve_all_registered_sessions` in this pull request https://github.com/ni/measurementlink-python/pull/688

Also, I will abandon this PR as the appropriate changes are not needed.

examples/nidaqmx_analog_input/teststand_nidaqmx.py
samchris007: Sure. Updated the proposed change in this PR https://github.com/ni/measurementlink-python/pull/688

PR #691, opened 2024-04-12
Update service to use I/O resource specialization instead of Pin

jonathanmendez:

- Adds new DataType.IOResource and DataType.IOResourceArray1D enum values for use annotating Configuration or Output parameters. These data types represent a pin or I/O resource like a channel, better supporting the ability to treat pin maps as optional.
- Maps the new IOResource and IOResourceArray1D data types and the existing Pin and PinArray1D data types to the "ioresource" type specialization.
- Adds a deprecation warning if a Pin or PinArray1D Datatype is used with a Configuration or Output parameter.

[AB#2666681](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2666681)
This updates the python measurement service libraries to better indicate support for both pins and channels.

Updated existing tests to test the new data types.

### DIFF
--- a/ni_measurementlink_service/_datatypeinfo.py
+++ b/ni_measurementlink_service/_datatypeinfo.py
@@ -51,7 +51,9 @@ _DATATYPE_TO_DATATYPEINFO_LOOKUP = {
         False,
         message_type=xydata_pb2.DoubleXYData.DESCRIPTOR.full_name,
     ),
-    DataType.IOResource: DataTypeInfo(type_pb2.Field.TYPE_STRING, False, TypeSpecialization.IOResource),
+    DataType.IOResource: DataTypeInfo(
+        type_pb2.Field.TYPE_STRING, False, TypeSpecialization.IOResource
+    ),
     DataType.Int32Array1D: DataTypeInfo(type_pb2.Field.TYPE_INT32, True),
     DataType.Int64Array1D: DataTypeInfo(type_pb2.Field.TYPE_INT64, True),
     DataType.UInt32Array1D: DataTypeInfo(type_pb2.Field.TYPE_UINT32, True),
@@ -60,7 +62,9 @@ _DATATYPE_TO_DATATYPEINFO_LOOKUP = {
     DataType.DoubleArray1D: DataTypeInfo(type_pb2.Field.TYPE_DOUBLE, True),
     DataType.BooleanArray1D: DataTypeInfo(type_pb2.Field.TYPE_BOOL, True),
     DataType.StringArray1D: DataTypeInfo(type_pb2.Field.TYPE_STRING, True),
-    DataType.PinArray1D: DataTypeInfo(type_pb2.Field.TYPE_STRING, True, TypeSpecialization.IOResource),
+    DataType.PinArray1D: DataTypeInfo(
+        type_pb2.Field.TYPE_STRING, True, TypeSpecialization.IOResource
+    ),
     DataType.PathArray1D: DataTypeInfo(type_pb2.Field.TYPE_STRING, True, TypeSpecialization.Path),
     DataType.EnumArray1D: DataTypeInfo(type_pb2.Field.TYPE_ENUM, True, TypeSpecialization.Enum),
     DataType.DoubleXYDataArray1D: DataTypeInfo(
@@ -68,5 +72,7 @@ _DATATYPE_TO_DATATYPEINFO_LOOKUP = {
         True,
         message_type=xydata_pb2.DoubleXYData.DESCRIPTOR.full_name,
     ),
-    DataType.IOResourceArray1D: DataTypeInfo(type_pb2.Field.TYPE_STRING, True, TypeSpecialization.IOResource),
+    DataType.IOResourceArray1D: DataTypeInfo(
+        type_pb2.Field.TYPE_STRING, True, TypeSpecialization.IOResource
+    ),
 }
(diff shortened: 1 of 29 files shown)


ni_measurementlink_service/_datatypeinfo.py
```diff
@@ -43,14 +43,17 @@ def get_type_info(data_type: DataType) -> DataTypeInfo:
     DataType.Double: DataTypeInfo(type_pb2.Field.TYPE_DOUBLE, False),
     DataType.Boolean: DataTypeInfo(type_pb2.Field.TYPE_BOOL, False),
     DataType.String: DataTypeInfo(type_pb2.Field.TYPE_STRING, False),
-    DataType.Pin: DataTypeInfo(type_pb2.Field.TYPE_STRING, False, TypeSpecialization.Pin),
+    DataType.Pin: DataTypeInfo(type_pb2.Field.TYPE_STRING, False, TypeSpecialization.IOResource),
```
bkeryan, line 46: I think changing the type specialization string from "pin" to "ioresource" breaks compatibility between the new ni-measurementlink-python and old versions of MeasurementLink.

ni_measurementlink_service/measurement/info.py
```diff
@@ -67,7 +67,7 @@ class TypeSpecialization(enum.Enum):
     """Enum that represents the type specializations for measurement parameters."""
 
     NoType = ""
-    Pin = "pin"
+    IOResource = "ioresource"
```
bkeryan, line 70: Removing `Pin` breaks source compatibility

bkeryan (changes requested):

jonathanmendez:
Note that I did not update the examples yet, but I suppose I will

ni_measurementlink_service/measurement/info.py
jonathanmendez: The TypeSpecialization does appear as a public class, although I tried to search many places for any usage and couldn't find any (besides the internal usages here).

Assuming that it's still important to avoid, and assuming that we still want to end up with "ioresource" on the wire (your other comment), would the following be valid (I'm not familiar with this structure of a class that appears to be an enum that has string values)? Specifically that there are two values that are the same.

```python
class TypeSpecialization(enum.Enum):
    """Enum that represents the type specializations for measurement parameters."""

    NoType = ""
    Pin = "ioresource"
    Path = "path"
    Enum = "enum"
    IOResource = "ioresource"
```

ni_measurementlink_service/_datatypeinfo.py
jonathanmendez: I spoke with @jasonmreding and @DelpireNI about this. We don't necessarily have clarity about which situations with new ni-measurementlink-python and old MeasurementLink we care about. Using a new DataType in a measurement (e.g. IOResource) wouldn't work with an old MeasurementLink, for example. But it does seem easier to run into trouble changing the behavior for an existing DataType, so I think it makes sense to keep this as "pin" as you suggest.

ni_measurementlink_service/measurement/info.py
jonathanmendez: I've put Pin back

ni_measurementlink_service/_datatypeinfo.py
jonathanmendez: I've replaced the Pin type specialization and maintain the use of it when using DataType.Pin to avoid the compatibility issue.

bkeryan: Re: your deleted/edited comment: enums can have duplicates unless they're decorated with `@enum.unique`
https://docs.python.org/3/library/enum.html#enum.EnumType

tests/unit/test_service.py
```diff
@@ -89,6 +89,7 @@ def test___measurement_service___add_configuration__configuration_added(
         ("PinArrayConfiguration", DataType.PinArray1D, ["Pin1", "Pin2"], "test instrument 2"),
     ],
 )
+@pytest.mark.filterwarnings("ignore:.*Pin.*:DeprecationWarning")
```
bkeryan, line 92: Instead of filtering out warnings, I prefer to use `with pytest.deprecated_call():`. You could mention the deprecated warning in the "then" clause since this is an additional assertion.

https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=deprecated_call#pytest.deprecated_call

bkeryan (approved):
Approved with suggestions

ni_measurementlink_service/measurement/service.py
```diff
@@ -513,6 +532,9 @@ def _make_annotations_dict(
         if type_specialization == TypeSpecialization.Pin:
             if instrument_type != "" or instrument_type is not None:
                 annotations["ni/pin.instrument_type"] = instrument_type
+        if type_specialization == TypeSpecialization.IOResource:
```
jasonmreding, line 535: Note: We eventually want to add an io_type specialization and some annotations for attributes. See this [PR comment thread](https://dev.azure.com/ni/DevCentral/_git/ASW/pullRequest/681888#1711575373) for some context. That doesn't have to be done as part of this PR, but I thought I would bring it to your attention.

jasonmreding (approved):

dixonjoel (approved):

tests/unit/test_service.py
jonathanmendez: That makes sense. I've updated this call to use `with pytest.deprecated_call():`.

ni_measurementlink_service/measurement/service.py
jonathanmendez: I'd prefer to deal with this in a separate PR. I'll work with @DelpireNI to figure out how to best capture the work.

jonathanmendez: AB#2711389

PR #736, opened 2024-05-24
Updating DAQmx example to allow it to run without pin maps in TestSta…

DelpireNI:
…nd. Also updating generator.

Similar to #735, but for DAQmx. DAQmx sessions will not be reserved when an empty string is passed for the PinMapId. For this reason, we should skip the code trying to create a task and let the measurement handle it.

I also forgot to update the code gen in the last PR.

Same as #735.
We want our examples to work in the non-pin centric workflow.

Successfully executed sequence by removing the update pin map step, and specifying Dev0/ai0 for the pin parameter (Dev0 simulated in MAX).


ni_measurementlink_generator/ni_measurementlink_generator/templates/_helpers.py.mako
jasonmreding, line 1: This is sort of unrelated to this change, but is there a reason the `TestStandSupport` class isn't part of our runtime library rather than code that is duplicated/generated for every single measurement? In particular, this is going to make changing `_PIN_MAP_ID_VAR` on line 13 like we talked about today for TestStand even more difficult. We will either have to continue publishing the setting under the old and new variable name or existing code modules will have to be rewritten to work with the variable name for the newest releases of MeasurementLink.

examples/nidaqmx_analog_input/teststand_nidaqmx.py
```diff
@@ -23,6 +23,10 @@ def create_nidaqmx_tasks(sequence_context: Any) -> None:
     with GrpcChannelPool() as grpc_channel_pool:
         teststand_support = TestStandSupport(sequence_context)
         pin_map_id = teststand_support.get_active_pin_map_id()
+
+        if not pin_map_id:
```
jasonmreding, line 27: Is there not a way to change our run time library code such that the logic for the code module doesn't need to change? What errors if you didn't early return here? Or is it because we end up registering a task with no channels, and then that causes weird down stream issues with the measurement?

ni_measurementlink_generator/ni_measurementlink_generator/templates/_helpers.py.mako
sreekantham: Correct.  After the 24Q1 release where we officially made APIs available we could have made a change to move the TestStandSupport into our library.  Before then we didn't really have an official library we were shipping.

examples/nidaqmx_analog_input/teststand_nidaqmx.py
DelpireNI: I'm not sure what the implications of doing this in the framework are. This is the error we get:
```
<class 'ValueError'>
No reserved sessions matched instrument type ID 'niDAQmx'. Expected single or multiple sessions, got 0 sessions.

Call Stack information:
  File "C:\Users\vagrant\Desktop\measurementlink-python-examples-1.4.0\nidaqmx_analog_input\teststand_nidaqmx.py", line 37, in create_nidaqmx_tasks
    with reservation.create_nidaqmx_tasks(
  File "C:\Users\vagrant\AppData\Local\Programs\Python\Python39\Lib\contextlib.py", line 119, in __enter__
    return next(self.gen)
  File "C:\Users\vagrant\Desktop\measurementlink-python-examples-1.4.0\nidaqmx_analog_input\.venv\Lib\site-packages\ni_measurementlink_service\session_management\_reservation.py", line 755, in _initialize_sessions_core
    raise ValueError(

ValueError: No reserved sessions matched instrument type ID 'niDAQmx'. Expected single or multiple sessions, got 0 sessions.
```

We're calling into `_reservation.py`'s `create_nidaqmx_tasks`, which calls `_initialize_sessions_core`. That is where the error is raised. We expect our `_reservation` object to have sessions that match niDAQmx. 

Maybe we should make it possible to return an empty array of tasks? I'm not entirely sure.

jasonmreding: Looking at the LV libraries, it looks like the core reservation code allows for an empty reservation. However, the shared code for creating driver specific sessions from the reservation will throw an error if there are zero matching session infos. For Python, it appears both the reservation class and the driver specific create session APIs will throw an error if zero matching sessions are found.

I'm not entirely sure what the right answer is here, but ideally the solution wouldn't require the user to change how their code is written. Maybe we're OK with this is if it only for DAQmx, but I think it would be better if the user didn't have to concern themselves with this at all. I'm inclined to say we should:
- Loosen up the reservation classes so that zero matching sessions is allowed. Or at least it is allowed for the multiple session reservation class. Throwing an error for the single reservation class still seems reasonable.
- Either do the same for all of the driver specific create session APIs or just make a special concession for the DAQmx specific entry point. I don't know if returning an empty enumerable is problematic with any of these APIs from a Python perspective.
- Live with the overhead of calling `session_management_client.register_sessions` and making a grpc call with an empty set or add some client side logic to early return.

ni_measurementlink_generator/ni_measurementlink_generator/templates/_helpers.py.mako
DelpireNI: Marking as closed. If/when we decide to pull this into the official APIs, that should be addressed as a separate work item.

dixonjoel (approved):

PR #738, opened 2024-05-29
[RFC] Allow returning empty reservation

DelpireNI:

No longer error when `_initialize_sessions_core` is called and no session_info matches the instrument type id. Instead just return an empty array.
 
For DAQmx measurements in TestStand using the I/O Discovery Service, we will rely on the measurement to create the session/tasks. This is because without a pin map, we can't know ahead of time what to create.

One fix for this is to address it in the example code. That exists change is proposed here:  #736
It was suggested to address this in the framework instead.

We should also decide if we want the single ` _initialize_session_core` to behave the same or not.

Fixed automated test. Still need to test with sequence and I/O Discovery service.

### DIFF
--- a/ni_measurementlink_service/session_management/_reservation.py
+++ b/ni_measurementlink_service/session_management/_reservation.py
@@ -741,11 +741,6 @@ class BaseReservation(_BaseSessionContainer):
         if not instrument_type_id:
             raise ValueError("This method requires an instrument type ID.")
         session_infos = self._get_matching_session_infos(instrument_type_id)
-        if len(session_infos) == 0:
-            raise ValueError(
-                f"No reserved sessions matched instrument type ID '{instrument_type_id}'. "
-                "Expected single or multiple sessions, got 0 sessions."
-            )
 
         if closing_function is None:
             closing_function = closing_session
--- a/tests/unit/test_reservation.py
+++ b/tests/unit/test_reservation.py
@@ -180,11 +180,8 @@ def test___no_session_infos___initialize_sessions___value_error_raised(
 ) -> None:
     reservation = MultiSessionReservation(session_management_client, [])
 
-    with pytest.raises(ValueError) as exc_info:
-        with reservation.initialize_sessions(construct_session, "nifake"):
-            pass
-
-    assert "No reserved sessions matched instrument type ID 'nifake'." in exc_info.value.args[0]
+    with reservation.initialize_sessions(construct_session, "nifake") as reserved_sessions:
+        assert reserved_sessions == []
 
 
 def test___session_already_exists___initialize_sessions___runtime_error_raised(


ni_measurementlink_service/session_management/_reservation.py
```diff
@@ -741,11 +741,6 @@ def _initialize_sessions_core(
         if not instrument_type_id:
             raise ValueError("This method requires an instrument type ID.")
         session_infos = self._get_matching_session_infos(instrument_type_id)
-        if len(session_infos) == 0:
```
jasonmreding, line 744: This approach generally seems OK to me, but I think it would be good to hear from @bkeryan since I think he and Paul were the primary designers around most of this behavior. I guess maybe it comes down to whether we think more users would be negatively impacted by not throwing an error which might make root causing a logic error more difficult vs. throwing an error and having to modify code module behavior just for DAQmx when not using a pin map. We could also consider only allowing this behavior for the DAQmx driver and still throw errors for all other drivers.

I'm also softening somewhat on the idea of just modifying our code module example. There probably aren't a lot of code modules out in the wild for DAQmx that would have to duplicate these changes if they want to work with or without a pin map.

bkeryan: Is the goal to allow performing a measurement on 0 pins?

What effect does this change have on multi-session examples like 
https://github.com/ni/measurementlink-python/blob/main/examples/nidcpower_source_dc_voltage/measurement.py
?

Does the measurement iterate over 0 sessions and return empty lists as results?

jasonmreding: The original goal was to allow existing measurements and code modules to work with or without a pin map without modification or conditional forks that required having to write unique code for each case. In the case of DAQmx, we decided there wasn't any meaningful reservation data we could return from reservation in the code module without either a pin map or a list of channels. So we could either:
1. Require code modules to have conditional code that performed reservation with a list of I/O channels when not using a pin map and use the pin/channel mappings defined in the pin map when using a pin map. This would also require duplicating information in code with information in the pin map and keeping them in sync.
2. Rely on the DAQmx session management plugin returning/reserving no sessions if no channel resources are specified so the rest of the code module logic would effectively noop (assuming it is written in a similar manner to our existing code modules for DAQmx). This would mean measurements would always create/close new sessions as part of execution since presumably the I/O channels are always specified as inputs to the measurement at that point in time.
3. Modify our existing code modules to early return or noop in this situation for DAQmx. However, it is probably unlikely for somebody who isn't already working from our code module example as a template to figure this out.

Chris' original PR linked in this PR description implemented option 3. I was suggesting maybe we should try for option 2. However, I don't feel as strongly about this as I did a couple of days ago.

To my knowledge, this is unique to DAQmx. For all the other drivers we looked at, we thought we could return some meaningful reservation information using just the `instrument_type` and what hardware was present in the system even though we don't have a list of devices or channels as part of the request. Time will tell whether or not this proves to be true.

DelpireNI:
Abandoning this PR. We will just address this in the code modules.

PR #740, opened 2024-05-31
Update lock file and fix docutils

dixonjoel:

docutils messed up a release and so `poetry lock` was giving an error "Package docutils (0.21.post1) not found."

See https://github.com/python-poetry/poetry/issues/9293

For poetry lock to work, we need to avoid attempting to use 0.21.post1. So specifying the version in pyproject.toml.

Running 'poetry lock'


pyproject.toml
```diff
@@ -44,6 +44,7 @@ nidmm = {version = ">=1.4.4", extras = ["grpc"], optional = true}
 nifgen = {version = ">=1.4.4", extras = ["grpc"], optional = true}
 niscope = {version = ">=1.4.4", extras = ["grpc"], optional = true}
 niswitch = {version = ">=1.4.4", extras = ["grpc"], optional = true}
+docutils = ">=0.16, !=0.21.post1"
```
bkeryan, line 47: Putting this in `[tool.poetry.dependencies]` makes the packages we upload to PyPI depend on docutils, which is unnecessary and affects clients.
 
I don't think anything outside of `[tool.poetry.group.docs.dependencies]` should depend on docutils, so move this there. Also add a comment about the workaround, because otherwise we don't need to depend on docutils at all.

bkeryan (changes requested):

PR #741, opened 2024-06-03
Bump versions in main to 2.0.0-dev0

dixonjoel:

In preparation for renaming the packages, bump the versions to 2.0.0-dev0.

Avoid confusion with the old package versions and those in releases/1.5 branch specifically.

None


dixonjoel:
@cfuentes-ni Now that we're changing package names, I was going to bump the major version to 2.0. This would appear more like we're continuing the development of the ni_measurementlink_service package with a different name. We could also go with a version 1.0 as if this is an entirely new package.

dixonjoel:
Version changes were pushed with another change.

PR #742, opened 2024-06-03
Rename 'ni_measurementlink_service' to 'ni_measurement_plugin_sdk'

dixonjoel:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/measurement-services-python/blob/main/CONTRIBUTING.md).

### What does this Pull Request accomplish?

Renames the 'ni_measurementlink_service' to 'ni_measurement_plugin'.

### Why should this Pull Request be merged?

Remove references to 'MeasurementLink' and avoid users having to search for 'MeasurementLink' in pypi.org.

### What testing has been done?

Ran tests locally. A follow-up change will modify the examples to use this new package.

PR #751, opened 2024-06-14
Rename package to ni_measurement_plugin_sdk_service

dixonjoel:

Rename the package to ni_measurement_plugin_sdk_service. Generator will subsequently be renamed to ni_measurement_plugin_sdk_generator. This is for consistency with LabVIEW where the top-level package will be ni-measurement-plugin-sdk. We don't have a top-level package in Python (yet) but this name will be reserved for that top-level package.

See #742 for comparison.

PO requested this renaming for consistency.

Ran pytest locally.


.github/ISSUE_TEMPLATE/bug_report.md
bkeryan, line 1: @dixonjoel @cfuentes-ni @mshafer-NI 

```
>>> len("ni-measurement-plugin-sdk-service")
33
```

How did we get so far from [PEP8's recommendation of short, all-lowercase names](https://peps.python.org/pep-0008/#package-and-module-names)?

> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

dixonjoel: @rfriedma

rfriedma: for the package names we lost a proper-noun that stood for the more descriptive words we've now inserted and we're trying to mirror naming across languages with different naming guidelines. 

is 33 characters for a package name too long? Is package name synonymous with module names?

mshafer-NI: Modules are the folder containing a `__init__.py` file.

> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. 

Packages are either other .py files or the overall project / wheel.

> Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

Personal preference, all lower case, separators are underscores (although Poetry has a different view on this).

I think we're all in agreement that this is not short...

dixonjoel: To clarify, we've always been having the module and package names match, right? Is that standard or common? And to change the package name, we'd just change this in the pyproject.toml?

```
[tool.poetry]
name = "ni_measurement_plugin_sdk_service"
```

dixonjoel: I don't know how we would shorten this without using potentially not-understandable abbreviations and still keep the information to distinguish it from other packages. We're going from 27 to 33 characters - I don't see how that puts us over the top somehow.

mshafer-NI: > Is that standard or common? 

Yes. Only exception to this I'm away of is beautifulsoup4 V bs4 (https://pypi.org/project/beautifulsoup4/), which resulted in them having to publish a dummy package to avoid name squatting: https://pypi.org/project/bs4/

> And to change the package name, we'd just change this in the pyproject.toml?

Yes, but the module should ALSO have a short name :)

dixonjoel: Ok. How about 'nipss'? Just kidding. @rfriedma @bkeryan could you live with `ni_measurement_sdk_service` and `ni_measurement_sdk_generator`? Then a top-level would be `ni_measurement_plugin_sdk`.

Or, `ni_measurement_generator`, `ni_measurement_service`?
Or, `ni_meas_sdk_service`, `ni_meas_sdk_generator`.

Of course all of these lead to inconsistency with LabVIEW.

rfriedma: I don't love the inconsistency between languages and the removal of description. We would tell python developers "you need the "measurement plug-in sdk" and that maps to the current proposal and isn't unique. While variations of "NI Measurement" is a pretty overloaded term. This is specifically the package developers need to write Measurement Plug-Ins

bkeryan: > We're going from 27 to 33 characters - I don't see how that puts us over the top somehow.

There is no hard limit at this level, but this name has always been uncomfortably long. I brought it up here because it seems like we're going in the wrong direction.

There is a hard limit of 260 characters at the filesystem level, if you don't enable long path support on Windows or your application (e.g. TestStand.exe) doesn't enable long path support. 6 more characters is not a dealbreaker overall, but it increases the chances of putting the path length over 260 characters.

As I pointed out in the previous rename PR, `ni_measurement_plugin_sdk` is one character shorter than `ni_measurementlink_service`, which is not much, but at least it's going in the right direction.

```
0000000000111111111122222222223333333333
1234567890123456789012345678901234567890
ni_measurement_service
ni_measurementlink_service
ni_measurement_plugin_sdk
ni_measurement_plugin_sdk_service
```

@dixonjoel @rfriedma If there are no acceptable alternatives, I'm willing to approve this, but I really wish we had chosen a shorter name.

> Yes. Only exception to this I'm away of is beautifulsoup4 V bs4 (https://pypi.org/project/beautifulsoup4/), which resulted in them having to publish a dummy package to avoid name squatting: https://pypi.org/project/bs4/

Yeah, I would love to use `ni` as a namespace package but there is an 0.0.1 of a microservices web framework package from 2016 called `ni`.

bkeryan: > This is specifically the package developers need to write Measurement Plug-Ins

@rfriedma This is specifically the package developers need to write services for the Measurement Plug-In SDK?

rfriedma: @bkeryan it's the package developers need to write measurement plug-ins and VIPM lists packages in a big table so "Measurement Plug-In SDK" became a prefix to relate the granular packages to each other.

rfriedma: it's more like "this is what you need to write a service using the Measurement Plug-In SDK"

rfriedma: if we can't maintain equivalence to LV naming then let's aim for structural equivalence and shorten words
ni_meas_plugin_sdk_service
ni_meas_plugin_sdk_generator

bkeryan: I'm not saying "you can't maintain equivalence to LV naming". I want to make sure you are aware that this is a long name and that you have considered whether there are any shorter alternatives.

It is also possible to abbreviate "service" to [svc](https://en.wiktionary.org/wiki/svc#English).
```
ni_measurement_service
ni_measurementlink_service
ni_measurement_plugin_sdk
ni_measurement_plugin_sdk_service
ni_meas_plugin_sdk_service
ni_meas_plugin_sdk_svc
```

However, I shudder to think of saying "enn aye measzh plugin ess dee kay ess vee sea" in conversation, so I give up. I approve this PR as-is.

bkeryan (approved):

.github/ISSUE_TEMPLATE/bug_report.md
dixonjoel: So am I understanding you prefer ni_measurement_plugin_sdk_service to ni_meas_plugin_sdk_service? I can roll back my last few commits.

bkeryan: Yes.

PR #752, opened 2024-06-14
Update discovery client JSON service path

lschroed:

Updates the discovery client to look for the discovery service JSON at `NISHAREDDIR64/Services/Services.json`.

The path for the discovery service JSON is being changed, the client needs to look for it in the new location.

Ran `pytest` locally.

### DIFF
--- a/ni_measurement_plugin_sdk_service/discovery/_support.py
+++ b/ni_measurement_plugin_sdk_service/discovery/_support.py
@@ -57,12 +57,7 @@ def _get_discovery_service_location() -> pathlib.PurePath:
 
 def _get_registration_json_file_path() -> pathlib.Path:
     if sys.platform == "win32":
-        json_path = _get_nipath("NISHAREDDIR64") / "Services" / "Services.json"
-        if not json_path.exists():
-            json_path = (
-                _get_nipath("NISHAREDDIR64") / "MeasurementLink" / "MeasurementLinkServices.json"
-            )
-        return json_path
+        return _get_nipath("NISHAREDDIR64") / "Services" / "Services.json"
     else:
         raise NotImplementedError("Platform not supported")
 
--- a/ni_measurement_plugin_sdk_service/discovery/_support.py
+++ b/ni_measurement_plugin_sdk_service/discovery/_support.py
@@ -57,7 +57,7 @@ def _get_discovery_service_location() -> pathlib.PurePath:
 
 def _get_registration_json_file_path() -> pathlib.Path:
     if sys.platform == "win32":
-        json_path = _get_nipath("NISHAREDDIR") / "Services" / "Services.json"
+        json_path = _get_nipath("NISHAREDDIR64") / "Services" / "Services.json"
         if not json_path.exists():
             json_path = (
                 _get_nipath("NISHAREDDIR64") / "MeasurementLink" / "MeasurementLinkServices.json"
--- a/ni_measurement_plugin_sdk_service/discovery/_support.py
+++ b/ni_measurement_plugin_sdk_service/discovery/_support.py
@@ -57,7 +57,7 @@ def _get_discovery_service_location() -> pathlib.PurePath:
 
 def _get_registration_json_file_path() -> pathlib.Path:
     if sys.platform == "win32":
-        return _get_nipath("NISHAREDDIR64") / "Services" / "Services.json"
+        return _get_nipath("NISHAREDDIR64") / "Services" / "NIServices.json"
     else:
         raise NotImplementedError("Platform not supported")
 
--- a/ni_measurement_plugin_sdk_service/discovery/_support.py
+++ b/ni_measurement_plugin_sdk_service/discovery/_support.py
@@ -57,7 +57,12 @@ def _get_discovery_service_location() -> pathlib.PurePath:
 
 def _get_registration_json_file_path() -> pathlib.Path:
     if sys.platform == "win32":
-        return _get_nipath("NISHAREDDIR64") / "Services" / "NIServices.json"
+        json_path = _get_nipath("NISHAREDDIR64") / "Services" / "NIServices.json"
+        if not json_path.exists():
+            json_path = (
+                _get_nipath("NISHAREDDIR64") / "MeasurementLink" / "MeasurementLinkServices.json"
+            )
+        return json_path
     else:
         raise NotImplementedError("Platform not supported")
 
--- a/ni_measurement_plugin_sdk_service/discovery/_support.py
+++ b/ni_measurement_plugin_sdk_service/discovery/_support.py
@@ -57,7 +57,12 @@ def _get_discovery_service_location() -> pathlib.PurePath:
 
 def _get_registration_json_file_path() -> pathlib.Path:
     if sys.platform == "win32":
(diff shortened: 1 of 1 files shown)


lschroed:
@jonlugoatx, could you confirm the updated path is correct?

jonlugoatx:
> @jonlugoatx, could you confirm the updated path is correct?

Yes, that's correct.

jonlugoatx (approved):

ni_measurement_plugin_sdk/discovery/_support.py
```diff
@@ -57,7 +57,12 @@ def _get_discovery_service_location() -> pathlib.PurePath:
 
 def _get_registration_json_file_path() -> pathlib.Path:
     if sys.platform == "win32":
-        return _get_nipath("NISHAREDDIR64") / "MeasurementLink" / "MeasurementLinkServices.json"
+        json_path = _get_nipath("NISHAREDDIR64") / "Services" / "Services.json"
+        if not json_path.exists():
```
dixonjoel, line 61: I don't see why we need this fallback. We plan to update the minimum required version of InstrumentStudio to 2024 Q3 in the docs, so I don't think we expect anyone to try our next release with older InstrumentStudio / MeasurementLink. @rfriedma ?

dixonjoel (approved):
Approved with suggestion to remove the fallback.

ni_measurement_plugin_sdk/discovery/_support.py
rfriedma: presuming the developer has 2024 Q3+ installed then it would seem a fallback isn't required

bkeryan: You are also renaming the package, so it's not like existing applications will automatically upgrade.

bkeryan (approved):

ni_measurement_plugin_sdk/discovery/_support.py
lschroed: I've removed the fallback with the latest update.

ni_measurement_plugin_sdk/discovery/_support.py
jasonmreding, line 1: > Updates the discovery client to look for the discovery service JSON at NISHAREDDIR64/Services/Services.json first. If that does not exist, then try to use the legacy MeasurementLink JSON path.

I don't see any fallback to the old path here. I'm not sure this is entirely needed unless we want to support the measurement running against older versions of InstrumentStudio. If the test suite is using older installations of our services, then this will definitely be needed.

lschroed: I removed the fallback based on @dixonjoel's suggestion. Is that not correct to remove? I've updated the PR description to reflect that change.

jasonmreding: I don't know that it is strictly needed. I just know that it's different than the [C# client](https://dev.azure.com/ni/DevCentral/_git/ASW?path=/Source/MeasurementServices/DiscoveryClient/ServicesConfigurationFileLocator.cs&version=GBmain&line=74&lineEnd=84&lineStartColumn=1&lineEndColumn=18&lineStyle=plain&_a=contents). @jonlugoatx is there anything I'm missing where you know the fallback is required?

For Python, I guess if you weren't using a virtual environment and somebody updated to the latest release of this package for the system, then it might break existing python based measurements if an older version of InstrumentStudio was installed. I know that's not a deployment strategy we would recommend though.

bkeryan: @jasonmreding 

> [bkeryan](https://github.com/bkeryan) [2 hours ago](https://github.com/ni/measurement-plugin-python/pull/752#discussion_r1640237247)
> You are also renaming the package, so it's not like existing applications will automatically upgrade.

jasonmreding: That's true. I forgot about those pending PRs in the context of this change. So I guess it is really a question of whether or not we care about supporting older versions of IS with the latest release of this repo or not. And more specifically, whether we care about supporting dev work flows for measurements where the measurement is started manually from the command line rather than from IS or TS.

rfriedma: we're moving the minimum requirements for this repo to IS 2024 Q3. I'm more concerned with existing measurements built on measurementlink continuing to be runnable without error where possible. If needed, I'd accept the scenario that old measurements can no longer automatically start the discovery service when running from the dev workflow. Hopefully the upgrade path is easy enough that developers can easily fix the dev workflow by simply upgrading to latest.

bkeryan:
BTW @lschroed please update the PR title so it isn't just your dev branch name.

ni_measurement_plugin_sdk/discovery/_support.py
```diff
@@ -57,7 +57,7 @@ def _get_discovery_service_location() -> pathlib.PurePath:
 
 def _get_registration_json_file_path() -> pathlib.Path:
     if sys.platform == "win32":
-        return _get_nipath("NISHAREDDIR64") / "MeasurementLink" / "MeasurementLinkServices.json"
+        return _get_nipath("NISHAREDDIR64") / "Services" / "Services.json"
```
jonlugoatx, line 60: After discussing with Ryan Eckenrode, we decided to change this to `NIServices.json` since searching the ancestor directory hierarchy for Services.json from the C# client is too generic and might pick up some JSON file we don't own.

jonlugoatx (changes requested):

ni_measurement_plugin_sdk/discovery/_support.py
lschroed: Updated to `NIServices.json`.

ni_measurement_plugin_sdk/discovery/_support.py
jasonmreding: @rfriedma If that's the case, then why bother to still support V1 measurements. It seems like we could slowly clean some things up if we're not going to support older clients any longer. I guess there is still the question about what to do with older TestStand clients.

rfriedma: I'm not sure I described that well then. Old measurements should keep working from TS and IS. My understanding is that Measurement Steps and IS both launch the Discovery Service. I was reading this as "there's potential that in dev workflows, the discovery service won't automatically get launched with the latest version of this SDK" which seemed like a fairly contained slice that doesn't impact older measurements running from TS and IS.

jonlugoatx (approved):

ni_measurement_plugin_sdk/discovery/_support.py
lschroed: Should we add the fallback path back in then? That doesn't seem to add risk and allows for later dropping support of old clients.

jasonmreding: I'll defer to code owners, but if it is easy and it is consistent with our clients in other languages, then I would say yes.

bkeryan: Restoring the fallback path is fine with me.

bkeryan (approved):

Issue #366, opened 2023-09-01
Enum arrays do not work with an empty default value


bkeryan:
# Bug Report

Deserializing an enum array does not work when the configuration's `default_value` parameter is set to `[]` (empty list).

## Repro or Code Sample

```diff
--- a/examples/sample_measurement/measurement.py
+++ b/examples/sample_measurement/measurement.py
@@ -55,6 +55,9 @@ class Color(Enum):
 @sample_measurement_service.configuration(
     "String Array In", nims.DataType.StringArray1D, ["String1", "String2"]
 )
+@sample_measurement_service.configuration(
+    "Enum Array In", nims.DataType.EnumArray1D, [], enum_type=Color
+)
 @sample_measurement_service.output("Float out", nims.DataType.Float)
 @sample_measurement_service.output("Double Array out", nims.DataType.DoubleArray1D)
 @sample_measurement_service.output("Bool out", nims.DataType.Boolean)
@@ -72,6 +75,7 @@ def measure(
     enum_input,
     protobuf_enum_input,
     string_array_in,
+    enum_array_in,
 ):
     """Perform a loopback measurement with various data types."""
     logging.info("Executing measurement")
@@ -81,6 +85,8 @@ def measure(

     sample_measurement_service.context.add_cancel_callback(cancel_callback)

+    logging.info("enum_array_in: %r", enum_array_in)
+
     float_output = float_input
     float_array_output = double_array_input
     bool_output = bool_input
```

Steps:
- Open **MeasurementLink UI Editor**.
- Drop a **Numeric Array Input**.
- Bind it to `Enum Array In`.
- Set the first element to 1.
- Run the measurement. 

## Expected Behavior

The measurement logs this message:
```
enum_array_in: [<Color.RED: 1>]
```

## Current Behavior

The measurement produces this error:
```
PS D:\dev\measurementlink-python\examples\sample_measurement> poetry run python .\measurement.py -v
2023-09-01 17:48:26,593 INFO: Measurement service hosted on port: 52794
2023-09-01 17:48:27,123 INFO: Successfully registered with discovery service.
Press enter to close the measurement service.
2023-09-01 17:49:09,385 INFO: gRPC server call /ni.measurementlink.measurement.v2.MeasurementService/GetMetadata responded OK in 0.9639 ms
2023-09-01 17:50:07,890 INFO: gRPC server call /ni.measurementlink.measurement.v2.MeasurementService/Measure responded OK in 0.6051 ms
[...]

PR #767, opened 2024-06-21
Replaced internal encoders with message encoder calling public API

LazeringDeath:

Replaced the Google's internal encoders with message encoders without affecting performance or results.

Resolve [issue#35](https://github.com/ni/measurement-plugin-python/issues/35) and make the serializer more robust/less complicated.

Passed all measurementlink tests and PR checks.


ni_measurement_plugin_sdk_service/_internal/parameter/metadata.py
```diff
@@ -50,22 +49,20 @@ def validate_default_value_type(parameter_metadata: ParameterMetadata) -> None:
     Raises:
         TypeError: If default value does not match the Datatype.
     """
+    from ni_measurement_plugin_sdk_service._internal.parameter.message_serializer import (
```
dixonjoel, line 52: Move this import up to the top of the module.

ni_measurement_plugin_sdk_service/_internal/parameter/serialization_strategy.py
dixonjoel, line 1: Consider renaming this file to `decoder_strategy.py` since it is only for decoding now.

ni_measurement_plugin_sdk_service/_internal/parameter/message_serializer.py
dixonjoel, line 1: I think this is a good chance to split up the encoder / decoder into two distinct modules. So I would just call this `encoder.py` and then add `decoder.py` when we're ready to add that.

ni_measurement_plugin_sdk_service/_internal/parameter/serializer.py
dixonjoel, line 1: Since this is only doing decoding now, let's rename this file to `decoder.py` and then we'll replace it with new implementation when we do the decoding side.

tests/unit/test_serialization_strategy.py
dixonjoel, line 1: Will be renamed to `test_decoder_strategy.py` when that file is renamed.

ni_measurement_plugin_sdk_service/_internal/parameter/message_serializer.py
```diff
@@ -0,0 +1,238 @@
...
+
+        parameter_value (Sequence[Any]): Parameter values to serialize.
+
+    Returns:
+        bytes: Serialized byte string containing parameter values.
+    """
+    pool = descriptor_pool.Default()
+    file_descriptor_proto = descriptor_pb2.FileDescriptorProto()
+    original_guid = uuid4()
+    new_guid = "msg" + "".join(filter(str.isalnum, str(original_guid)))[:16]
```
dixonjoel, line 33: Add a str() around all of this so we don't have to do it multiple times below. I would rename the variable, e.g. `unique_descriptor_name`

tests/unit/test_decoder.py
dixonjoel, line 1: This file name isn't accurate. This tests the encoder and decoder. I'd make an additional test file `test_encoder.py` and move the serializer tests there.

bkeryan:
Please update the PR description.

bkeryan (changes requested):
I'm not done reviewing this. FYI, my team is doing release finalization now, so I might not get back to this for a few days.

ni_measurement_plugin_sdk_service/_internal/parameter/decoder_strategy.py
```diff
@@ -245,7 +150,17 @@ def get_decoder(
...
+        type_pb2.Field.TYPE_UINT64: int(),
+        type_pb2.Field.TYPE_BOOL: bool(),
+        type_pb2.Field.TYPE_STRING: str(),
+        type_pb2.Field.TYPE_ENUM: int(),
+    }
     if repeated:
         return list()
-    type_default_value = _TYPE_DEFAULT_MAPPING.get(type)
-    return type_default_value
+    return _type_default_mapping.get(type)
```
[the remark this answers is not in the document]
LazeringDeath: I moved it there to make everything under one function, I wasn't worried about avoiding any bugs or anything like that. Although I don't think I need to use copy.copy to avoid that bug so I'll move it above the function and leave it as that, is that ok?

ni_measurement_plugin_sdk_service/_internal/parameter/metadata.py
```diff
@@ -54,18 +54,12 @@ def validate_default_value_type(parameter_metadata: ParameterMetadata) -> None:
...
-    )
+    expected_type = type(get_type_default(parameter_metadata.type, parameter_metadata.repeated))
     display_name = parameter_metadata.display_name
     enum_values_annotation = get_enum_values_annotation(parameter_metadata)
 
     if parameter_metadata.repeated:
-        expected_element_type = type(
-            serialization_strategy.get_type_default(parameter_metadata.type, False)
-        )
+        expected_element_type = type(get_type_default(parameter_metadata.type, False))
```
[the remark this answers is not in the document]
LazeringDeath: I'm not sure why it's there since it was already there before I changed anything, although I think it's to get the class from get_type_default() since it would return 0, 0.0, "", or  False.  So I didn't intend it to make a copy of the function or anything like that.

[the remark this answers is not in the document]
bkeryan: Nevermind, I was misreading this. `type(x)` queries the type of `x`. I was reading it as if `type` was a variable that contained a type.

ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
```diff
@@ -0,0 +1,222 @@
...
+
+        parameter_value (Sequence[Any]): Parameter values to serialize.
+
+    Returns:
+        bytes: Serialized byte string containing parameter values.
+    """
+    pool = descriptor_pool.Default()
+    file_descriptor_proto = descriptor_pb2.FileDescriptorProto()
+    original_guid = str(uuid4())
+    unique_descriptor_name = str("msg" + "".join(filter(str.isalnum, original_guid))[:16])
```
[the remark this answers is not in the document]
LazeringDeath: I think it's something we came up with, Joel commented I should changed from 'new_guid' to 'unique_descriptor_name'. Do you want me to change it back? Or remove the 'unique' part of the name?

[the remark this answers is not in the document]
bkeryan: I think we may need to pass more information into these functions so that they can reuse the same descriptor proto rather than creating a new one each time.

ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
```diff
@@ -0,0 +1,222 @@
...
+    file_descriptor_proto.name = unique_descriptor_name
+    file_descriptor_proto.package = unique_descriptor_name
+
+    # Create a DescriptorProto for the message
+    message_proto = file_descriptor_proto.message_type.add()
```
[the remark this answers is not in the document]
LazeringDeath: Me and Joel talked and were confused how Python wouldn't be able throw the descriptor proto in the garbage collector since it wouldn't be used again after calling the function.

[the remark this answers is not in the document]
bkeryan: The descriptor pool is a global object: `pool = descriptor_pool.Default()`. There is one default descriptor pool per process. `descriptor_pool.Default()` is an accessor function that returns the default descriptor pool: https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/descriptor_pool.py#L1354

Every time you call this function, you add a new descriptor to the pool. If you call it 500 times, the pool will contain 500 descriptors. If you call it 1,000,000 times, the pool will contain 1,000,000 descriptors.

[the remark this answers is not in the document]
bkeryan: Put differently: you are creating a new message type every time you call this function. If you call it multiple times for the same service, you are creating unnecessary message types and storing them in the descriptor pool so they are never garbage collected.

If we passed more information to this function, then it could make a decision about whether to create a new message type or reuse a message type that it already created. Then it would only create a new message type the first time it's called for a specific service.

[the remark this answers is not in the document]
LazeringDeath: Is it ok to to have one file descriptor proto then one message type per service in the pool instead of creating one file descriptor then adding the message types to it? It's not possible to update the file descriptor, that's already in the pool, with a new message type. Also, me and Joel couldn't find a way to remove a file descriptor from the pool to see if we could make a new, updated file descriptor then add that to the pool while removing the old one. So this was the best method Joel came up with.

ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
```diff
@@ -0,0 +1,202 @@
...
+    parameter_metadata_dict: Dict[int, ParameterMetadata],
+    parameter_values: Sequence[Any],
+    service_info: ServiceInfo,
+) -> bytes:
+    """Serialize the parameter values in same order based on the metadata_dict.
+
+    Args:
+        parameter_metadata_dict (Dict[int, ParameterMetadata]): Parameter metadata by ID.
+
+        parameter_value (Sequence[Any]): Parameter values to serialize.
```
dixonjoel, line 28: Add an 'arg' statement for service_info

ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
```diff
@@ -0,0 +1,202 @@
...
+            enum_value_descriptor.name = name
+            enum_value_descriptor.number = number.value
+    # checks enum if it's protobuf or python
+    try:
+        field_descriptor.type_name = ProtobufColor.DESCRIPTOR.full_name
+    except TypeError:
+        field_descriptor.type_name = param.__class__.__name__
+
+
+def _create_fields(message_proto: Any, metadata: ParameterMetadata, index: int) -> Any:
```
dixonjoel, line 172: I'd make this `_create_field` since it's only doing one. Also update the parameters in the docstrings - there's some extra ones.

dixonjoel (approved):
Approved with suggestions.

ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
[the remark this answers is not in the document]
bkeryan: Creating a separate file descriptor per service is fine. Making an infinite number of calls to those services would consume a finite amount of memory, given a finite set of services.

My concern was that creating a file descriptor for each call consumes an infinite amount of memory, given a finite set of services.

bkeryan (commented):
I'll try to get back to this today.

ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
```diff
@@ -0,0 +1,206 @@
...
+    try:
+        message_proto = pool.FindMessageTypeByName(f"{message_name}.{message_name}")
+    except KeyError:
+        message_proto = _create_message_type(
+            parameter_values, parameter_metadata_dict, message_name, pool
+        )
+    message_instance = message_factory.GetMessageClass(message_proto)()
+
+    for i, parameter in enumerate(parameter_values, start=1):
+        field_name = f"field_{i}"
```
bkeryan, line 47: I would prefer to use the configuration/output names specified by the measurement service. Using numeric field names loses some of the debugging benefits of protobuf, like the ability to print a message as a string. 

You may need to "sanitize" them, if protobuf won't accept spaces or other weird characters that we allow. If you do this, it may make sense for the `ParameterMetadata` to save the sanitized name so we don't have to build a bunch of temporary strings whenever we serialize fields.

bkeryan (changes requested):

ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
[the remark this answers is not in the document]
bkeryan: I think you should put both the Configurations and Outputs messages in the same file descriptor so they can share enums.

ni_measurement_plugin_sdk_service/_internal/parameter/_decoder_types.py
dixonjoel, line 1: I don't think we need these definitions anymore. Is this used by anyone?

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
dixonjoel, line 1: Update this doc string.

dixonjoel:
@LazeringDeath generally looks good. I'll take a more thorough look when all the checks are passing.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    service_name = "".join(char for char in service_name if char.isalpha())
+    try:
+        pool.FindFileByName(service_name)
+    except KeyError:
+        file_descriptor = descriptor_pb2.FileDescriptorProto()
+        file_descriptor.name = service_name
+        file_descriptor.package = service_name
+
+        _create_message_type(input_metadata, "Inputs", file_descriptor)
```
bkeryan, line 142: "Inputs" is inconsistent with the rest of the codebase. Like it or not, these things are called "configurations", not "inputs".

https://github.com/ni/ni-apis/blob/main/ni/measurementlink/measurement/v2/measurement_service.proto#L78

packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
```diff
@@ -363,5 +386,11 @@ def Measure(  # noqa: N802 - function name should be lowercase
 
     def _serialize_response(self, outputs: Any) -> v2_measurement_service_pb2.MeasureResponse:
         return v2_measurement_service_pb2.MeasureResponse(
-            outputs=_serialize_outputs(self._output_metadata, outputs)
+            outputs=_serialize_outputs(
+                self._output_metadata, outputs, self._get_service_name() + ".Outputs"
+            )
         )
+
+    def _get_service_name(self) -> str:
```
bkeryan, line 394: Don't duplicate this method. Make a free function at top level so both classes can use it.

ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
```diff
@@ -0,0 +1,206 @@
...
+        Service_info (ServiceInfo): Unique service name.
+
+    Returns:
+        bytes: Serialized byte string containing parameter values.
+    """
+    pool = descriptor_pool.Default()
+    message_name = "".join(char for char in service_info.service_class if char.isalpha())
+    # Tries to find a message type in pool with message_name else it creates one
+    try:
+        message_proto = pool.FindMessageTypeByName(f"{message_name}.{message_name}")
```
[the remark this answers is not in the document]
bkeryan: > I think it would be entirely reasonable to use the service class as the package part and use "Configurations" or "Outputs" as the message part, depending on which one you are creating. 

Please do this.

_get_service_name is stripping the dots from the package name. Can you use it as-is, or is there some problem with that approach?

Name the messages "Configurations" and "Outputs", not "Inputs" and "Outputs".

ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    default_value_parameter_array = [
+        parameter.default_value for parameter in parameter_metadata_dict.values()
+    ]
+    return serialize_parameters(
+        parameter_metadata_dict, default_value_parameter_array, service_info
+    )
+
+
+def _create_message_type(
```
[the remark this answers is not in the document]
bkeryan: ok, good

ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
```diff
@@ -0,0 +1,206 @@
...
+        pool (descriptor_pool.DescriptorPool): Descriptor pool holding file descriptors.
+
+    Returns:
+        Any: A message descriptor based on a defined message_descriptor
+    """
+    file_descriptor = descriptor_pb2.FileDescriptorProto()
+    file_descriptor.name = message_name
+    file_descriptor.package = message_name
+    message_proto = file_descriptor.message_type.add()
+    message_proto.name = message_name
```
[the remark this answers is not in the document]
bkeryan: ok, good

ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
bkeryan: Protobuf field names should be in snake_case, but ParameterMetadata.sanitized_display_name is removing the underscores.

I was suggesting that you add the sanitized name to ParameterMetadata instead of building a new string every you serialize or deserialize a property. Does this matter for performance? I don't know, I'd have to benchmark it. A compromise would be to scan the string for disallowed characters and only build a new string when any are present.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/metadata.py
```diff
@@ -40,6 +42,15 @@ class ParameterMetadata(NamedTuple):
     Required when 'type' is Kind.TypeMessage. Ignored for any other 'type'.
     """
 
+    def sanitized_display_name(self) -> str:
```
bkeryan, line 45: Use `@property` and simplify the docstring (no "Returns:").

This codebase has type hints, so we don't have to duplicate the types in the docstring.

bkeryan, line 45: "sanitized_display_name" is not wrong, but something like "field_name" or "protobuf_name" would better describe what this is for.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
```diff
@@ -0,0 +1,177 @@
...
+        file_descriptor: Descriptor of proto file.
+
+    Returns:
+        Any: Matching enum class in a str type or None when enum is protobuf.
+    """
+    for enum_type in file_descriptor.enum_type:
+        enum_names = [enum_value.name for enum_value in enum_type.value]
+        if sorted(metadata_enum_list) == sorted(enum_names):
+            return enum_type.name
+    return None
```
bkeryan, line 34: Why is this an Any? From the code, it looks like an `Optional[str]`, assuming that `enum_type.name` is a`str` (is it?)

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
```diff
@@ -0,0 +1,177 @@
...
+    field_descriptor: FieldDescriptorProto,
+) -> None:
+    """Implement a enum class in 'file_descriptor'.
+
+    Args:
+        file_descriptor (FileDescriptorProto): Descriptor of a proto file.
+
+        parmeter_metadata (ParameterMetadata): Metadata of current field.
+
+        field_descriptor (FieldDescriptorProto): Descriptor of a field.
```
bkeryan, line 49: Everywhere: don't duplicate types in the docstring.

For internal functions, consider not writing the full docstring boilerplate.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
```diff
@@ -0,0 +1,177 @@
...
+
+    Returns:
+        None: Only creates a enum class in file_descriptor.
```
bkeryan, line 52: "Returns:" is typically not written for functions that return None.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
```diff
@@ -0,0 +1,177 @@
...
+from typing import Any, List
+
+from google.protobuf import descriptor_pb2, descriptor_pool
+from google.protobuf.descriptor_pb2 import FieldDescriptorProto
+
+from ni_measurement_plugin_sdk_service._annotations import ENUM_VALUES_KEY
+from ni_measurement_plugin_sdk_service._internal.parameter.metadata import (
+    ParameterMetadata,
+)
+from tests.utilities.stubs.loopback.types_pb2 import ProtobufColor
```
bkeryan, line 13: This is shipping code. It can't import test code.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
```diff
@@ -0,0 +1,177 @@
...
+    enum_dict = loads(parameter_metadata.annotations[ENUM_VALUES_KEY])
+    if parameter_metadata.default_value is None:
+        enum_type_name = _get_output_enum_type(
+            metadata_enum_list=[enum for enum in enum_dict.keys()], file_descriptor=file_descriptor
+        )
+    else:
+        enum_type_name = _get_enum_type(parameter_metadata).__name__
+
+    if enum_type_name == "int" or enum_type_name is None:
+        field_descriptor.type_name = ProtobufColor.DESCRIPTOR.full_name
```
bkeryan, line 63: This is where it references test code. Delete this or replace it with working code that doesn't rely on hardcoding details of the tests.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
```diff
@@ -0,0 +1,177 @@
...
+
+        pool (DescriptorPool): Descriptor pool holding file descriptors and enum classes.
+
+    Returns:
+        None: Only creates a file and two message descriptors.
+    """
+    service_name = "".join(char for char in service_name if char.isalpha())
+    try:
+        pool.FindFileByName(service_name)
+    except KeyError:
```
bkeryan, line 137: Good, this handles the case where the file descriptor is already defined.

It would also make sense for this to compare the previously registered file descriptor to the configuration/output metadata and error if they don't match.

bkeryan (changes requested):

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
dixonjoel: Is there a concern that these will change? As far as I could think, they'll get created upon start of the service and put into the pool. After that, it won't change unless the Python process also restarts in which case the pool will be empty again.

Would this check just be checking that we added it correctly during service start or is there some other concern about it changing after service start?

bkeryan: Yes, the concern is that a test or an advanced user will start/stop services multiple times with slightly different config/output parameters. I don't think it matters for basic usage like in the example services.

If FileDescriptor messages are comparable, then it would be easy to do this check: generate the FileDescriptor again and compare the old and new FileDescriptors. 

If it's not that easy, I think it's ok not to bother.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
LazeringDeath: It is, I didn't know how to make the return type be 2 different types so that's why I made it as a Any.

mshafer-NI: Whenever you're dealing with two different possible types, you use `typing.Union[type1, type2]` (parameter or return). `typing.Optional[foo]` is shorthand for `typing.Uniont[None, foo]`.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
LazeringDeath: Is it fine to create a enum type every time there is a enum passing in, regardless if it's a python or a protobuf enum? Since I'm not sure how to get the full_name from a protobuf enum which is why I was importing in the test code since it was defined in there.

If so, then what would be the name for the enum type when it's a protobuf? I've made it where it get's the first letter of each enum field name then joining them together, similar to a acronym without the dots but I don't think this is good enough.

bkeryan: It is not fine to create new Python types every time you serialize or deserialize something.

If the service uses an enum, the Python type for the enum should already exist.

For example, this service uses the Function enum: https://github.com/ni/measurement-plugin-python/blob/main/examples/nidmm_measurement/measurement.py
Registering the service should not create another Python type for that enum. It can create a protobuf EnumDescriptor or whatever, though.

bkeryan: Are you asking whether it's ok to create a new EnumDescriptor when the service is already using a protobuf enum? That might be ok.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
LazeringDeath: Me and Joel couldn't find a easy way to compare these because both file descriptors return different types from each other and it also compares the pointers, not the contents in them.

bkeryan: Thanks, I'll drop this issue.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
```diff
@@ -1,76 +1,43 @@
...
     enum_dict = loads(parameter_metadata.annotations[ENUM_VALUES_KEY])
-    if parameter_metadata.default_value is None:
-        enum_type_name = _get_output_enum_type(
-            metadata_enum_list=[enum for enum in enum_dict.keys()], file_descriptor=file_descriptor
-        )
-    else:
-        enum_type_name = _get_enum_type(parameter_metadata).__name__
+    enum_type_name = _get_enum_type(parameter_metadata).__name__
+
+    # if enum is a protobuf then enum_type_name is 1st letter of each enum name
```
bkeryan, line 27: This comment makes it sound like the first-letter thing is a protobuf behavior that you are trying to match, not a behavior that you came up with because you don't have the enum name.

If this is a protobuf enum, we have its name, but we currently throw it away after building the annotation JSON. You could change that code to pass enum_type into the ParameterMetadata initialize function.

There is also a related bug: https://github.com/ni/measurement-plugin-python/issues/366
Inferring the enum type from the default value does not work when the default value is an empty list.

See
https://github.com/ni/measurement-plugin-python/blob/a6bc38ea4fbe7f6dc624e752382b681aecbda6f6/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py#L415

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
```diff
@@ -1,76 +1,43 @@
...
+    # e.g. {"NONE": 0, "RED": 1, "GREEN": 2} -> NRG
+    if enum_type_name == "int" or enum_type_name == "NoneType":
```
bkeryan, line 29: Why does this check for both int and NoneType? Protobuf doesn't use both of these, does it? The comment says we do this for protobuf enums but are we doing it for some other case too?

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/metadata.py
```diff
@@ -42,14 +42,30 @@ class ParameterMetadata(NamedTuple):
...
-        """Parameter display name of alpha/numerical characters.
+    field_name: str = ""
+    """display_name in snake_case format."""
 
-        Returns:
-            str: Alpha/numerical characters of 'display_name'.
 
-        """
-        return "".join(char for char in self.display_name if char.isalnum())
+def initialize(
```
bkeryan, line 49: [style] Make this a `@classmethod` inside the ParameterMetadata class so you can import ParameterMetadata and then call ParameterMetadata.initialize. I don't want it to be imported as just `initialize` like you did in the tests because that fails the "what if two people did this?" test and the resulting code is not clear what it's initializing.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
```diff
@@ -1,76 +1,43 @@
...
-            enum_value_descriptor.name = name
+            enum_value_descriptor.name = f"{enum_type_name}_{name}"
             enum_value_descriptor.number = number
-        field_descriptor.type_name = enum_descriptor.name
-    else:
-        field_descriptor.type_name = enum_type_name
+    field_descriptor.type_name = enum_type_name
 
 
 def _get_enum_type(parameter_metadata: ParameterMetadata) -> type:
```
bkeryan, line 43: decoder.py imports _get_enum_type, so you should remove the underscore to show that this is part of the public interface of the serialization_descriptors submodule (which is itself still private)

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/_get_type.py
```diff
@@ -0,0 +1,36 @@
+from typing import Any
+
+from google.protobuf.descriptor_pb2 import FieldDescriptorProto
+from google.protobuf.type_pb2 import Field
+
+_TYPE_DEFULAT_MAPPING = {
```
bkeryan, line 6: spelling

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/_get_type.py
```diff
@@ -0,0 +1,36 @@
...
+    Field.TYPE_INT32: int(),
+    Field.TYPE_INT64: int(),
+    Field.TYPE_UINT32: int(),
+    Field.TYPE_UINT64: int(),
+    Field.TYPE_BOOL: bool(),
+    Field.TYPE_STRING: str(),
+    Field.TYPE_ENUM: int(),
+}
+
+_TYPE_FIELD_MAPPING = {
```
bkeryan, line 18: serialization_descriptors.py imports _TYPE_FIELD_MAPPING, so you should remove the underscore to show that this is part of the public interfaces of the _get_type submodule (which is itself still private)

Or you can add a function wrapping this dict, like you did for get_type_default.

bkeryan (changes requested):
I think you're on the home stretch.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
```diff
@@ -0,0 +1,107 @@
...
+    parameter_metadata: ParameterMetadata,
+    field_descriptor: FieldDescriptorProto,
+) -> None:
+    """Implement a enum class in 'file_descriptor'."""
+    enum_dict = loads(parameter_metadata.annotations[ENUM_VALUES_KEY])
+    enum_type_name = _get_enum_type(parameter_metadata).__name__
+
+    # if enum is a protobuf then enum_type_name is 1st letter of each enum name
+    # e.g. {"NONE": 0, "RED": 1, "GREEN": 2} -> NRG
+    if enum_type_name == "int" or enum_type_name == "NoneType":
```
bkeryan, line 29: Comparing types by name seems kind of hacky. What if you did `enum_type = _get_enum_type(parameter_metadata)` and `if enum_type is int or enum_type is NoneType:`?

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
bkeryan: For the case where we really don't know the enum type name, can we derive it from the parameter name? Converting the parameter name to PascalCase would be a good naming scheme.

You can convert snake_case to PascalCase by splitting on "_" and capitalizing each section.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
LazeringDeath: Int is what I was getting for the type of the default value since the default values for protobuf enums are a integer, not a enum. Same thing for None since the default value for output values is None for some reason.

I just realized this condition catches both python and protobuf output enums, which I somehow thought it was ok to do. I'm not planning to use this in the next commit so this is basically resolved.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/decoder.py
```diff
@@ -57,14 +51,18 @@ def deserialize_parameters(
...
-    return enum_type(field_value)
+    try:
+        # ValueType is defined when field_value is a protobuf enum
+        metadata.enum_type.ValueType
+        return field_value
+    except AttributeError:
+        enum_type = metadata.enum_type
+        if metadata.repeated:
+            return [enum_type(value) for value in field_value]
+        return enum_type(field_value)
```
bkeryan, line 68: Here's how to do this without throwing an exception.
```suggestion
    if hasattr(metadata.enum_type, "ValueType"):
        # ValueType is defined when field_value is a protobuf enum
        metadata.enum_type.ValueType
        return field_value
    else:
        enum_type = metadata.enum_type
        if metadata.repeated:
            return [enum_type(value) for value in field_value]
        return enum_type(field_value)
```

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/metadata.py
```diff
@@ -45,28 +45,38 @@ class ParameterMetadata(NamedTuple):
...
-    message_type: str = "",
-) -> ParameterMetadata:
-    """Initialize ParameterMetadata with field_name."""
-    underscore_display_name = display_name.replace(" ", "_")
-    if all(char.isalnum() or char == "_" for char in underscore_display_name):
-        field_name = underscore_display_name
-    else:
-        field_name = "".join(
-            char for char in underscore_display_name if char.isalnum() or char == "_"
+    enum_type: Any = None
```
bkeryan, line 48: The correct type hint is here: https://github.com/ni/measurement-plugin-python/blob/a6bc38ea4fbe7f6dc624e752382b681aecbda6f6/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py#L71

Here's what to put in metadata.py:

```python

...

if TYPE_CHECKING:
    from google.protobuf.internal.enum_type_wrapper import _EnumTypeWrapper

    SupportedEnumType = Union[Type[Enum], _EnumTypeWrapper]

...

enum_type: Optional[SupportedEnumType] = None
```

from __future__ import annotations allows type hints to use types like SupportedEnumType that only exist at type check time, not at run time.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
```diff
@@ -22,15 +23,19 @@ def _create_enum_type_class(
...
     enum_dict = loads(parameter_metadata.annotations[ENUM_VALUES_KEY])
-    enum_type_name = _get_enum_type(parameter_metadata).__name__
 
-    # if enum is a protobuf then enum_type_name is 1st letter of each enum name
-    # e.g. {"NONE": 0, "RED": 1, "GREEN": 2} -> NRG
-    if enum_type_name == "int" or enum_type_name == "NoneType":
-        enum_field_names = list(enum_dict.keys())[:]
-        enum_type_name = "".join(name[0] for name in enum_field_names)
+    # True if enum is protobuf
+    if not type(parameter_metadata.enum_type) is type(Enum):
```
bkeryan, line 28: Why not do the same check as the other file?

```suggestion
    if hasattr(parameter_metadata.enum_type, "ValueType"):
```

bkeryan (changes requested):
Hopefully this is the last round.

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py
LazeringDeath: You want me to change the try and except clause under that to the hasattr() too? Or is it best to keep as is?

bkeryan: It looks like you are checking for the same condition: "is this a protobuf enum or a Python enum?" Is that right?

The best way to do the same thing in two places is to share the code. The second best way is to use the same logic. Using different logic is not as good because there may be cases where the two approaches have different results.

So the best way is to add a function that does the hasattr check and call it in both places.

LazeringDeath: Earlier you commented this:

> For the case where we really don't know the enum type name, can we derive it from the parameter name? Converting the parameter name to PascalCase would be a good naming scheme.
> 
> You can convert snake_case to PascalCase by splitting on "_" and capitalizing each section.

Which is why there's the try and except to handle the case where enum_type.DESCRIPTOR.name isn't defined or do we not need it? I didn't ask why we need it since I didn't really think about it but is there a situation where the name isn't defined? Since from what I know, the name is required to create the enum type for a protobuf enum, right?

bkeryan (approved):

PR #781, opened 2024-07-09
Create a top-level ni-measurement-plugin-sdk meta package

dixonjoel:

Creates a new `ni-measurement-plugin-sdk` top-level package that pulls in the service and generator packages as dependencies.

We're also reorganizing the repo to have a more consistent and clear package organization.
* Make top-level `packages` directory with `sdk`, `service`, and `generator` sub-folder
* Update workflow scripts with the new directories.

Makes installation simpler by providing a single package that brings in everything from the SDK and cleans up the organization of the packages so they're consistent.

Built locally.


packages/service/tests/utilities/measurements/sample_measurement/_stubs/__init__.py
dixonjoel, line 1: @bkeryan Note that I duplicated these color stubs because the loopback_measurement was trying to import it from `examples.sample_measurement._stubs` and I wasn't sure how to make that work.

packages/sdk/README.md
```diff
@@ -0,0 +1,23 @@
...
+---
+
+## Introduction
+
+Measurement Plug-In SDK for Python (`ni-measurement-plugin-sdk`) is a
+tool for generating and writing reusable measurement plug-ins using gRPC services.
+
+The purpose of this package is as a meta-package that depends on `ni-measurement-plugin-sdk-generator`
+and `ni-measurement-plugin-sdk-service` so as to easily install all the things needed for the
+Measurmeent Plug-In SDK.
```
bkeryan, line 11:
```suggestion
Measurement Plug-In SDK.
```

packages/service/_docs_source/conf.py
bkeryan, line 1: Will this SDK have a single ReadTheDocs site documenting all of its modules, or separate ReadTheDocs sites for each module? nimi-python started with the former and changed to the latter. However, the reasons for splitting the nimi-python docs may not apply here.

Did you create a new ReadTheDocs site for this project? https://measurementlink-python.readthedocs.io/ has "MeasurementLink" in the URL.

I expect moving `_docs_source/conf.py` to break ReadTheDocs integration because it's referenced by `/.readthedocs.yml`. If the plan is to publish one ReadTheDocs site covering multiple packages, then I think it makes sense to keep `_docs_source` in the root, unless something requires it to be included in the package's source distribution.

I think it's ok to keep the ReadTheDocs stuff centered around ni_measurement_plugin_sdk_service for now and update it to support multiple packages later, when the client package is added.

packages/service/scripts/install_examples.py
bkeryan, line 1: If you're going to move this, you need to update its relative path to ROOT_DIR.

packages/service/README.md
bkeryan, line 1: This is currently identical to `/README.md`. That's fine for now but it could make sense to differentiate them in the future. The README.md inside the -service package is uploaded to PyPI as the description for the -service package.

.github/workflows/check_nims.yml
```diff
@@ -76,7 +80,7 @@ jobs:
           mkdir -p docs   
           poetry run sphinx-build _docs_source docs -b html -W --keep-going
       - name: Revert docs
-        run:  git clean -dfx docs/ && git restore docs/
+        run:  git clean -dfx ../../docs/ && git restore ../../docs/
```
bkeryan, line 83: Are you sure this is dirtying /docs and not /packages/service/docs?

.github/workflows/run_system_tests.yml
```diff
@@ -34,8 +39,10 @@ jobs:
           key: run-system-tests-${{ runner.os }}-${{ matrix.configuration }}-${{ hashFiles('poetry.lock') }}
       - name: Install dependencies
         run: poetry install -v
+        working-directory: ./packages/service
       - name: Run system tests
         run: poetry run tox
+        working-directory: ./packages/service
       - name: Upload test results
```
bkeryan, line 46: I think the defaults cover these.

.github/workflows/check_nims.yml
```diff
@@ -12,6 +12,10 @@ jobs:
   check_nims:
     name: Check NIMS
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        # Set the working-directory for all steps in this job.
+        working-directory: ./packages/service
```
bkeryan, line 18: I don't think this affects actions/cache because it is not a "run" step. It is probably caching an empty/nonexistent .venv directory now. This goes for all of the caching in the workflows.

.github/workflows/check_nimg.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     defaults:
       run:
         # Set the working-directory for all steps in this job.
-        working-directory: ./ni_measurement_plugin_sdk_generator
+        working-directory: ./packages/generator
```
bkeryan, line 18: Line 37 has a reference to the old package path.

.github/workflows/Publish_NIMS.yml
```diff
@@ -84,7 +90,13 @@ jobs:
         if: ${{ startsWith(github.event.release.target_commitish, 'main') || startsWith(github.event.release.target_commitish, 'releases/') }}
         run: |
           poetry publish --build --username __token__ --password ${{ secrets.PYPI_TOKEN }}
-        working-directory: ./ni_measurement_plugin_sdk_generator
+        working-directory: ./packages/generator
+
+      - name: Build SDK Python package and publish to PyPI
```
bkeryan, line 95: FYI: good, this one is last so it won't publish if the others fail :)

bkeryan (changes requested):

.github/workflows/check_nims.yml
bkeryan: Yeah, it's also computing a bad hash key because `${{ hashFiles('poetry.lock') }}` evaluates to an empty string when it can't find `poetry.lock`.

https://github.com/ni/measurement-plugin-python/actions/runs/9863829285/job/27237445507?pr=781
> Cache restored from key: ni-measurement-plugin-sdk-service-all-extras-Linux-py3.9.[19](https://github.com/ni/measurement-plugin-python/actions/runs/9863829285/job/27237445507?pr=781#step:6:20)-

packages/service/_docs_source/conf.py
dixonjoel: I don't know much about ReadTheDocs - I though you created the project for it and the configuration. I'll look into these. I'm trying to remember why I moved the _docs_source to this level - something failed on the PR.

What is the /.readthedocs.yml for? I'm not sure what it affects.

.github/workflows/check_nims.yml
dixonjoel: When I run it locally, it is putting them in /packages/service/docs. But when I left it as 'docs/', it gave an error that there was no docs/ directory. Maybe because there's not originally any source in GitHub at /packages/service/docs? I'll try again with docs/. Perhaps I need to make it just call 'rm -rf docs' again instead?

.github/workflows/check_nims.yml
dixonjoel: So we change it to `hashFiles('packages/service/poetry.lock')`? I'll try that. Or is there some other 'working-directory' type property on the cache action?

packages/service/scripts/install_examples.py
dixonjoel: Updated and tested locally.

.github/workflows/run_system_tests.yml
dixonjoel: Removed

.github/workflows/check_nims.yml
dixonjoel: Yeah. Here's the error. I guess I'll try 'rm -rf docs'
image

packages/service/_docs_source/conf.py
dixonjoel: Would you feel ok if we follow up and fix the docs after this checkin? I created a new project in readthedocs and we can update it to also remove 'measurementlink'
https://readthedocs.org/projects/measurement-plugin-python/

.github/workflows/check_nims.yml
dixonjoel: That worked.

.github/workflows/run_unit_tests.yml
```diff
@@ -40,6 +40,7 @@ jobs:
           key: ni-measurement-plugin-sdk-service-no-extras-${{ runner.os }}-py${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('poetry.lock') }}
```
bkeryan, line 40: This is still looking for poetry.lock in the root of the repo.

docs/CNAME
bkeryan, line 1: FYI: this is used to make GitHub Pages redirect to ReadTheDocs. You can move this to a less prominent location in the follow-up PR, and update the GitHub repo's settings:

image

dixonjoel:
@bkeryan I looked over all the caching steps to ensure they had valid keys.

bkeryan (approved):

mshafer-NI (approved):

PR #784, opened 2024-07-10
Add .serviceignore files to the examples

nick-beer:

This PR adds `.serviceignore` files to each of the examples in preparation of the Plug-in Library release.

When publishing a plugin to a Plug-in Library, we should exclude the `.venv` folder and the `__pycache__` folder if they are present.  The presence of the `.serviceignore` file will instruct the Plug-in Library CLI to ignore these folders.

Verified that when publishing one of the examples to a Plug-in Library, we ignore these two folders.

**Note**
This PR should not be completed prior to the upcoming release of Plug-in Library.

### DIFF
--- a/examples/game_of_life/.serviceignore
+++ b/examples/game_of_life/.serviceignore
@@ -1,0 +1,4 @@
+# This file specifies that when we publish this plugin to a Plug-in Library,
+# we should not copy the following directories:
+.venv
+__pycache__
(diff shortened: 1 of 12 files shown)


dixonjoel:
@nick-beer @bkeryan Since the repo was greatly reorganized yesterday, I'm going to make another prerelease. If we don't want this in the prerelease or the final release, I'll make them both out of the releases/2.0 branch and not cherry-pick this PR until we're ready for the final release. Any problems you see with that?

dixonjoel (approved):

bkeryan (approved):

PR #786, opened 2024-07-11
Fix ReadTheDocs build by including working directory for post-install

dixonjoel:

ReadTheDocs build is failing because it is not finding pyproject.toml at the root directory:
https://readthedocs.org/projects/measurement-plugin-python/builds/24965585/

Fix the docs build

None. Will look at the RTD build after submission.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,7 +10,7 @@ build:
       # Tell poetry to not use a virtual environment
       - poetry config virtualenvs.create false
     post_install:
-      - pushd packages/service/ && poetry install --with docs
+      - poetry -C packages/service/ install --with docs
 
 sphinx:
   configuration: packages/service/_docs_source/conf.py
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -11,6 +11,7 @@ build:
       - poetry config virtualenvs.create false
     post_install:
       - poetry install --with docs
+      working-directory: packages/service
 
 sphinx:
   configuration: packages/service/_docs_source/conf.py
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,10 +7,8 @@ build:
   jobs:
     post_create_environment:
       - pip install poetry==1.8.2
-      # Tell poetry to not use a virtual environment
-      - poetry config virtualenvs.create false
     post_install:
-      - poetry -C packages/service/ install --with docs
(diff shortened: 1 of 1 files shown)


.readthedocs.yml
```diff
@@ -11,6 +11,7 @@ build:
       - poetry config virtualenvs.create false
     post_install:
       - poetry install --with docs
+      working-directory: packages/service
```
bkeryan, line 14: GitHub and ReadTheDocs have separate YAML schemas. I see no reason to expect this to work. RTD uses underscores not dashes, and I don't see this key documented in https://docs.readthedocs.io/en/stable/config-file/v2.html

I would try `poetry -C packages/service/ install --with docs` or `pushd packages/service/ && poetry install --with docs`.

To test it, go to https://readthedocs.org/projects/measurement-plugin-python/versions/ and activate your dev branch.

bkeryan (changes requested):

.readthedocs.yml
bkeryan: poetry -C seems to work. Now the error is that Sphinx can't find `autoapi`.

It looks like RTD's Poetry instructions have changed since I set this up: https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-poetry

.readthedocs.yml
```diff
@@ -10,7 +10,7 @@ build:
       # Tell poetry to not use a virtual environment
       - poetry config virtualenvs.create false
```
bkeryan, line 11: Try deleting this

.readthedocs.yml
```diff
@@ -10,7 +10,7 @@ build:
...
     post_install:
-      - poetry install --with docs
+      - poetry -C packages/service/ install --with docs
```
bkeryan, line 13: And try adding this
```suggestion
      - VIRTUAL_ENV=$READTHEDOCS_VIRTUALENV_PATH poetry -C packages/service/ install --with docs
```

.readthedocs.yml
bkeryan: Looks like it worked: https://measurement-plugin-python.readthedocs.io/en/users-jdixon-fix-post-install-readthedocs/index.html

bkeryan (approved):

.readthedocs.yml
dixonjoel: Thanks, Brad!

PR #826, opened 2024-08-02
Add enumerate services API

Jotheeswaran-Nandagopal:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/measurement-services-python/blob/main/CONTRIBUTING.md).

### What does this Pull Request accomplish?

- Implementation of the enumerate service API, which return the active measurement services
- Added automation test

### Why should this Pull Request be merged?

- This implementation supports the Interactive Approach for the [Measurement Plug-In client API](https://dev.azure.com/ni/DevCentral/_git/ASW?path=/Specs/Test%20Automation%20Frameworks/E2719769_Support%20for%20Measurement%20Plug-In%20client%20APIs/E2719769_Support%20for%20Measurement%20Plug-In%20client%20APIs.md&version=GBmain&_a=preview&anchor=interactive-approach)'s in python 

### What testing has been done?

Existing and new automated tests passed.

PR #827, opened 2024-08-02
Update measurement service to get version from .serviceconfig

nick-beer:

Add support for reading a measurement's version from its `.serviceconfig` file, as well as passing that version through discovery service's `RegisterService` API.

To do this:
- Make `version` and `ui_file_paths` parameters to the `MeasurementService` constructor optional
- If version is specified in the constructor, use it unless it is also specified in the `.serviceconfig` file AND the two values are different. If the values are different, raise an error
- Add a `version` property to `ServiceInfo`
- Pass `ServiceInfo.version` through the `RegisterServiceRequest` inside the `DiscoveryClient`.

We're working to add version support to our plugins.  The intention is for the version of a plugin to be stored in its `.serviceconfig` file and passed to discovery service when it registers itself.

A couple new tests have been written, and a couple existing tests have been updated to account for the presence of a version in the service config.

### DIFF
--- a/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
@@ -241,7 +241,7 @@ class MeasurementService:
             service_class=service["serviceClass"],
             description_url=service["descriptionUrl"],
             provided_interfaces=service["providedInterfaces"],
-            version=version,
+            versions=[version],
             annotations={
                 key: convert_value_to_str(value)
                 for key, value in service.get("annotations", {}).items()
--- a/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
@@ -177,8 +177,8 @@ class MeasurementService:
     def __init__(
         self,
         service_config_path: Path,
-        version: str,
-        ui_file_paths: List[Path],
+        version: Optional[str] = None,
+        ui_file_paths: List[Path] = [],
         service_class: Optional[str] = None,
     ) -> None:
         """Initialize the Measurement Service object.
@@ -215,6 +215,14 @@ class MeasurementService:
                 f"Service class '{service_class}' not found in '{service_config_file}'"
             )
 
+        config_version = service.get("version")
+        if config_version is not None:
+            if version is not None and version != config_version:
+                raise RuntimeError(
+                    f"Version mismatch: .serviceconfig version is '{config_version}', but version parameter is '{version}'.",
+                )
+            version = config_version
+
         # Note: sphinx-autoapi uses the public attributes' type hints and docstrings.
         self.measurement_info: MeasurementInfo = MeasurementInfo(
             display_name=service["displayName"],
@@ -233,6 +241,7 @@ class MeasurementService:
             service_class=service["serviceClass"],
             description_url=service["descriptionUrl"],
             provided_interfaces=service["providedInterfaces"],
+            version=version,
             annotations={
                 key: convert_value_to_str(value)
                 for key, value in service.get("annotations", {}).items()
--- a/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
@@ -189,9 +189,12 @@ class MeasurementService:
         Args:
             service_config_path (Path): Path to the .serviceconfig file.
 
-            version (str): Version of the measurement service.
+            version (str): Version of the measurement service. Do not use this
+                parameter.  Instead, specify the "version" field in the
+                .serviceconfig file. Default value is "".
 
             ui_file_paths (List[Path]): List of paths to supported UIs.
+                Default value is [].
 
             service_class (str): The service class from the .serviceconfig to use.
                 Default value is None, which will use the first service in the
--- a/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
@@ -177,7 +177,7 @@ class MeasurementService:
     def __init__(
         self,
         service_config_path: Path,
-        version: Optional[str] = None,
+        version: str = "",
         ui_file_paths: List[Path] = [],
         service_class: Optional[str] = None,
     ) -> None:
@@ -217,7 +217,7 @@ class MeasurementService:
 
         config_version = service.get("version")
         if config_version is not None:
-            if version is not None and version != config_version:
+            if version and version != config_version:
                 raise RuntimeError(
                     f"Version mismatch: .serviceconfig version is '{config_version}', but version parameter is '{version}'.",
                 )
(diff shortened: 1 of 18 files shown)


packages/service/tests/unit/test_discovery_client.py
```diff
@@ -393,3 +394,7 @@ def _assert_service_info_equal(
     assert set(expected.provided_interfaces) == set(actual.provided_interfaces)
     assert expected.service_class == actual.service_class
     assert expected.annotations == actual.annotations
+    if isinstance(actual, GrpcServiceDescriptor):
+        assert expected.version == actual.versions[0]
+    else:
+        assert expected.version == actual.version
```
bkeryan, line 400: This highlights an impending conflict with #826. You are adding a singular `version` field to `ServiceInfo`, but @Jotheeswaran-Nandagopal is adding support for `enumerate_services`. According to the API docs, `EnumerateServicesResponse` contains a list of versions.

```
  // This field is used differently based on the message in which it is used.
  //
  // When used in a RegisterServiceRequest, this field must contain a single
  // version, and that version must follow the major.minor.patch format.
  // The version 0.0.1 is reserved for internal use, and cannot be specified.
  // If no version is specified, the version 0.0.1 will be used internally
  // for the registered service.
  //
  // When returned from EnumerateServicesResponse, this field will contain the
  // list of versions that are available for the associated service.
  //
  // When returned from ResolveServiceWithInformationResponse, this field will
  // contain the version of the resolved service.
  repeated string versions = 6;
```

bkeryan (changes requested):

packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
```diff
@@ -233,6 +241,7 @@ def convert_value_to_str(value: object) -> str:
             service_class=service["serviceClass"],
             description_url=service["descriptionUrl"],
             provided_interfaces=service["providedInterfaces"],
+            version=version,
```
bkeryan, line 244: `MeasurementInfo` and `ServiceInfo` expect a `str` but you are passing an `Optional[str]`.

```
ni_measurement_plugin_sdk_service/measurement/service.py:229: error: Argument "version" to "MeasurementInfo" has incompatible type "Optional[str]"; expected "str"  [arg-type]
ni_measurement_plugin_sdk_service/measurement/service.py:244: error: Argument "version" to "ServiceInfo" has incompatible type "Optional[str]"; expected "str"  [arg-type]
Found 2 errors in 1 file (checked 204 source files)
```

Some ways to handle this:
- Use the default argument `version: str = ""` and rely on "falsiness" instead of checking for None.
- Add a fallback case like `if version is None: version = ""`. I'm not sure if mypy will treat this as a type narrowing, though.
- `version="" if version is None else version,`

packages/service/tests/unit/test_discovery_client.py
nick-beer: Sorry - I reset you without seeing this message.  Thanks for pointing this out - I think it makes sense for ServiceInfo to have `versions` rather than `version`.  I'll make the update.

jayaseelan-james:
@nick-beer 
[Question] I'm just curious to know that with this change, will we update the template in our measurement-plugin-generator to have a version field in the .serviceconfig going forward?

nick-beer:
> @nick-beer [Question] I'm just curious to know that with this change, will we update the template in our measurement-plugin-generator to have a version field in the .serviceconfig going forward?

Good catch! Yes, this PR has been updated to include the necessary changes to the generator.

packages/service/tests/unit/test_discovery_client.py
```diff
@@ -393,3 +394,7 @@ def _assert_service_info_equal(
     assert set(expected.provided_interfaces) == set(actual.provided_interfaces)
     assert expected.service_class == actual.service_class
     assert expected.annotations == actual.annotations
+    if isinstance(actual, GrpcServiceDescriptor):
```
dixonjoel, line 397: Why the if since the asserts are the same?

dixonjoel (changes requested):
I'd like to see the changes with multiple versions.

packages/service/tests/unit/test_discovery_client.py
nick-beer: sloppiness - they didn't used to be the same, and I forgot to update the check.  Updated now.

nick-beer:
> I'd like to see the changes with multiple versions.

@dixonjoel - can you explain what you mean by this?  I don't follow.

dixonjoel:
@nick-beer "Sorry - I reset you without seeing this message. Thanks for pointing this out - I think it makes sense for ServiceInfo to have versions rather than version. I'll make the update." Maybe I misunderstood this. You resolved this conversation. What changed?

dixonjoel:
@nick-beer Oh, this update? Sorry I missed this.
https://github.com/ni/measurement-plugin-python/pull/827/commits/f438446ea0fa39aa71fe2e67a71fb6082dc41014

dixonjoel (approved):

nick-beer:
> @nick-beer Oh, this update? Sorry I missed this. [f438446](https://github.com/ni/measurement-plugin-python/commit/f438446ea0fa39aa71fe2e67a71fb6082dc41014)

@dixonjoel - yeah, that's the one that updated `ServiceInfo` to have the possibility of multiple versions.

packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
```diff
@@ -189,9 +189,12 @@ def __init__(
         Args:
             service_config_path (Path): Path to the .serviceconfig file.
 
-            version (str): Version of the measurement service.
+            version (str): Version of the measurement service. Do not use this
```
bkeryan, line 192: There is a mechanism we can use to tell clients not to use this argument: `warnings.warn("The MeasurementService 'version' parameter is deprecated. Instead, specify the 'version' field in the .serviceconfig file.", DeprecationWarning)`.

I don't know if I like emitting a warning that 100% of users will run into, but I guess that's what `PendingDeprecationWarning` is for.

bkeryan (approved):
Approved with suggestion

packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
nick-beer: Yeah - I'm inclined not to warn for now - at least until we can come up with an automatic mechanism for bumping your version in the .serviceconfig file.  I agree that it's too noisy for too little (initial) benefit.

For .NET, my plan is to generate the `.serviceconfig` file from the build process via a set of msbuild properties/items.  I'm not sure if something similar for python exists?  If so, then it becomes compelling to put the version in the `.serviceconfig` file rather than the `measurement.py` file.  As things stand now, you have to manually update one or the other.

The only significant benefit (at this point) to the .sericeconfig file is that if you don't put the version in the .serviceconfig file, the Plug-in Library CLI will prevent you from publishing the service to a Plug-in Library.

bkeryan: Ok, the current PR is fine.

PR #676, opened 2024-03-25
example: update nidcpower measurement to support multiple instrument sessions

Jayaseelan-James:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/measurement-services-python/blob/main/CONTRIBUTING.md).

### What does this Pull Request accomplish?
Updates the `nidcpower_source_dc_voltage` measurement to support multiple instrument sessions. 

### Why should this Pull Request be merged?
Fixes #673.

### What testing has been done?
- Verified the measurement execution in InstrumentStudio using individual site values and "All Sites".
- Verified the measurement execution in TestStand using sequential and batch process models.

PR #842, opened 2024-08-14
example: use the tuple with the updated "in_compliance" value returned by _replace()

jayaseelan-james:

* While expanding the `nidcpower_source_dc_voltage` measurement to support multisite in #676, I overlooked the update to the `in_compliance` value in the `_Measurement` tuple. Since tuples are immutable, invoking `_replace()` on a tuple doesn't modify the original tuple; instead, it returns a new tuple with the updated values.
* In the current implementation, the object returned by the `_replace()` method wasn't captured or assigned to any variable, resulting in the use of the original tuple, which did not have the correct "in_compliance" value. This PR ensures that the tuple returned by the `_replace()` method is properly appended to the final `measurements` list.

* Fixes the bug where the `in_compliance` value was not assigned properly.

* Manually verified by running the measurement.

### DIFF
--- a/examples/nidcpower_source_dc_voltage/measurement.py
+++ b/examples/nidcpower_source_dc_voltage/measurement.py
@@ -123,9 +123,7 @@ def measure(
                         in_compliance = session_info.session.channels[
                             channel_mapping.channel
                         ].query_in_compliance()
-                        measurement._replace(in_compliance=in_compliance)
-
-                    measurements.extend(session_measurements)
+                        measurements.append(measurement._replace(in_compliance=in_compliance))
 
                 # Reset the channels to a known state
                 for session_info in session_infos:


examples/nidcpower_source_dc_voltage/measurement.py
bkeryan, line 1: > in the right way

> did not update the measurement tuple appropriately

> is updated properly

Before you merge this, please update the PR description to describe what was wrong or inappropriate and what is right/proper.

Hint: you probably need words like "immutable", "mutate", and "assign"

bkeryan (approved):

examples/nidcpower_source_dc_voltage/measurement.py
jayaseelan-james: I didn't choose the best words to describe the issue earlier. Thank you for pointing that out. I've now updated the description.

PR #854, opened 2024-08-29
Add Pin Map Client to Measurement PlugIn SDK Service

MounikaBattu17:

As we need pin map client in Measurement Client, adding Pin Map Client to Measurement PlugIn SDK Service.

- Added new module for pin map client.
- Moved `pin_map_client.py` from `test_utilities` to the newly created module.
- Updated tests to use the updated imports.
- Added automated tests for `PinMapClient`.

Added automated tests.


packages/service/tests/assets/unit/pin_map/1Smu1ChannelGroup1Pin1Site.pinmap
MounikaBattu17, line 1: This sample pin map file has been added for unit testing and can be used for testing any future APIs in the pin map client.

packages/service/tests/unit/test_pin_map_client.py
```diff
@@ -0,0 +1,80 @@
...
+)
+from ni_measurement_plugin_sdk_service._internal.stubs.ni.measurementlink.pinmap.v1.pin_map_service_pb2_grpc import (
+    PinMapServiceStub,
+)
+from ni_measurement_plugin_sdk_service.discovery import DiscoveryClient
+from ni_measurement_plugin_sdk_service.grpc.channelpool import GrpcChannelPool
+from ni_measurement_plugin_sdk_service.pin_map._client import PinMapClient
+
+
+def test__valid_pin_map_file__register_pin_map___sends_request_and_returns_id(
```
jayaseelan-james, line 24: The "sends request" is the act and should not be quoted under the assert phrase.
```suggestion
def test___valid_pin_map_file___register_pin_map___returns_pin_map_id(
```

packages/service/tests/unit/test_pin_map_client.py
```diff
@@ -0,0 +1,80 @@
...
+    registered_pin_map_id = pin_map_client.update_pin_map(pin_map_path)
+
+    pin_map_stub.UpdatePinMapFromXml.assert_called_once()
+    request: UpdatePinMapFromXmlRequest = pin_map_stub.UpdatePinMapFromXml.call_args.args[0]
+    assert request.pin_map_id == pin_map_path
+    assert request.pin_map_xml == _get_pin_map_file_contents(pin_map_path)
+    assert registered_pin_map_id == pin_map_path
+
+
+def test__invalid_pin_map_file_path__register_pin_map___raises_file_not_found_error(
```
jayaseelan-james, line 41:
```suggestion
def test___invalid_pin_map_file_path___register_pin_map___raises_file_not_found_error(
```

packages/service/ni_measurement_plugin_sdk_service/pin_map/_client.py
```diff
@@ -30,7 +30,15 @@ def __init__(
         grpc_channel: Optional[grpc.Channel] = None,
         grpc_channel_pool: Optional[GrpcChannelPool] = None
     ) -> None:
-        """Initialize the pin map client."""
+        """Initialize the pin map client.
+
+        Args:
+            discovery_client: An optional discovery client (recommended).
+
+            grpc_channel: An optional session management gRPC channel.
```
bkeryan, line 38:
```suggestion
            grpc_channel: An optional pin map gRPC channel.
```

packages/service/tests/unit/conftest.py
```diff
@@ -94,3 +95,9 @@ def session_management_client(
 def single_session_reservation(mocker: MockerFixture) -> Mock:
     """Test fixture that creates a mock SingleSessionReservation."""
     return mocker.create_autospec(SingleSessionReservation)
+
+
+@pytest.fixture(scope="module")
```
bkeryan, line 100: Let's stop adding scope=module to new fixtures. If something is worth caching between tests, it's usually worth caching between tests in multiple modules. 

These directory-name fixtures are lightweight, so it's not very important to cache them. The default of scope=function is fine unless there are wider-scoped fixtures that depend on them.

packages/service/ni_measurement_plugin_sdk_service/pin_map/_client.py
```diff
@@ -72,9 +80,10 @@ def update_pin_map(self, pin_map_path: Union[str, pathlib.Path]) -> str:
             The resource id of the pin map that is registered to the pin map service.
         """
         # By convention, the pin map id is the .pinmap file path.
+        with open(pin_map_path, "r", encoding="utf-8-sig") as file:
+            xml_content = file.read()
         request = pin_map_service_pb2.UpdatePinMapFromXmlRequest(
-            pin_map_id=str(pin_map_path),
-            pin_map_xml=pathlib.Path(pin_map_path).read_text(encoding="utf-8"),
+            pin_map_id=str(pin_map_path), pin_map_xml=xml_content
```
bkeryan, line 86: Why replace read_text with open+read? You can pass utf-8-sig to read_text.

packages/service/tests/unit/test_pin_map_client.py
```diff
@@ -0,0 +1,80 @@
...
+    pin_map_client: PinMapClient,
+) -> None:
+    pin_map_path = "InvalidPinmap.pinmap"
```
bkeryan, line 44: nitpick: consistency
```suggestion
    pin_map_path = "InvalidPinMap.pinmap"
```

bkeryan (changes requested):

packages/service/tests/unit/test_pin_map_client.py
bkeryan, line 24: Test names should use 3 underscores as a separator, not 2

packages/service/tests/unit/test_pin_map_client.py
MounikaBattu17: Updated the test case name.

packages/service/tests/unit/conftest.py
MounikaBattu17: Sure, removed the scope as module.

packages/service/ni_measurement_plugin_sdk_service/pin_map/_client.py
MounikaBattu17: I think I overlooked this case. I went with context manager purely for resources management. But seems like `pathlib.path` internally handles the file open and close operations. Hence, I have reverted back the context manager change.

dixonjoel (approved):

bkeryan (approved):

PR #856, opened 2024-08-29
Add stream_measure API in Measurement Plug-In Client

Jotheeswaran-Nandagopal:

- Added stream-measure API support to Measurement Plug-In Client.
- Added Test cases for stream-measure API.

- Stream-measure API supports user to interact with streaming measurement data services.

- Tested manually.
- Executed existing and new automated tests.


packages/generator/tests/acceptance/test_non_streaming_measurement_client.py
```diff
@@ -33,19 +36,45 @@ def test___measurement_plugin_client___measure___returns_output(
...
+        io_array_out=["resource1", "resource2"],
+        integer_out=10,
+        xy_data_out=None,
+    )
+    measurement_plugin_client = test_measurement_client_type()
+
+    measure_output = measurement_plugin_client.stream_measure()
+
+    for output in measure_output:
+        assert output == expected_measure_output
```
bkeryan, line 62: The non-streaming measurement should return exactly one response. This will pass if the stream returns too few or too many responses.

Do something like this:
```
    responses = [response for response in response_iterator]
    assert len(responses) == 1
    assert responses[0] == expected_measure_output
```

packages/generator/tests/acceptance/test_non_streaming_measurement_client.py
```diff
@@ -33,19 +36,45 @@ def test___measurement_plugin_client___measure___returns_output(
     assert measure_output == expected_measure_output
 
 
+def test___measurement_plugin_client___stream_measure___returns_output(
+    measurement_plugin_client_module: ModuleType,
+) -> None:
+    test_measurement_client_type = getattr(measurement_plugin_client_module, "TestMeasurement")
+    output_type = getattr(measurement_plugin_client_module, "Output")
+    expected_measure_output = output_type(
```
bkeryan, line 44: Rename to `expected_output`. You originally used `measure_output` vs `output` to disambiguate between the output type and output instance.

packages/generator/tests/acceptance/test_non_streaming_measurement_client.py
```diff
@@ -33,19 +36,45 @@ def test___measurement_plugin_client___measure___returns_output(
...
+        path_out="path/test",
+        path_array_out=["path/test1", "path/ntest2"],
+        io_out="resource",
+        io_array_out=["resource1", "resource2"],
+        integer_out=10,
+        xy_data_out=None,
+    )
+    measurement_plugin_client = test_measurement_client_type()
+
+    measure_output = measurement_plugin_client.stream_measure()
```
bkeryan, line 59: Rename to `output_iterator` or `response_iterator`

packages/generator/tests/acceptance/test_streaming_measurement_client.py
```diff
@@ -0,0 +1,95 @@
...
+
+    measure_output = measurement_plugin_client.stream_measure()
+
+    for count, output in enumerate(measure_output, start=0):
+        expected_measure_output = output_type(
+            name="<Name>",
+            index=count,
+            data=[i for i in range(count + 1)],
+        )
+        assert output == expected_measure_output
```
bkeryan, line 46: Asserting in a loop loses information: which iteration failed?

Instead, build a list of expected outputs and a list of actual outputs, and compare them.

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -163,3 +168,29 @@ class ${class_name}:
                 result[k - 1] = v
 
         return Output._make(result)
+
+    def stream_measure(
+        self,
+        ${configuration_parameters_with_type_and_default_values}
+    ) -> Generator[${output_type}, None, None] :
```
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: Thanks for pointing that out. We tried adding ReturnType to the game_of_life measurement, and as you said, the stream-measure API is now getting the return response correctly.

packages/generator/tests/acceptance/test_non_streaming_measurement_client.py
Jotheeswaran-Nandagopal: Renamed to `response_iterator`

packages/generator/tests/acceptance/test_non_streaming_measurement_client.py
Jotheeswaran-Nandagopal: Sure, renamed to `expected_output`.

packages/generator/tests/acceptance/test_non_streaming_measurement_client.py
Jotheeswaran-Nandagopal: Sure, updated as suggested.

packages/generator/tests/acceptance/test_streaming_measurement_client.py
```diff
@@ -0,0 +1,95 @@
...
+import pathlib
+from types import ModuleType
+from typing import Generator
+
+import pytest
+from ni_measurement_plugin_sdk_service.measurement.service import MeasurementService
+
+from ni_measurement_plugin_sdk_generator.client import create_client
+from tests.utilities.discovery_service_process import DiscoveryServiceProcess
+from tests.utilities.measurements import streaming_data_measurement
```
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: I didn't add import-paths, instead, used py.typed for that and I was re-using the assets. 

Now,
- I have removed the py.typed file and added the measurements required for generator and deleted the measurement measurement added in packages/service/tests/assets.
- I have also added __init__.py in  packages/generator/tests.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -163,3 +168,29 @@ class ${class_name}:
...
+            % else:
+            result = []
+            % endif
+            output_values = deserialize_parameters(
+                self._output_metadata, response.outputs.value, f"{self._service_class}.Outputs"
+            )
+
+            for k, v in output_values.items():
+                result[k - 1] = v
+            yield Output._make(result)
```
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: I have moved this to a method.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -138,16 +152,7 @@ class ${class_name}:
...
-                parameter_values=parameter_values, 
-                service_name=f"{self._service_class}.Configurations"
-            )
-        )
-        request = v2_measurement_service_pb2.MeasureRequest(
-            configuration_parameters=serialized_configuration
-        )
+        request = self._get_measure_request(parameter_values)
         % if output_metadata:
         result = [None] * max(self._output_metadata.keys())
```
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: Sure, I have created a method to de-serialize the response, as suggested in the [below comment](https://github.com/ni/measurement-plugin-python/pull/856#discussion_r1736650767). So, `result` is initialized on every iteration.

packages/generator/tests/acceptance/test_streaming_measurement_client.py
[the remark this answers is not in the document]
bkeryan: You did add import paths. You just didn't know it.
- `pyproject.toml` is installing `ni-measurement-plugin-sdk-service` from `../../packages/service`.
- Each package is using "flat layout": https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/
- Because `ni-measurement-plugin-sdk-service` uses flat layout, this adds `../../packages/service` to the import path. (If it used src layout, I think it would add `../../packages/service/src`.)
- The tests are also in this directory, so they also get added to the import path.

This will break when we create the next prerelease and comment out the ../../ stuff in `pyproject.toml`, because it will use `.venv/lib/site-packages/ni-measurement-plugin-sdk-service` rather than adding `../../packages/service` to the import path.

I agree with adding test assets/utilities to `packages/generator/tests` and removing `py.typed` from the tests directory.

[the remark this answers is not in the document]
bkeryan: This is also because you set `develop = true`. Removing this makes the tests behave like they should:

```
PS D:\dev\measurement-plugin-python\packages\generator> poetry run pytest -v
ImportError while loading conftest 'D:\dev\measurement-plugin-python\packages\generator\tests\conftest.py'.
tests\conftest.py:11: in <module>
    from tests.utilities.discovery_service_process import DiscoveryServiceProcess
E   ModuleNotFoundError: No module named 'tests.utilities'
```

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -128,38 +128,63 @@ class ${class_name}:
...
+        self, response: v2_measurement_service_pb2.MeasureResponse
+    ) -> Output:
         % if output_metadata:
         result = [None] * max(self._output_metadata.keys())
         % else:
         result = []
         % endif
+        output_values = deserialize_parameters(
+            self._output_metadata, response.outputs.value, f"{self._service_class}.Outputs"
+        )
```
bkeryan, line 155: This doesn't have to be codegen, does it?

```suggestion
        if self._output_metadata:
            result = [None] * max(self._output_metadata.keys())
        else:
            result = []
   
        output_values = deserialize_parameters(
            self._output_metadata, response.outputs.value, f"{self._service_class}.Outputs"
        )
```

bkeryan (approved):
Approved with suggestions:
- Remove `develop = true` and run `poetry lock --no-update`
- Change `if output_metadata` conditional to occur at run time

packages/generator/tests/acceptance/test_streaming_measurement_client.py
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: Sure, I got it. Thanks. I have removed `develop = true`.

PR #870, opened 2024-09-02
Add Cancel API to Measurement Plug-In Client

Jotheeswaran-Nandagopal:

- Added Cancel API support to Measurement Plug-In Client.
- Added Test cases for Cancel API.

- Cancel API helps user to cancel the measure and stream-measure gRPC calls.

- Tested manually.
- Executed existing and new automated tests.

### DIFF
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -67,6 +67,7 @@ class ${class_name}:
         self._grpc_channel_pool = grpc_channel_pool
         self._discovery_client = discovery_client
         self._stub: Optional[v2_measurement_service_pb2_grpc.MeasurementServiceStub] = None
+        self._measure_response: Optional[v2_measurement_service_pb2.MeasureResponse] = None
         self._configuration_metadata = ${configuration_metadata}
         self._output_metadata = ${output_metadata}
         if grpc_channel is not None:
@@ -168,10 +169,17 @@ class ${class_name}:
         """
         parameter_values = [${measure_api_parameters}]
         request = self._create_measure_request(parameter_values)
-
-        for response in self._get_stub().Measure(request):
-            result = self._deserialize_response(response)
-        return result
+        self._measure_response = self._get_stub().Measure(request)
+        
+        try:
+            for response in self._measure_response:
+                result = self._deserialize_response(response)
+            return result
+        except grpc.RpcError as e:
+            if e.code() == grpc.StatusCode.CANCELLED:
+                print("Measure call has been cancelled.")
+            else:
+                raise           
 
     def stream_measure(
         self,
@@ -184,6 +192,17 @@ class ${class_name}:
         """
         parameter_values = [${measure_api_parameters}]
         request = self._create_measure_request(parameter_values)
+        self._measure_response = self._get_stub().Measure(request)
 
-        for response in self._get_stub().Measure(request):
-            yield self._deserialize_response(response)
+        try:
+            for response in self._measure_response:
+                yield self._deserialize_response(response)  
+        except grpc.RpcError as e:
+            if e.code() == grpc.StatusCode.CANCELLED:
+                print("Measure call has been cancelled.")
+            else:
+                raise      
+
+    def cancel(self) -> None:
+        if self._measure_response and self._measure_response.is_active():
+            self._measure_response.cancel()
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -172,7 +172,7 @@ class ${class_name}:
         )
         for response in stream_measure_response:
             result = response
-        return result          
+        return result
 
     def stream_measure(
         self,
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -143,6 +143,7 @@ class ${class_name}:
             configuration_parameters=serialized_configuration
         )
 
+    % if output_metadata:
     def _deserialize_response(
         self, response: v2_measurement_service_pb2.MeasureResponse
     ) -> Outputs:
@@ -158,6 +159,7 @@ class ${class_name}:
             result[k - 1] = v
         return Outputs._make(result)
 
+    % endif
     def measure(
         self,
         ${configuration_parameters_with_type_and_default_values}
@@ -167,12 +169,20 @@ class ${class_name}:
         Returns:
             Measurement outputs.
         """
-        stream_measure_response = self.stream_measure(
-            ${measure_api_parameters}
-        )
+        with self._initialization_lock:
+            if self._measure_response is not None:
+                print("A measure call is already in progress.")
+                return
+            stream_measure_response = self.stream_measure(
+                ${measure_api_parameters}
+            )
         for response in stream_measure_response:
+        % if output_metadata:
             result = response
         return result
+        % else:
+            pass
+        % endif
 
     def stream_measure(
         self,
@@ -184,18 +194,31 @@ class ${class_name}:
             Stream of measurement outputs.
         """
         parameter_values = [${measure_api_parameters}]
-        request = self._create_measure_request(parameter_values)
-        self._measure_response = self._get_stub().Measure(request)
+        with self._initialization_lock:
+            if self._measure_response is not None:
+                print("A measure call is already in progress.")
+                return
+            request = self._create_measure_request(parameter_values)
+            self._measure_response = self._get_stub().Measure(request)
 
         try:
             for response in self._measure_response:
-                yield self._deserialize_response(response)  
+                % if output_metadata:
+                yield self._deserialize_response(response)
+                % else:
+                yield
+                % endif
         except grpc.RpcError as e:
             if e.code() == grpc.StatusCode.CANCELLED:
-                print("Measure call has been cancelled.")
+                print("The measure call is canceled.")
             else:
-                raise      
+                raise
+        finally:
+            self._measure_response = None
 
     def cancel(self) -> None:
         if self._measure_response and self._measure_response.is_active():
-            self._measure_response.cancel()
+            with self._initialization_lock:
+                self._measure_response.cancel()
+        else:
+            print("The measure call is not initialized at this moment.")
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -173,8 +173,8 @@ class ${class_name}:
         
         try:
             for response in self._measure_response:
-                result = self._deserialize_response(response)
-            return result
+                result = response
+            return self._deserialize_response(result)
         except grpc.RpcError as e:
             if e.code() == grpc.StatusCode.CANCELLED:
                 print("Measure call has been cancelled.")
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -192,8 +192,7 @@ class ${class_name}:
         parameter_values = [${measure_api_parameters}]
         with self._initialization_lock:
             if self._measure_response is not None:
-                print("A measure call is already in progress.")
-                return
+                raise RuntimeError("A measure call is already in progress.")
             request = self._create_measure_request(parameter_values)
             self._measure_response = self._get_stub().Measure(request)
 
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -174,9 +174,6 @@ class ${class_name}:
         stream_measure_response = self.stream_measure(
             ${measure_api_parameters}
         )
-        % if output_metadata:
-        result = None
-        % endif
         for response in stream_measure_response:
         % if output_metadata:
             result = response
@@ -197,7 +194,9 @@ class ${class_name}:
         parameter_values = [${measure_api_parameters}]
         with self._initialization_lock:
             if self._measure_response is not None:
-                raise RuntimeError("A measure call is already in progress.")
+                raise RuntimeError(
+                    "A measurement is currently in progress. To make concurrent measurement requests, please create a new client instance."
+                )
             request = self._create_measure_request(parameter_values)
             self._measure_response = self._get_stub().Measure(request)
 
@@ -210,9 +209,8 @@ class ${class_name}:
                 % endif
         except grpc.RpcError as e:
             if e.code() == grpc.StatusCode.CANCELLED:
-                _logger.debug("The measure call is canceled.")
-            else:
-                raise
+                _logger.debug("The measurement is canceled.")
+            raise
         finally:
             with self._initialization_lock:
                 self._measure_response = None
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -67,7 +67,9 @@ class ${class_name}:
         self._grpc_channel_pool = grpc_channel_pool
         self._discovery_client = discovery_client
         self._stub: Optional[v2_measurement_service_pb2_grpc.MeasurementServiceStub] = None
-        self._measure_response: Optional[v2_measurement_service_pb2.MeasureResponse] = None
+        self._measure_response: Optional[
+            Generator[v2_measurement_service_pb2.MeasureResponse, None, None]
+        ] = None
         self._configuration_metadata = ${configuration_metadata}
         self._output_metadata = ${output_metadata}
         if grpc_channel is not None:
@@ -172,6 +174,9 @@ class ${class_name}:
         stream_measure_response = self.stream_measure(
             ${measure_api_parameters}
         )
+        % if output_metadata:
+        result = None
+        % endif
         for response in stream_measure_response:
         % if output_metadata:
             result = response
@@ -205,15 +210,17 @@ class ${class_name}:
                 % endif
         except grpc.RpcError as e:
             if e.code() == grpc.StatusCode.CANCELLED:
-                print("The measure call is canceled.")
+                _logger.debug("The measure call is canceled.")
             else:
                 raise
         finally:
-            self._measure_response = None
+            with self._initialization_lock:
+                self._measure_response = None
 
     def cancel(self) -> bool:
-        if self._measure_response and self._measure_response.is_active():
-            self._measure_response.cancel()
-            return True
-        else:
-            return False
+        """Cancels the active measure call."""
+        with self._initialization_lock:
+            if self._measure_response:
+                return self._measure_response.cancel()
+            else:
+                return False
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -169,13 +169,9 @@ class ${class_name}:
         Returns:
             Measurement outputs.
         """
-        with self._initialization_lock:
-            if self._measure_response is not None:
-                print("A measure call is already in progress.")
-                return
-            stream_measure_response = self.stream_measure(
-                ${measure_api_parameters}
-            )
+        stream_measure_response = self.stream_measure(
+            ${measure_api_parameters}
+        )
         for response in stream_measure_response:
         % if output_metadata:
             result = response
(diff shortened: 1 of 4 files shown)


packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -168,10 +169,17 @@ class ${class_name}:
...
-
-        for response in self._get_stub().Measure(request):
-            result = self._deserialize_response(response)
-        return result
+        self._measure_response = self._get_stub().Measure(request)
+        
+        try:
+            for response in self._measure_response:
+                result = self._deserialize_response(response)
+            return result
```
jayaseelan-james, line 177: Can we consider deserializing the final response alone?
```suggestion
            for response in self._measure_response:
                result = response
            return self._deserialize_response(result)
```

jayaseelan-james (changes requested):

packages/generator/tests/acceptance/test_streaming_measurement_client.py
```diff
@@ -49,6 +51,44 @@ def test___measurement_plugin_client___stream_measure___returns_output(
...
+    time.sleep(2)
+
+    measurement_plugin_client.cancel()
+
+    measure_thread.join()
+    captured = capsys.readouterr()
+    assert "Measure call has been cancelled." in captured.out
+
+
+def test___measurement_plugin_client___cancel___cancels_stream_measure_call(
```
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: Sure, I have updated it.

dixonjoel (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -184,6 +185,17 @@ class ${class_name}:
...
+                yield self._deserialize_response(response)  
+        except grpc.RpcError as e:
+            if e.code() == grpc.StatusCode.CANCELLED:
+                print("Measure call has been cancelled.")
+            else:
+                raise      
+
+    def cancel(self) -> None:
+        if self._measure_response and self._measure_response.is_active():
+            self._measure_response.cancel()
```
jayaseelan-james, line 201: The measure response should be nullified upon cancellation and measurement completion.

jayaseelan-james (changes requested):
The client module should be updated to handle measurements with no outputs.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -184,6 +185,17 @@ class ${class_name}:
...
         parameter_values = [${measure_api_parameters}]
         request = self._create_measure_request(parameter_values)
-
-        for response in self._get_stub().Measure(request):
-            yield self._deserialize_response(response)
+        self._measure_response = self._get_stub().Measure(request)
+
+        try:
+            for response in self._measure_response:
+                yield self._deserialize_response(response)
```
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: Sure, Thanks for pointing this. I have updated it.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
Jotheeswaran-Nandagopal: Sure. I have nullified the measure response in finally block.

packages/generator/tests/acceptance/test_streaming_measurement_client.py
```diff
@@ -49,6 +51,73 @@ def test___measurement_plugin_client___stream_measure___returns_output(
...
+
+
+def test___streaming_measurement_execution___cancel___cancels_stream_measure_call(
+    measurement_plugin_client_module: ModuleType,
+    capsys: pytest.CaptureFixture[str],
+) -> None:
+    test_measurement_client_type = getattr(measurement_plugin_client_module, "TestMeasurement")
+    measurement_plugin_client = test_measurement_client_type()
+    measure_thread = threading.Thread(
+        target=lambda: list(map(lambda i: print(i), measurement_plugin_client.stream_measure()))
```
jayaseelan-james, line 95: Please simplify this if this is intended to iterate the response from stream_measure.
```suggestion
        target=tuple(measurement_plugin_client.stream_measure())
```

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -183,7 +194,31 @@ class ${class_name}:
...
+                % else:
+                yield
+                % endif
+        except grpc.RpcError as e:
+            if e.code() == grpc.StatusCode.CANCELLED:
+                print("The measure call is canceled.")
+            else:
+                raise
+        finally:
+            self._measure_response = None
```
jayaseelan-james, line 217: Similar to the `_get_stub` method. Can we also place this try...except block within the `initialization_lock`?

jayaseelan-james (changes requested):

packages/generator/tests/acceptance/test_streaming_measurement_client.py
Jotheeswaran-Nandagopal: Thanks for the suggestion, simplified with tuple. As we are calling the `stream_measure` in a thread, I have placed the tuple in `lambda`.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -183,7 +194,31 @@ class ${class_name}:
...
+
+    def cancel(self) -> None:
+        if self._measure_response and self._measure_response.is_active():
+            with self._initialization_lock:
+                self._measure_response.cancel()
+        else:
+            print("The measure call is not initialized at this moment.")
```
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: Updated to boolean.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
Jotheeswaran-Nandagopal: To restrict the users to avoid calling `measure` multiple time simultaneuosly from diffrent threads, placed this try...except block outside the `initialization_lock`.

packages/generator/tests/acceptance/test_streaming_measurement_client.py
```diff
@@ -49,6 +51,73 @@ def test___measurement_plugin_client___stream_measure___returns_output(
...
+def test___call_measure_in_differnet_threads___measure___print_process_in_progress_message(
+    measurement_plugin_client_module: ModuleType,
+    capsys: pytest.CaptureFixture[str],
+) -> None:
+    test_measurement_client_type = getattr(measurement_plugin_client_module, "TestMeasurement")
+    measurement_plugin_client = test_measurement_client_type()
+
+    measure_thread = threading.Thread(target=measurement_plugin_client.measure)
+    measure_thread.start()
+    measurement_plugin_client.measure()
```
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: If the second measure call is on worker thread, pytest is unable to capture the exception thrown from child thread. So, proceeding with the current implementation. 
Added a comment for this in the test.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -183,7 +190,30 @@ class ${class_name}:
...
+                print("The measure call is canceled.")
+            else:
+                raise
+        finally:
+            self._measure_response = None
+
+    def cancel(self) -> bool:
+        if self._measure_response and self._measure_response.is_active():
+            self._measure_response.cancel()
+            return True
```
bkeryan, line 217: This has two potential race conditions. 
- The RPC may complete between the call to `is_active()` and `cancel()`.
- `self._measure_response` may be set to `None` between the call to `is_active()` and `cancel()`.

Do you really need to call `is_active()`?

Does the underlying future return a bool? I think it does.

Try something like this:
```
response = self._measure_response
if response:
    return response.cancel()
return False
```

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -62,11 +62,12 @@ class ${class_name}:
 
             grpc_channel_pool: An optional gRPC channel pool.
         """
-        self._initialization_lock = threading.Lock()
+        self._initialization_lock = threading.RLock()
         self._service_class = "${service_class}"
         self._grpc_channel_pool = grpc_channel_pool
         self._discovery_client = discovery_client
         self._stub: Optional[v2_measurement_service_pb2_grpc.MeasurementServiceStub] = None
+        self._measure_response: Optional[v2_measurement_service_pb2.MeasureResponse] = None
```
bkeryan, line 70: This is not consistent with `self._measure_response = self._get_stub().Measure(request)`

In the v2 interface, `Measure` is a server-streaming RPC, so the stub's `Measure` method is actually a [UnaryStreamMultiCallable](https://github.com/shabbyrobe/grpc-stubs/blob/master/grpc-stubs/__init__.pyi#L674) object, which has a `__call__` method that returns a `CallIterator[TResponse]`. 

`MeasureResponse` is a gRPC message object, not an iterator, and Mypy shouuldn't let you do `for response in self._measure_response` when it's annotated as an `Optional[MeasureResponse]`.

Before you fix the type hint, please figure out why Mypy accepted this code with the incorrect type hint. Is it because we aren't running mypy on the codegen output?

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -183,7 +190,30 @@ class ${class_name}:
...
+        try:
+            for response in self._measure_response:
+                % if output_metadata:
+                yield self._deserialize_response(response)
+                % else:
+                yield
+                % endif
+        except grpc.RpcError as e:
+            if e.code() == grpc.StatusCode.CANCELLED:
+                print("The measure call is canceled.")
```
bkeryan, line 208: Don't print to stdout. This is library code, so it shouldn't write random stuff to stdout. Use the logger if you want to print a debug message.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -183,7 +190,30 @@ class ${class_name}:
...
+            else:
+                raise
+        finally:
+            self._measure_response = None
```
bkeryan, line 212: Reacquire the lock before modifying the variable protected by the lock.

packages/generator/tests/acceptance/test_streaming_measurement_client.py
```diff
@@ -49,6 +51,64 @@ def test___measurement_plugin_client___stream_measure___returns_output(
...
+
+
+def test___non_streaming_measurement_execution___cancel___returns_true(
+    measurement_plugin_client_module: ModuleType,
+) -> None:
+    test_measurement_client_type = getattr(measurement_plugin_client_module, "TestMeasurement")
+    measurement_plugin_client = test_measurement_client_type()
+    measure_thread = threading.Thread(target=measurement_plugin_client.measure)
+    measure_thread.start()
+    time.sleep(2)  # Wait for 2 seconds to call Cancel API.
```
bkeryan, line 76: Hardcoded delays in tests are discouraged because they contribute to both slowness and flakiness of the test suite.

You could write a test measurement that signals an event and then waits until canceled. Then the test can wait for the event before calling cancel().

packages/generator/tests/acceptance/test_streaming_measurement_client.py
```diff
@@ -49,6 +51,64 @@ def test___measurement_plugin_client___stream_measure___returns_output(
...
+def test___measurement_plugin_client___invoke_measure_from_two_threads___initiates_first_measure_and_rejects_second_measure(
+    measurement_plugin_client_module: ModuleType,
+) -> None:
+    test_measurement_client_type = getattr(measurement_plugin_client_module, "TestMeasurement")
+    measurement_plugin_client = test_measurement_client_type()
+
+    with pytest.raises(RuntimeError) as exc_info:
+        measure_thread = threading.Thread(target=measurement_plugin_client.measure)
+        measure_thread.start()
+        measurement_plugin_client.measure()  # Calling Measure in main thread, to catch error in pytest.
```
bkeryan, line 63: This makes it sound like there is a bug in pytest.

bkeryan, line 63: This is a race condition. There is no guarantee that the background thread calls measure first.

Actually it's two race conditions, if the call to measure() takes a finite amount of time.

You might be better off using concurrent.futures.ThreadPoolExecutor to make two concurrent calls to measure, then assert that one of them raises the exception.

packages/generator/tests/acceptance/test_streaming_measurement_client.py
```diff
@@ -49,6 +51,64 @@ def test___measurement_plugin_client___stream_measure___returns_output(
...
+    assert is_canceled
+
+
+def test___streaming_measurement_execution___cancel___returns_true(
+    measurement_plugin_client_module: ModuleType,
+) -> None:
+    test_measurement_client_type = getattr(measurement_plugin_client_module, "TestMeasurement")
+    measurement_plugin_client = test_measurement_client_type()
+    measure_thread = threading.Thread(
+        target=lambda: tuple(measurement_plugin_client.stream_measure())
```
bkeryan, line 90: Why convert the result to a tuple? Is that to iterate over it? I would use list for that.

This highlights the fact that threading.Thread throws away the result. Using concurrent.futures.ThreadPoolExecutor would allow you to inspect the returned values.

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -183,7 +190,30 @@ class ${class_name}:
...
+
+    def cancel(self) -> bool:
```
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: Added docstring.

packages/generator/tests/acceptance/test_streaming_measurement_client.py
Jotheeswaran-Nandagopal: Event is added in measurement already. This delay was given, to wait for the measure_response gRPC call to send to complete few iterations (as this is a streaming measurement) before cancelling, before cancelling. I see, that is not a best practice. So, removed the delay.

packages/generator/tests/acceptance/test_streaming_measurement_client.py
Jotheeswaran-Nandagopal: I have removed this and updated with `concurrent.futures.ThreadPoolExecutor`.

packages/generator/tests/acceptance/test_streaming_measurement_client.py
Jotheeswaran-Nandagopal: Updated with `concurrent.futures.ThreadPoolExecutor` and used list instead of tuple.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -39,7 +39,7 @@ class Output(NamedTuple):
     """Measurement result container."""
 
     ${output_parameters_with_type}
-<% output_type = "Output" %>\
+<% output_type = "Optional[Output]" %>\
```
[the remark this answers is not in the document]
bkeryan: This reduces usability because it requires callers to check whether the return value is None. Also, it highlights the fact that stream_measure doesn't allow the caller to determine whether the call was canceled or not.

Perhaps it would be better to propagate gRPC CANCELLED exceptions to the caller rather than trying to sweep them under the rug?

How will the LabVIEW client handle this? Does it indicate whether the call was canceled?

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -183,7 +195,32 @@ class ${class_name}:
             Stream of measurement output.
         """
         parameter_values = [${measure_api_parameters}]
-        request = self._create_measure_request(parameter_values)
-
-        for response in self._get_stub().Measure(request):
-            yield self._deserialize_response(response)
+        with self._initialization_lock:
+            if self._measure_response is not None:
+                raise RuntimeError("A measure call is already in progress.")
```
jayaseelan-james, line 200: Elaborating on the error message could help users understand how to avoid this error.

```suggestion
                raise RuntimeError("A measurement is currently in progress. Concurrent measurement requests are not allowed.")
```
or

```suggestion
                raise RuntimeError("A measurement is currently in progress. To make concurrent measurement requests, please create a new client instance.")
```

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -183,7 +195,32 @@ class ${class_name}:
...
+        try:
+            for response in self._measure_response:
+                % if output_metadata:
+                yield self._deserialize_response(response)
+                % else:
+                yield
+                % endif
+        except grpc.RpcError as e:
+            if e.code() == grpc.StatusCode.CANCELLED:
+                _logger.debug("The measure call is canceled.")
```
jayaseelan-james, line 213:
```suggestion
                _logger.debug("The measurement is canceled.")
```

packages/generator/tests/acceptance/test_streaming_measurement_client.py
```diff
@@ -49,6 +50,61 @@ def test___measurement_plugin_client___stream_measure___returns_output(
...
+        with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
+            future_measure_1 = executor.submit(measurement_plugin_client.measure)
+            future_measure_2 = executor.submit(measurement_plugin_client.measure)
+            future_measure_1.result()
+            future_measure_2.result()
+
+    assert "A measure call is already in progress." in exc_info.value.args[0]
+
+
+def test___non_streaming_measurement_execution___cancel___returns_true(
```
jayaseelan-james, line 69: The "then" phrase of the test name should emphasize the outcome of the action. Here, "returns true" states the value and doesn't indicate the outcome.
```suggestion
def test___non_streaming_measurement_execution___cancel___cancels_measurement(
```

packages/generator/tests/acceptance/test_streaming_measurement_client.py
```diff
@@ -49,6 +50,61 @@ def test___measurement_plugin_client___stream_measure___returns_output(
...
+
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        measure = executor.submit(measurement_plugin_client.measure)
+        is_canceled = measurement_plugin_client.cancel()
+        measure.result()
+
+    assert is_canceled
+
+
+def test___streaming_measurement_execution___cancel___returns_true(
```
jayaseelan-james, line 83: Similarly, the "then" phrase of the test name should emphasize the outcome of the action. Here, "returns true" states the value and doesn't indicate the outcome.
```suggestion
def test___streaming_measurement_execution___cancel___cancels_measurement(
```

packages/generator/tests/acceptance/test_streaming_measurement_client.py
```diff
@@ -49,6 +50,61 @@ def test___measurement_plugin_client___stream_measure___returns_output(
...
+
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        measure = executor.submit(lambda: list(measurement_plugin_client.stream_measure()))
+        is_canceled = measurement_plugin_client.cancel()
+        measure.result()
+
+    assert is_canceled
+
+
+def test___measurement_client___cancel_without__measure___returns_false(
```
jayaseelan-james, line 97: Typo: Additional understand before measure.
```suggestion
def test___measurement_client___cancel_without_measure___returns_false(
```

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
Jotheeswaran-Nandagopal: updated to `raise RuntimeError("A measurement is currently in progress. To make concurrent measurement requests, please create a new client instance.")`

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: Sure, I have removed the error suppression and updated to raise gRPC CANCELLED exception. 

In LabVIEW, the cancel is not implemented yet. It is still in prototype stage, where it is planned to raise error or show cancelation message alone.

bkeryan:
@jayaseelan-james Since you requested changes, you need to approve this before it can be merged.

Jotheeswaran-Nandagopal:
Updated the test assets, after rebasing with main.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -183,7 +192,33 @@ class ${class_name}:
...
+        except grpc.RpcError as e:
+            if e.code() == grpc.StatusCode.CANCELLED:
+                _logger.debug("The measurement is canceled.")
+            raise
+        finally:
+            with self._initialization_lock:
+                self._measure_response = None
+
+    def cancel(self) -> bool:
+        """Cancels the active measure call."""
```
jayaseelan-james, line 219:
```suggestion
        """Cancels the active measurement call."""
```

or 

```suggestion
        """Cancels the current ongoing measurement call."""
```

packages/generator/tests/test_assets/example_renders/measurement_plugin_client/non_streaming_data_measurement_client.py
jayaseelan-james, line 1: Please consider having a work item, to extend the automated tests to generate a measurement client for a measurement with no outputs.

jayaseelan-james (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
Jotheeswaran-Nandagopal: Sure, I have updated to `"""Cancels the active measurement call."""`

packages/generator/tests/test_assets/example_renders/measurement_plugin_client/non_streaming_data_measurement_client.py
Jotheeswaran-Nandagopal: Sure. I am adding a work item to track adding a automated test.

PR #841, opened 2024-08-14
Create Measurement Plug-In Client generator using One line approach

Jotheeswaran-Nandagopal:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/measurement-services-python/blob/main/CONTRIBUTING.md).

### What does this Pull Request accomplish?

- Re-structured the Measurement Plugin SDK generator folder .
- Added `ni-measurement-plugin-client-generator` command.
- Implemented [one-line approach](https://dev.azure.com/ni/DevCentral/_git/ASW?path=/Specs/Test%20Automation%20Frameworks/E2719769_Support%20for%20Measurement%20Plug-In%20client%20APIs/E2719769_Support%20for%20Measurement%20Plug-In%20client%20APIs.md&version=GBmain&_a=preview&anchor=one-line-command-approach)  (creation of a Measurement Plug-In client-module with a single-line command) for Measurement Plug-in Client generation. [Enum support is not included in this implementation, whereas other datatypes - Integer, Float, Double, String, Arrays, Path, IO Resource, Pin and Doublexydatatypes are supported ]
- Implemented Measure API in Measurement Plug-in Client
- Added acceptance testing for the Measurement Plug-in Client generation

### Why should this Pull Request be merged?

- Implemented Measurement Plug-in Client generation and Measure API for the users to interact with measurement in CLI

### What testing has been done?

- Manually created Measurement Plug-in Client and tested
- Executed existing and new automated tests

PR #871, opened 2024-09-02
Added Register Pin Map API to Measurement Client

MounikaBattu17:

Adds `register_pin_map` API to measurement client.
Implements [Task 2827581](https://dev.azure.com/ni/DevCentral/_workitems/edit/2827581): Add register_pin_map API.

- Added `register_pin_map` and its usage in mako template.
- Added automated tests with new pin aware measurement.
- Added test assets accordingly.

Added automated tests


packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -187,3 +194,18 @@ class ${class_name}:
...
+        """Registers the pin map with the pin map service.
+        
+        Args:
+            pin_map_path: Absolute path of the pin map file.
+
+        Returns:
+            Registered pin map id.
+        """
+        pin_map_client = PinMapClient()
+        global _pin_map_path
```
Avinash2Suresh, line 208: Instead of keeping it as a global variable, could you please add it as a parameter for the class?

MounikaBattu17: Moved `pin_map_path` and `sites` as class members.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -187,3 +196,17 @@ class ${class_name}:
...
+        """Registers the pin map with the pin map service.
+        
+        Args:
+            pin_map_path: Absolute path of the pin map file.
+
+        Returns:
+            Registered pin map id.
+        """
+        pin_map_client = PinMapClient()
+        self._pin_map_path = pin_map_path
```
jayaseelan-james, line 210: Currently, the pin_map_id is the same as our pin map path, hence it may seem okay to store and use either of them but this could change in the future and will result in unexpected behavior. Therefore, we should use the return value from the update_pin_map method which will be the pin map ID.

packages/generator/tests/acceptance/test_pin_aware_measurement_client.py
```diff
@@ -0,0 +1,96 @@
...
+    measurement_plugin_client_module: ModuleType,
+) -> None:
+    test_measurement_client_type = getattr(measurement_plugin_client_module, "TestMeasurement")
+    measurement_plugin_client = test_measurement_client_type()
+
+    with pytest.raises(grpc.RpcError) as exc_info:
+        _ = measurement_plugin_client.measure()
+
+    assert exc_info.value.code() == grpc.StatusCode.UNKNOWN
+    assert "No sessions reserved." in (exc_info.value.details() or "")
```
jayaseelan-james, line 47:
```suggestion
    assert "No sessions reserved." in exc_info.value.details()
```

jayaseelan-james (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
MounikaBattu17: Assigned the registered `pin_map_id` to `_pin_map_path`.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -72,6 +76,8 @@ class ${class_name}:
         if grpc_channel is not None:
             self._stub = v2_measurement_service_pb2_grpc.MeasurementServiceStub(grpc_channel)
         self._create_file_descriptor()
+        self._pin_map_path = ""
+        self._sites = [0]
```
Avinash2Suresh, line 80: Can we expose the sites parameter as a public attribute and set its default value to 0? This would give users the flexibility to specify the sites from the client side.

packages/generator/tests/acceptance/test_pin_aware_measurement_client.py
MounikaBattu17: This change causes this mypy error. So, I don't think you should apply this suggestion.
[https://github.com/ni/measurement-plugin-python/actions/runs/10684062844/job/29613584248?pr=871#step:9:14](url)

jayaseelan-james: Okay. This is purely to satisfy the mypy compiler. Also, please consider asserting the complete error message here.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -187,3 +196,17 @@ class ${class_name}:
...
+    def register_pin_map(self, pin_map_path: str) -> str:
+        """Registers the pin map with the pin map service.
+        
+        Args:
+            pin_map_path: Absolute path of the pin map file.
+
+        Returns:
+            Registered pin map id.
+        """
+        pin_map_client = PinMapClient()
```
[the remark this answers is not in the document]
MounikaBattu17: Added a `cached_property` for pin map client.

dixonjoel (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -72,7 +77,15 @@ class ${class_name}:
         if grpc_channel is not None:
             self._stub = v2_measurement_service_pb2_grpc.MeasurementServiceStub(grpc_channel)
         self._create_file_descriptor()
-        
+        self.pin_map_id: str = ""
+        self.sites: list[int] = [0]
+
+    @cached_property
```
mshafer-NI, line 83: I'm guessing the reason to use a cache here is that the PinMapClient construction is expensive (timely) or that we want to only ever construct one per class.; however, that renders this property as immutable...

Can the _discovery_client or the _grcp_channel_pool ever change?

Is there a reason to prefer a lazy-cached property over creating the PinMapClient during `__init__` and storing the result in an object field?

packages/generator/tests/acceptance/test_pin_aware_measurement_client.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+@pytest.fixture(scope="module")
```
mshafer-NI, line 50: altering the fixture scope (from default) is a code-smell.

If you need to do this, please add a comment (here and on the others) *why*
Are these expensive to setup? (while still be safe to re-use?)

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -72,7 +77,15 @@ class ${class_name}:
         if grpc_channel is not None:
             self._stub = v2_measurement_service_pb2_grpc.MeasurementServiceStub(grpc_channel)
         self._create_file_descriptor()
-        
+        self.pin_map_id: str = ""
+        self.sites: list[int] = [0]
```
jayaseelan-james, line 81: Use `List` for type-hinting.
```suggestion
        self.sites: List[int] = [0]
```

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -187,3 +201,16 @@ class ${class_name}:
...
+        
+        Args:
+            pin_map_path: Absolute path of the pin map file.
+
+        Returns:
+            Registered pin map id.
+        """
+        self.pin_map_id = self.pin_map_client.update_pin_map(pin_map_path)
+
+        return self.pin_map_id
```
jayaseelan-james, line 216: Do we have any need to return the pin_map_id?

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
Jotheeswaran-Nandagopal, line 81: Here, sites is exposed a public property. Can we proceed with this? 
May I know your suggestions. @dixonjoel

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -187,3 +201,16 @@ class ${class_name}:
 
         for response in self._get_stub().Measure(request):
             yield self._deserialize_response(response)
+
+    def register_pin_map(self, pin_map_path: str) -> str:
```
Jotheeswaran-Nandagopal, line 205: `register_pin_map` method is used only to update the pin map alone. Can we proceed keeping this as method or make it as property to update the pin map?
May I know your suggestions. @dixonjoel @mshafer-NI

dixonjoel: Seems reasonable to use a method to me.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
dixonjoel: Is there any reason to expose it at this time? I would lean towards keeping it private.

Avinash2Suresh (changes requested):
Waiting for the existing comments to be addressed.

packages/generator/tests/acceptance/test_pin_aware_measurement_client.py
MounikaBattu17: The scope of this fixture is changed to `module` as this fixture creates a client for the measurement service. Since this setup is expensive, we made this fixture to be reusable throughout the module. This method of caching the fixtures is used in many tests. So, please let us know if you want to add an explicit comment only on this.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
MounikaBattu17: Replaced the cached property by creating a client during `__init__` and storing the result in a private field.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
MounikaBattu17: Introduced a setter for sites, which strictly expects the datatype to be `List[int]`. So, removing the explicit datatype mention here.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
MounikaBattu17: No, not necessary. Hence removed.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
MounikaBattu17: Made the `sites` as a read-only property by returning it as a `tuple` and added a setter to make it changeable by the users.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -58,41 +63,77 @@ class ${class_name}:
...
             discovery_client: An optional discovery client.
 
+            pin_map_client: An optional pin map client.
+
             grpc_channel: An optional gRPC channel targeting a measurement service.
 
             grpc_channel_pool: An optional gRPC channel pool.
         """
-        self._initialization_lock = threading.Lock()
+        self._initialization_lock = threading.RLock()
```
Avinash2Suresh, line 72: why are we using Reentrant lock here?

MounikaBattu17: To create other clients, we use separate locks, which can lead to deadlocks when calling methods from here like `_get_discovery_client`. Using RLock in this context allows the same thread to release a previously acquired lock if necessary.

MounikaBattu17:
@mshafer-NI @jayaseelan-james I've fixed all your comments, please share your review.

packages/generator/tests/acceptance/test_pin_aware_measurement_client.py
```diff
@@ -0,0 +1,96 @@
...
+import grpc
+import pytest
+from ni_measurement_plugin_sdk_service.measurement.service import MeasurementService
+
+from ni_measurement_plugin_sdk_generator.client import create_client
+from tests.utilities.discovery_service_process import DiscoveryServiceProcess
+from tests.utilities.measurements import pin_aware_measurement
+
+
+def test___measurement_plugin_client___measure_with_pin_map_registration___returns_output(
```
jayaseelan-james, line 15: Please create a unit test for multi-site measurement execution.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
jayaseelan-james: > ... return a tuple of the sites and provide a "set" method to change it.

@mshafer-NI, any specific reasons to prefer a set method but not a property setter (@sites.setter) here?

jayaseelan-james (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -58,41 +63,76 @@ class ${class_name}:
...
+        self._pin_map_id = ""
+        self._sites = [0]
+
+    @property
+    def sites(self) -> Tuple[int]:
+        """Specifies the list of sites being monitored."""
+        return tuple(self._sites)
+    
+    def set_sites(self, val):
+        if not isinstance(val, list):
```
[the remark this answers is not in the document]
MounikaBattu17: `Iterable` allows room for more datatypes. Thus going with `Iterable` than `List`.

packages/generator/tests/acceptance/test_pin_aware_measurement_client.py
mshafer-NI: please add this as a comment 
```suggestion
# Fixture is setup to be re-used across tests due to expense to setup.
@pytest.fixture(scope="module")
```
(or something like that)

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -58,41 +63,76 @@ class ${class_name}:
...
-        
+        self._pin_map_id = ""
+        self._sites = [0]
+
+    @property
+    def sites(self) -> Tuple[int, ...]:
+        """Sites for which the measurement is being executed."""
+        return tuple(self._sites)
+    
+    def set_sites(self, val):
```
Avinash2Suresh, line 91: Please add type-hinting for the parameter & rename the parameter to sites instead of val

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -58,41 +63,76 @@ class ${class_name}:
...
+        if not isinstance(val, Iterable):
+            raise ValueError("Sites must be a iterable")
```
Avinash2Suresh, line 93: an iterable.

Avinash2Suresh (approved):

packages/generator/tests/acceptance/test_pin_aware_measurement_client.py
bkeryan: @mshafer-NI IMO, scope=module is fine for fixtures in a test module, and I would rather not write comments stating the obvious.

What has been a problem lately is moving fixtures from a test module to conftest.py without changing the scope. I see no reason for conftest.py to ever use scope=module. If it's expensive enough to cache, then it's expensive enough to cache across modules.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -58,41 +63,76 @@ class ${class_name}:
...
+        self._pin_map_client = pin_map_client
         self._stub: Optional[v2_measurement_service_pb2_grpc.MeasurementServiceStub] = None
         self._configuration_metadata = ${configuration_metadata}
         self._output_metadata = ${output_metadata}
         if grpc_channel is not None:
             self._stub = v2_measurement_service_pb2_grpc.MeasurementServiceStub(grpc_channel)
         self._create_file_descriptor()
-        
+        self._pin_map_id = ""
+        self._sites = [0]
```
bkeryan, line 84: Consider replacing these with `self._pin_map_context: Optional[PinMapContext] = None`

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
mshafer-NI: A property setter is implemented as a method -> and that just boils down to api -> does it make sense to be able to do "foo.sites = bar" or not.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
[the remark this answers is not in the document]
bkeryan: Cc: @mshafer-NI since you suggested this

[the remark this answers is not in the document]
mshafer-NI: I replied here -> https://github.com/ni/measurement-plugin-python/pull/871#discussion_r1752308594

we have 3 threads on this going on...

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
mshafer-NI: (And I did not mean to imply a decision for either `foo.set_sites(...)` or `foo.sites = ` via property setter -> this style choice should be driven by what "makes sense" in the context of what is getting set, and the rest of the API uses.)

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -187,3 +228,11 @@ class ${class_name}:
 
         for response in self._get_stub().Measure(request):
             yield self._deserialize_response(response)
+
+    def register_pin_map(self, pin_map_path: str) -> None:
```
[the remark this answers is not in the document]
bkeryan: BTW, an optional `pin_map_context` parameter would allow users to use a single client object to make multiple concurrent calls to the same service with different sites. However, we have already agreed to put cancellation state in the client object, so it's not necessary to make the client object entirely stateless.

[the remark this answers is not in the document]
MounikaBattu17: We only support registering pin map to the pin map service and not unregistering. Users won't be able to unregister an already registered pin map using this API because we don't have specific grpc calls for unregistering pin map in [pin_map_service.proto](https://github.com/ni/ni-apis/blob/45ef02ecc751cf3aecf7e6130d9ac8efc54f5e0a/ni/measurementlink/pinmap/v1/pin_map_service.proto).

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
bkeryan: The other client objects use lazy initialization in order to defer the expensive operations (DiscoveryService.resolve_service, creating gRPC channels, and DNS lookups) until needed. I previously requested that this client do the same: https://github.com/ni/measurement-plugin-python/pull/841#discussion_r1718982257

_discovery_client and _grpc_channel_pool do not change.

I prefer lazy initialization, but not using `@cached_property`, because its locking behavior was too coarse in older Python versions.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
MounikaBattu17: Not applicable as we removed sites property.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
MounikaBattu17: Not applicable as we removed sites property.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -139,7 +179,8 @@ class ${class_name}:
             )
         )
         return v2_measurement_service_pb2.MeasureRequest(
-            configuration_parameters=serialized_configuration
+            configuration_parameters=serialized_configuration,
+            pin_map_context=PinMapContext(pin_map_id=self._pin_map_id, sites=self._sites),
```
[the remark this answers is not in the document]
MounikaBattu17: Sure! I've added if else to determine whether to include `pin_map_context` parameter in the measure request or not based on `self.pin_map_context`.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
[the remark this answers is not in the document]
MounikaBattu17: Changed to `pathlib.Path`.

packages/generator/tests/test_assets/example_renders/pin_map/1Smu1ChannelGroup1Pin1Site.pinmap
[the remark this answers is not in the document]
MounikaBattu17: The code generator doesn't generate pin maps. That's why I explicitly added the required pin maps. Moved pin map folder to test_assets.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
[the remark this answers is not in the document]
MounikaBattu17: > users to use a single client object to make multiple concurrent calls to the same service with different sites

As we discussed in [HLD](https://dev.azure.com/ni/DevCentral/_git/ASW/pullRequest/751696#1721325101), the current implementation does not support multiple concurrent calls from a single client object. Instead, calls must be made sequentially, one after the other. 

So, I've considered option 1 of adding a property for `pin_map_context` with getter and setter. This way users can update the `pin_map_context` for each measure call. For example, like below

```
    client = NIDCPowerSourceDCVoltageClient()
    client.pin_map_context = PinMapContext(pin_map_id="NIDCPowerSourceDCVoltage.pinmap", sites=[0])
    client.measure()
    client.pin_map_context = PinMapContext(pin_map_id="NIDCPowerSourceDCVoltageMultiSite.pinmap", sites=[1])
    client.measure()
```

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
MounikaBattu17: Replaced with `self._pin_map_context`.

Also, since sites is a repeated field, we couldn't directly assign value to sites like this `client.pin_map_context.sites = [1,2,3]`. Instead if users wants to update sites, they should be creating a new `PinMapContext` and set `sites` through dependency injection. 
For example,
```
    client = NIDCPowerSourceDCVoltageClient()
    client.pin_map_context = PinMapContext(pin_map_id="NIDCPowerSourceDCVoltageMultiSite.pinmap", sites=[0, 1])
    client.measure()
```

I've also added `sites` as an optional parameter to `register_pin_map` such that users can specify sites while registering pin map.

```
    client = NIDCPowerSourceDCVoltageClient()
    client.register_pin_map(pin_map_path, sites=[0, 1])
    client.measure()
```

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -58,41 +63,76 @@ class ${class_name}:
...
         self._output_metadata = ${output_metadata}
         if grpc_channel is not None:
             self._stub = v2_measurement_service_pb2_grpc.MeasurementServiceStub(grpc_channel)
         self._create_file_descriptor()
-        
+        self._pin_map_id = ""
+        self._sites = [0]
+
+    @property
+    def sites(self) -> Tuple[int, ...]:
```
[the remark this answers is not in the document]
bkeryan: I see.

This also highlights a problem with the `PinMapContext` class:
https://github.com/ni/measurement-plugin-python/blob/main/packages/service/ni_measurement_plugin_sdk_service/session_management/_types.py#L22

It's a named tuple, so you can't assign to `sites`. However, `sites` is an `Optional[List]`, so if it's not `None`, you can mutate it.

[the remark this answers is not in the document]
bkeryan: Anyway, it seems weird to use a `Tuple[int, ...]` here for a field that is a `List[int]` in other parts of this API.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
bkeryan: Don't use the gRPC type in public interfaces. Use the named tuple: https://github.com/ni/measurement-plugin-python/blob/main/packages/service/ni_measurement_plugin_sdk_service/session_management/_types.py#L22

Then call `_to_grpc` to convert to the gRPC type.

> Also, since sites is a repeated field, we couldn't directly assign value to sites...

The named tuple prevents assignment too, but it allows replacing fields.

```
@sites.setter
def sites(self, val):
    self._pin_map_context = self._pin_map_context._replace(sites=val)
```

> Instead if users wants to update sites, they should be creating a new PinMapContext and set sites through dependency injection.

I don't see how this is dependency injection.

> I've also added sites as an optional parameter to register_pin_map such that users can specify sites while registering pin map if needed.

I need to think about this some more.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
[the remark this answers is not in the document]
mshafer-NI: API design, yay!

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
MounikaBattu17: Thanks for the suggestion! I've replaced the gRPC type `PinMapContext` with named tuple and also, added `sites` property to set the sites externally.

MounikaBattu17: > I've also added sites as an optional parameter to register_pin_map such that users can specify sites while registering pin map if needed.

I proposed this previously because we didn't have a direct means to set sites. To set sites, we need to create new `PinMapContext` object. But now with the named tuple approach, user can set sites directly, So I've removed this sites parameter from r`egister_pin_map` which will now only update `pin_map_id` in `self._pin_map_context`.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -138,9 +186,15 @@ class ${class_name}:
...
-        )
+        if self._pin_map_context is None:
+            return v2_measurement_service_pb2.MeasureRequest(
+                configuration_parameters=serialized_configuration,
+            )
+        else:
+            return v2_measurement_service_pb2.MeasureRequest(
+                configuration_parameters=serialized_configuration,
+                pin_map_context=self._pin_map_context._to_grpc(),
+            )
```
bkeryan, line 197: Protobuf allows setting pin_map_context=None, so you don't have to duplicate the entire constructor call.
```suggestion
        return v2_measurement_service_pb2.MeasureRequest(
            configuration_parameters=serialized_configuration,
            pin_map_context=self._pin_map_context._to_grpc() if self._pin_map_context else None,
        )
```

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -187,3 +241,15 @@ class ${class_name}:
...
+        """Registers the pin map with the pin map service.
+
+        Args:
+            pin_map_path: Absolute path of the pin map file.
+        """
+        pin_map_id = self._get_pin_map_client().update_pin_map(pin_map_path)
+        if self._pin_map_context is None:
+            self._pin_map_context = PinMapContext(pin_map_id=pin_map_id, sites=[0])
+        else:
+            self._pin_map_context._replace(pin_map_id=pin_map_id)
```
bkeryan, line 255: This does not work because named tuples are immutable and _replace has no side effects.

Please add test coverage for this case before fixing this.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -187,3 +241,15 @@ class ${class_name}:
...
+
+    def register_pin_map(self, pin_map_path: pathlib.Path) -> None:
+        """Registers the pin map with the pin map service.
+
+        Args:
+            pin_map_path: Absolute path of the pin map file.
+        """
+        pin_map_id = self._get_pin_map_client().update_pin_map(pin_map_path)
+        if self._pin_map_context is None:
+            self._pin_map_context = PinMapContext(pin_map_id=pin_map_id, sites=[0])
```
bkeryan, line 253: Will the corresponding LabVIEW API also default to sites=[0]?

packages/generator/tests/test_assets/example_renders/measurement_plugin_client/non_streaming_data_measurement_client.py
```diff
@@ -298,27 +305,70 @@ def __init__(
...
+
+    @property
+    def pin_map_context(self) -> PinMapContext:
+        """Get the pin map context for the RPC."""
+        return self._pin_map_context
+
+    @pin_map_context.setter
+    def pin_map_context(self, val: PinMapContext) -> None:
+        if not isinstance(val, PinMapContext):
+            raise ValueError("pin_map_context must be an instance of PinMapContext.")
```
bkeryan, line 318: Use TypeError.

Include the invalid type in the error message.

See https://codesearch.natinst.com/search/text?q=raise%20TypeError%20repo%3Agithub-meas&fold_case=auto&regex=true&context=true for examples.

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
bkeryan: I think the combination of `client.register_pin_map(path)` and `client.sites = [1,2,3]` is fine.

I am not concerned with the use of a mutable list. If users do `client.pin_map_context.sites.append(4)`, it will do the right thing, because the PInMapContext named tuple is the single source of truth for the selected sites.

packages/generator/tests/test_assets/example_renders/measurement_plugin_client/non_streaming_data_measurement_client.py
MounikaBattu17: Updated the error message.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
MounikaBattu17: Thanks for pointing out this issue! To fix this issue, I have assigned the replaced value back to `self._pin_map_context`. I have also added test cases to check this behavior.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
MounikaBattu17: Currently, LabVIEW client also sets the sites to` [0]` by default.
image

In LabVIEW, users cannot set external sites like registered pin map id, but it will be supported soon.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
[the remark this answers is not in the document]
bkeryan: ok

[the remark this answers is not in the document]
bkeryan: ok, and the cancellation PR enforces this

packages/generator/tests/test_assets/example_renders/measurement_plugin_client/non_streaming_data_measurement_client.py
bkeryan: Please include the invalid type in the error message.

bkeryan (approved):
Approved with suggestions (include invalid type in error message)

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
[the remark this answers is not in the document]
MounikaBattu17: Resolving this thread. Please refer this [thread](https://github.com/ni/measurement-plugin-python/pull/871#discussion_r1752287928) for details.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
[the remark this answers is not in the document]
MounikaBattu17: Resolving this thread. Please refer this [thread](https://github.com/ni/measurement-plugin-python/pull/871#discussion_r1752287928) for details.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
MounikaBattu17: Resolving this thread. Please refer this [thread](https://github.com/ni/measurement-plugin-python/pull/871#discussion_r1752287928) for details.

packages/generator/tests/acceptance/test_pin_aware_measurement_client.py
```diff
@@ -0,0 +1,197 @@
...
+    measurement_plugin_client = test_measurement_client_type()
+
+    with pytest.raises(grpc.RpcError) as exc_info:
+        _ = measurement_plugin_client.measure()
+
+    assert exc_info.value.code() == grpc.StatusCode.UNKNOWN
+    assert "No sessions reserved." in (exc_info.value.details() or "")
+
+
+def test___measurement_plugin_client___register_pin_map_without_pin_map_context___updates_pin_map_id_and_sites(
```
jayaseelan-james, line 51: Should the "then" phrase say `...___creates_pin_map_context_with_pin_map_id_and_sites` because we're asserting the pin map context object?

Or 

Should you only assert the pin map id and sites values separately?

packages/generator/tests/acceptance/test_pin_aware_measurement_client.py
```diff
@@ -0,0 +1,197 @@
...
+    test_measurement_client_type = getattr(measurement_plugin_client_module, "TestMeasurement")
+    measurement_plugin_client = test_measurement_client_type()
+
+    measurement_plugin_client.register_pin_map(pin_map_path)
+
+    expected_pin_map_context = PinMapContext(pin_map_id=str(pin_map_path), sites=[0])
+    assert measurement_plugin_client.pin_map_context == expected_pin_map_context
+
+
+def test___measurement_plugin_client___register_pin_map_with_pin_map_context___updates_pin_map_id(
```
jayaseelan-james, line 65: Similarly, here `___updates_pin_map_context_with_pin_map_id`?

packages/generator/tests/acceptance/test_pin_aware_measurement_client.py
```diff
@@ -0,0 +1,197 @@
...
+    expected_output = output_type(
+        pin_map_id=str(pin_map_path),
+        sites=[0, 1],
+        session_names=["DCPower1/0, DCPower1/1, DCPower1/2, DCPower1/3"],
+        resource_names=["DCPower1/0, DCPower1/1, DCPower1/2, DCPower1/3"],
+        channel_lists=["DCPower1/0, DCPower1/2"],
+    )
+    test_measurement_client_type = getattr(measurement_plugin_client_module, "TestMeasurement")
+    measurement_plugin_client = test_measurement_client_type()
+    measurement_plugin_client.register_pin_map(pin_map_path)
```
jayaseelan-james, line 141: Is this register_pin_map call required?

MounikaBattu17: Thanks for pointing this out. register pin map call is not needed here, it negates the point of this test.

packages/generator/tests/acceptance/test_pin_aware_measurement_client.py
MounikaBattu17: Checking the `pin_map_context` is the m=aim of this test so updated the test case name.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -58,41 +62,88 @@ class ${class_name}:
...
+    @property
+    def sites(self) -> List[int]:
+        """Sites for which the measurement is being executed."""
+        if self._pin_map_context is not None:
+            return self._pin_map_context.sites
+
+    @sites.setter
+    def sites(self, val: List[int]) -> None:
+        if self._pin_map_context is None:
+            raise AttributeError("Cannot set sites because pin map context is None.")
```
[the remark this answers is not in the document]
MounikaBattu17: Thanks! This message will helps to understand the error better.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -58,41 +62,88 @@ class ${class_name}:
...
         self._output_metadata = ${output_metadata}
         if grpc_channel is not None:
             self._stub = v2_measurement_service_pb2_grpc.MeasurementServiceStub(grpc_channel)
         self._create_file_descriptor()
-        
+        self._pin_map_context: Optional[PinMapContext] = None
+
+    @property
+    def pin_map_context(self) -> PinMapContext:
+        """Get the pin map context for the RPC."""
```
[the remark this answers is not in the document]
MounikaBattu17: I referred this docstring from other parts of repo. But "RPC" here doesn't make much sense so changed to 
`"""Get the pin map context for the measurement."""`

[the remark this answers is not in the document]
jayaseelan-james: Should we say, "Get or set the pin map ..." because the same doc is displayed when getting and setting the pin_map_context?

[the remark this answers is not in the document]
MounikaBattu17: Makes sense! Noted, will address in upcoming PRs.

PR #872, opened 2024-09-03
Support for Path and String datatypes with escape sequences

Jotheeswaran-Nandagopal:

- Added Path handling in Measurement Plug-In Client, which allows users to provide `pathlib.Path` input. 
- Added the escape sequence handlings for `Path` and `string` inputs.

- Let users to give `pathlib.Path` as input and convert it to string in the Client before passing it to the measurement. (This is a work-around as Measurement Plug-In does not accepts `pathlib.Path` input)
- Added conversions to handle the escape sequences (like `\n`, `\t`) in `string` and `Path` inputs.

- Tested manually.
- Executed existing automated tests.


packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -157,6 +157,33 @@ class ${class_name}:
             result[k - 1] = v
         return Output._make(result)
 
+    def _validate_parameters(self, parameter_values: List[Any]) -> List[Any]:
```
Jotheeswaran-Nandagopal, line 160: I have converted the Path to string here. Passing `Path` to serialization throws error, as serialization in Measurement Plug-In Service excepts string type.

dixonjoel (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
mshafer-NI, line 160: This isn't doing validation -> it's doing type shifting....

Would `_ensure_parameter_types` or `_convert_paths_to_strings` be a better name?

Jotheeswaran-Nandagopal: Sure. I have renamed to `_convert_paths_to_strings`.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -177,17 +177,13 @@ def get_configuration_parameters_with_type_and_default_values(
         default_value = metadata.default_value
         parameter_type = _get_python_type_as_str(metadata.type, metadata.repeated)
         if isinstance(default_value, str):
-            default_value = f'"{default_value}"'
+            default_value = default_value.encode("unicode-escape")
```
bkeryan, line 180: This generates `bytes` literals, not `str` literals. Also, the way it does this is somewhat convoluted:
- `default_value.encode("unicode-escape")` evaluates to a `bytes` containing the escaped string, without any surrounding quotes.
- `f"... = {default_value}"` prints a `bytes` literal or b-string, which escapes everything again. I'm not sure if this generates the right number of quotes and backslashes because none of the test cases use these characters.

Can you use `default_value = repr(default_value)` instead?

packages/generator/tests/test_assets/example_renders/measurement_plugin_client/non_streaming_data_measurement_client.py
```diff
@@ -420,11 +442,11 @@ def stream_measure(
...
         double_array_in: List[float] = [0.1, 0.2, 0.3],
         bool_in: bool = False,
-        string_in: str = "sample string",
+        string_in: str = b"sample string",
         string_array_in: List[str] = ["String1", "String2"],
-        path_in: Path = r"path/test",
+        path_in: Path = b"path/test",
         path_array_in: List[Path] = ["path/test1", "path/ntest2"],
-        io_in: str = "resource",
+        io_in: str = b"resource",
```
bkeryan, line 449: These type hints are false. `b"sample string"` is a `bytes`, not a `str`; likewise, `b"path/test"` is a `bytes`, not a `Path`. Mypy accepts your type hints at face value regardless of whether they are truthful or not.

packages/generator/tests/test_assets/example_renders/measurement_plugin_client/non_streaming_data_measurement_client.py
```diff
@@ -420,11 +442,11 @@ def stream_measure(
         float_in: float = 0.05999999865889549,
         double_array_in: List[float] = [0.1, 0.2, 0.3],
         bool_in: bool = False,
-        string_in: str = "sample string",
+        string_in: str = b"sample string",
```
bkeryan, line 445: None of these test cases use escapable characters. Please add test cases that use backslash, single quote, and double quote.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -157,6 +157,33 @@ class ${class_name}:
...
+            # Convert each Path in the list to string and convert '\' to '\\' in strings.
+            if isinstance(parameter_value, list):
+                converted_list = []
+                for value in parameter_value:
+                    % if "from pathlib import Path" in built_in_import_modules:
+                    if isinstance(value, (Path, str)):
+                    % else:
+                    if isinstance(value, str):
+                    % endif
+                        value = str(value).encode('unicode-escape')
```
bkeryan, line 173: Why are you escaping the string at run time? You only need to escape strings when generating Python code. gRPC can handle any string you pass to it, and the measurement service on the other side will not apply Python-specific unescaping rules.

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Jotheeswaran-Nandagopal: Sure. I have updated with `repr`.

packages/generator/tests/test_assets/example_renders/measurement_plugin_client/non_streaming_data_measurement_client.py
Jotheeswaran-Nandagopal: Sure, now I have updated this with `repr`.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
bkeryan: > I am escaping the string at run-time to send the output in a presentable format to the users.

That is a presentation issue. The network client class should not do this. 

The point is to send the string to the measurement service, and most services will not unescape values received over gRPC. Why would they? gRPC can send tabs, newlines, and backslashes without needing escape sequences.

> Here, lets take a example string of - "sample\test\newstring\file".

How are you using this string? 

Are you using this string as a Python string literal? If so, Python is interpreting the escape sequences and you are sending <tab>est<newline>ewstring, etc. over the network.

Are you entering this string into a string control in the Measurement Plug-In UI Editor? If so, the behavior depends on whether **Show escape sequences** is checked.
- If **Show escape sequences** is unchecked (default), then Measurement Plug-In UI Editor sends backslash-t, backslash-n, etc. over the network.
- If **Show escape sequences** is checked, then Measurement Plug-In UI Editor converts backslash-t to tab, backslash-n to newline, etc. before sending it over the network. However, if you already entered the string before you checked the option, then it doubles all of the backslashes in order to preserve them.

> Assuming this as loopback measurement, when user tries to print the output string, the result will be
> 
> > sample est
> > ewstring\xocile

If you are actually sending <tab>, <newline>, etc. over the network and you have **Show escape sequences** unchecked, then this behavior is correct.

For testing with a loopback measurement, add something like
```
print(string_value)
print(repr(string_value))
```
so you can see the value that was received by the service, regardless of how Measurement Plug-In UI Editor or InstrumentStudio displays it.

Also, please test with a service that actually uses the string value for something. For example, `nidigital_spi`.
- Install NI-Digital Pattern Driver
- Copy examples/.env.simulation to examples/.env
- cd examples/nidigital_spi
- poetry install
- Add `print(specifications_file)`
- poetry run python measurement.py -v
- Open the project in InstrumentStudio
- Try different values for the specifications file

For example, if I use the browse button to navigate to the specifications file at this path: `D:\dev\measurement-plugin-python\examples\nidigital_spi\Specifications.specs`
then InstrumentStudio displays this:
image
and it prints the path with single backslashes:
```
PS D:\dev\measurement-plugin-python\examples\nidigital_spi> poetry run python .\measurement.py -v
2024-09-11 17:15:45,821 INFO: Measurement service listening on: http://[::1]:53668
2024-09-11 17:15:46,338 INFO: Successfully registered with discovery service.
Press enter to close the measurement service.
2024-09-11 17:15:53,849 INFO: gRPC server call /ni.measurementlink.measurement.v2.MeasurementService/GetMetadata responded OK in 1.1913 ms
2024-09-11 17:15:53,987 INFO: Starting test: pin_names=['SPI_PINS']
D:\dev\measurement-plugin-python\examples\nidigital_spi\Specifications.specs
2024-09-11 17:15:54,066 INFO: Completed test: passing_sites=[0] failing_sites=[]
2024-09-11 17:15:54,066 INFO: gRPC server call /ni.measurementlink.measurement.v2.MeasurementService/Measure responded OK in 79.4516 ms
2024-09-11 17:16:00,592 INFO: Successfully unregistered with discovery service.
2024-09-11 17:16:00,594 INFO: Measurement service closed.
```
It did not convert `\nidigital_spi` to `<newline>idigital_spi`, and it did not print double backslashes.

bkeryan: Also, I don't think the Show escape sequences checkbox is working correctly, so I filed AB#2855335

Jotheeswaran-Nandagopal: - Sure, thanks for pointing this. I get the point. Now I have removed the encoding of strings to escape unicode-escaping. 

Now, after this change,
If the user gives input as `string_input = sample\test\new`,

`print(string_input)` will give below as the output, 

> sample    est
> ew


`print(repr(string_input))` will give below as the output, 

> "sample\test\new"

bkeryan: ok

packages/generator/tests/test_assets/example_renders/measurement_plugin_client/non_streaming_data_measurement_client.py
bkeryan: These type hints are still wrong because the data types are wrong. `"path/test" is a `str` not a `Path`. To make the type hints correct, you must call the Path constructor:

```
path_in: Path = Path("path/test"),
path_array_in: List[Path] = [Path("path/test1"), Path("path/test2")],
```

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -156,33 +156,26 @@ class ${class_name}:
...
         return Output._make(result)
-
+    
+    % if "from pathlib import Path" in built_in_import_modules:
     def _convert_paths_to_strings(self, parameter_values: List[Any]) -> List[Any]:
         result = []
 
         for parameter_value in parameter_values:
-            # Convert each Path in the list to string and convert '\' to '\\' in strings.
+            # Converts Path type to string.
```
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: Removed the comment.

packages/generator/tests/test_assets/example_renders/measurement_plugin_client/non_streaming_data_measurement_client.py
Jotheeswaran-Nandagopal: I have updated values of path as follows
```
path_in: Path = WindowsPath("path/test"),
path_array_in: List[Path] = [WindowsPath("path/test1"), WindowsPath("path/test2")],
```
As path conversion in list send the data as `WindowsPath` to the generated file. To maintain consistency, I updated this for both Path and List[Path].

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -68,7 +68,7 @@ class ${class_name}:
         self._discovery_client = discovery_client
         self._stub: Optional[v2_measurement_service_pb2_grpc.MeasurementServiceStub] = None
         self._measure_response: Optional[
-            Generator[v2_measurement_service_pb2.MeasureResponse, None, None]
+            grpc.CallIterator[v2_measurement_service_pb2.MeasureResponse]
```
Jotheeswaran-Nandagopal, line 71: I have updated the type to proper type hinting.

Jotheeswaran-Nandagopal:
fixed pin aware measurement test case, as the tests are failing, after merge with main.

packages/generator/tests/utilities/measurements/non_streaming_data_measurement/__init__.py
```diff
@@ -21,13 +21,13 @@
...
-@measurement_service.configuration("String In", nims.DataType.String, "sample string")
+@measurement_service.configuration("String In", nims.DataType.String, 'sample "string"')
 @measurement_service.configuration(
     "String Array In", nims.DataType.StringArray1D, ["String1", "String2"]
 )
-@measurement_service.configuration("Path In", nims.DataType.Path, "path/test")
+@measurement_service.configuration("Path In", nims.DataType.Path, "path\test")
 @measurement_service.configuration(
-    "Path Array In", nims.DataType.PathArray1D, ["path/test1", "path/ntest2"]
+    "Path Array In", nims.DataType.PathArray1D, ["path\test1", "path\ntest2"]
```
bkeryan, line 30: This correctly tests escape sequences in paths, but testing only that case is extremely misleading. Users almost never want escape sequences in paths. The 99% use case for backslashes in paths is as a Windows path separator.

Please test forward slash, backslash (you need to escape it or use r-strings), single quote, double quote, and escape sequences. Do this for both strings and path arrays. It's ok for the single string/path to test a subset of these.

Example:
- `["path/with/forward/slash", "path\\with\\backslash", "path with 'single quotes'", 'path with "double quotes"', "path\twith\ttabs"]`

bkeryan (changes requested):

packages/generator/tests/utilities/measurements/non_streaming_data_measurement/__init__.py
Jotheeswaran-Nandagopal: Sure, I have updated the test cases for string array and path array.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -214,8 +214,28 @@ class ${class_name}:
         for k, v in output_values.items():
             result[k - 1] = v
         return Outputs._make(result)
-
+    
+    % endif
+    % if "from pathlib import Path" in built_in_import_modules:
+    def _convert_paths_to_strings(self, parameter_values: List[Any]) -> List[Any]:
```
jayaseelan-james, line 220: nit: This need not be a class instance method as it doesn't use any of the class properties or methods.

jayaseelan-james (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
Jotheeswaran-Nandagopal: Moved this method outside the class.

bkeryan (approved):

PR #876, opened 2024-09-06
[Minor] Add dependency to Measurement Plug-In Client

Jotheeswaran-Nandagopal:

- Add "black" to dependency in Measurement Plug-In Client.

- This implementation contains the addition of `black` as package dependency, as fresh machine installation requires `black`.

- Existing automation tests passed.

### DIFF
--- a/packages/generator/pyproject.toml
+++ b/packages/generator/pyproject.toml
@@ -24,6 +24,7 @@ Mako = "^1.2.1"
 click = ">=8.1.3"
 grpcio = "^1.49.1"
 protobuf = "^4.21"
+black = ">=24.8.0"
 # ni-measurement-plugin-sdk-service = {version = "^2.0.0"}
 
 [tool.poetry.group.dev.dependencies]
--- a/packages/generator/poetry.lock
+++ b/packages/generator/poetry.lock
@@ -907,4 +907,4 @@ files = [
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.8"
-content-hash = "f276841e13db46d813e139a356e2e3d6f1a71f4ec95de0c04e31b21125e2b4fa"
+content-hash = "d08cfb61e14d582ad229b454eb84b738e01edd0afdc3a6968ca11edeaa2630de"


jayaseelan-james (approved):
> Added missing dependency.

Please update the description explaining why black is required as a dependency.

packages/generator/pyproject.toml
```diff
@@ -24,6 +24,7 @@ Mako = "^1.2.1"
 click = ">=8.1.3"
 grpcio = "^1.49.1"
 protobuf = "^4.21"
+black = ">=24.8.0"
```
Avinash2Suresh, line 27: why it is added as `>=` instead of `^`

dixonjoel:
> > Added missing dependency.
> 
> Please update the description explaining why black is required as a dependency.

I don't know why this is needed. We already have 'black' in the poetry.lock. Do you need a specific version for some reason?
https://github.com/ni/measurement-plugin-python/blob/5974af6c97d207c7c2e738890fad074a4d2831e2/packages/generator/poetry.lock#L4

packages/generator/pyproject.toml
Jotheeswaran-Nandagopal: `>=` is used because any higher versions than 24.8.0 can be used but "^" only supports version of 24.x.x but not 25.

Jotheeswaran-Nandagopal: `>=` is used because any higher versions than 24.8.0 can be used but "^" only supports version of 24.x.x but not 25.

Jotheeswaran-Nandagopal:
> > > Added missing dependency.
> > 
> > 
> > Please update the description explaining why black is required as a dependency.
> 
> I don't know why this is needed. We already have 'black' in the poetry.lock. Do you need a specific version for some reason?
> 
> https://github.com/ni/measurement-plugin-python/blob/5974af6c97d207c7c2e738890fad074a4d2831e2/packages/generator/poetry.lock#L4

WIthout adding the black to the 

> > > Added missing dependency.
> > 
> > 
> > Please update the description explaining why black is required as a dependency.
> 
> I don't know why this is needed. We already have 'black' in the poetry.lock. Do you need a specific version for some reason?
> 
> https://github.com/ni/measurement-plugin-python/blob/5974af6c97d207c7c2e738890fad074a4d2831e2/packages/generator/poetry.lock#L4

Without adding the "black" in `[tool.poetry.dependencies]`, fresh machine testing throws error with `black package is missing` error. So, adding "black".

Jotheeswaran-Nandagopal:
> > > > Added missing dependency.
> > > 
> > > 
> > > Please update the description explaining why black is required as a dependency.
> > 
> > 
> > I don't know why this is needed. We already have 'black' in the poetry.lock. Do you need a specific version for some reason?
> > https://github.com/ni/measurement-plugin-python/blob/5974af6c97d207c7c2e738890fad074a4d2831e2/packages/generator/poetry.lock#L4
> 

> 
> > > > Added missing dependency.
> > > 
> > > 
> > > Please update the description explaining why black is required as a dependency.
> > 
> > 
> > I don't know why this is needed. We already have 'black' in the poetry.lock. Do you need a specific version for some reason?
> > https://github.com/ni/measurement-plugin-python/blob/5974af6c97d207c7c2e738890fad074a4d2831e2/packages/generator/poetry.lock#L4
> 
> Without adding the "black" in `[tool.poetry.dependencies]`, fresh machine testing throws error with `black package is missing`. So, adding "black" in dependencies to fix this.

Currently, black is in dev dependency, but we are using black in the Measurement Plug-In Client generator for formatting the generated code, which requires `black` to be added in package dependency. So, I have updated the `pyproject.toml` with `black`.

dixonjoel (approved):

PR #880, opened 2024-09-10
Added enum support to Measurement Client

MounikaBattu17:

Added enum support for Measurement Client
Implements [Task 2827579](https://dev.azure.com/ni/DevCentral/_workitems/edit/2827579): Extend support to enums.

With this implementation, Measurement Client supports measurements with enum parameters.
- Created a new enum class using enum values from metadata annotations.
- The new enum class name in created using parameter name suffixed with `Enum` in pascal case. Also, removed special characters from the class name.
  For example,
```
  parameter name = Enum In
  enum class name = EnumInEnum     # PascalCase(parameter name) + "Enum"
```
- Stored the created enum class in `enum_values_by_type_name` which acts as a map to eliminate repeating enums.
For example, Configuration `Enum In` and Output `Enum Out` will use the same Enum class, so `enum_values_by_type_name` will be useful to avoid creating duplicate enums for both parameters. 
- Applied `enum_type` while creating pools for the metadata.
- While black formatting the rendered template, replaced this enum class pattern.
  `enum_type=<enum 'EnumInEnum'>) -> enum_type= EnumInEnum`
Added automated tests.
Generated and verified the client with enum params like enums, enum arrays, and sparse enums.


MounikaBattu17:
We construct new enum class using enum values from `metadata.annotations`. Unlike enums, LabVIEW rings, doesn't pass ring values in `metadata.annotations`. Thus for rings, we have treated them as integers. So this is the behaviour or any enum and ring

**enum parameter in measurement:**
image
**enum parameter in python client:**
` u32_animal_enum_in_array: List[U32AnimalEnumInEnum] = [U32AnimalEnumInEnum.Gopher, U32AnimalEnumInEnum.Rabbi ]`

**ring parameter in measurement:**
image
image

**ring parameter in python client:**
` u32_color_ring_in: int = 1`

So, ring values can be only be set to integers in client, we cannot use the name (RED, GREEN, BLUE) here.

Please share your thoughts on this

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -284,3 +323,26 @@ def _get_python_type_as_str(type: Field.Kind.ValueType, is_array: bool) -> str:
...
+
+
+def _get_enum_type(parameter: Any, enum_values_by_type_name: Dict[str, Dict[str, Any]]) -> Any:
+    if parameter.type == FieldDescriptorProto.TYPE_ENUM:
+        loaded_enum_values = json.loads(parameter.annotations["ni/enum.values"])
+        enum_values = dict((key, value) for key, value in loaded_enum_values.items())
+
+        for existing_enum_name, existing_enum_values in enum_values_by_type_name.items():
+            if existing_enum_values == enum_values:
+                return Enum(existing_enum_name, existing_enum_values)
```
bkeryan, line 335: It's kind of weird to create the same enum type more than once. I'm not sure whether this returns the same class object or a new one. It would be less weird to cache the enum types in the dict too.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -284,3 +323,26 @@ def _get_python_type_as_str(type: Field.Kind.ValueType, is_array: bool) -> str:
...
+        return Enum(new_enum_type_name, enum_values)
+    return None
+
+
+def _get_enum_class_name(name: str) -> str:
+    class_name = name.title().replace(" ", "")
+    invalid_chars = "`~!@#$%^&*()-+={}[]\|:;',<>.?/ \n_"
+    pattern = f"[{re.escape(invalid_chars)}]"
+    class_name = re.sub(pattern, "", class_name)
+    return class_name + "Enum"
```
bkeryan, line 348: See comments on 2nd copy

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -157,6 +173,27 @@ class ${class_name}:
...
+            enum_values_by_type_name[new_enum_type_name] = enum_values
+            return Enum(new_enum_type_name, enum_values)
+        return None
+
+    def _get_enum_class_name(self, name: str) -> str:
+        class_name = name.title().replace(" ", "")
+        invalid_chars = "`~!@#$%^&*()-+={}[]\|:;',<>.?/ \n_"
+        pattern = f"[{re.escape(invalid_chars)}]"
+        class_name = re.sub(pattern, '', class_name)
+        return class_name + "Enum"
```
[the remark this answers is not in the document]
MounikaBattu17: > Maybe it's fine to turn camelCase and PascalCase into garbage, though? 

I wanted to create Enums in Pascal case based on other enums' naming conventions in the repository. Since `title()` doesn't work, I've created a method that filters out illegal characters, splits the string, and capitalizes each segment. This will convert any case inputs to Pascal case by excluding illegal characters. 
Since the enum class name is derived from measurement configuration or output parameter name, I hope we don't have to look for rare unusual cases. Please let me know your thoughts on this.

> Why are you creating enum classes at run time, anyway? The generated code already defines the enum classes in code. 

The generated code defines the enum classes, however, when using the client APIs, in `_create_file_descriptor`, we might not know the enum type for a particular enum parameter. Thus, we need to get the enum types at runtime also. 
At this stage, `configuration` is of type `measurement_service_pb2.ConfigurationParameter`, so we won't be having `enum_type` attribute in it.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -284,3 +323,26 @@ def _get_python_type_as_str(type: Field.Kind.ValueType, is_array: bool) -> str:
     if is_array:
         return f"List[{python_type.__name__}]"
     return python_type.__name__
+
+
+def _get_enum_type(parameter: Any, enum_values_by_type_name: Dict[str, Dict[str, Any]]) -> Any:
```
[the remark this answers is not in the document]
MounikaBattu17: Updated the method to proper type hints.
> The Any in Dict[str, Dict[str, Any]] is probably an int, right?

I verified enum parameters in python/LabVIEW measurements, which has validation to have enum value zero. Thus, I hope we can safely say as `int` type.

> parameter is a ParameterMetadata

The callers of this method pass two different types of data. One is of type `measurement_service_pb2.ConfigurationParameter` while the other is `ParameterMetadata`. So, I have the type hint as `Any`. Please share your thoughts on this.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -284,3 +323,26 @@ def _get_python_type_as_str(type: Field.Kind.ValueType, is_array: bool) -> str:
...
+        enum_values = dict((key, value) for key, value in loaded_enum_values.items())
+
+        for existing_enum_name, existing_enum_values in enum_values_by_type_name.items():
+            if existing_enum_values == enum_values:
+                return Enum(existing_enum_name, existing_enum_values)
+
+        new_enum_type_name = _get_enum_class_name(parameter.name)
+        enum_values_by_type_name[new_enum_type_name] = enum_values
+        return Enum(new_enum_type_name, enum_values)
+    return None
```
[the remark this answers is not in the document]
MounikaBattu17: The `enum_type` in `ParameterMetadata` accepts `None` for non-enum parameters and an `Enum` class for enum parameters. I initially designed the method to return either an `Enum` class or `None` based on the parameter type. However, this approach seemed confusing when None was returned. Therefore, I updated the method to return only the Enum class, with the expectation that callers will ensure they are passing only enum parameters to this method.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
MounikaBattu17: Thanks for pointing this out. I've refined this method such that we cache the created enum also, so that we won't have to create multiple enums.
Updated:
```
enum_values_by_type: Dict[str], Dict[str, int]]  -> enum_values_by_type: Dict[Type[Enum], Dict[str, int]] 
```

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
MounikaBattu17: Refactored this method based on your previous comments. Closing this and continuing the conversation in this [thread](https://github.com/ni/measurement-plugin-python/pull/880#discussion_r1755491884).

MounikaBattu17:
There is a mypy error and will be fixed soon.

MounikaBattu17:
> There is a mypy error and will be fixed soon.

Fixed.

The [error](https://github.com/ni/measurement-plugin-python/actions/runs/10848059636/job/30104419153#step:9:14) was on this line
`new_enum_type = Enum(new_enum_type_name, enum_values)   # error: Enum() expects a string literal as the first argument  [misc]`

seems like mypy static analysis checks for a string literal but we pass a variable.

To resolve this error, I've used enum `__call__` method directly to create enums dynamically at runtime.
`new_enum_type = Enum.__call__(new_enum_type_name, enum_values)`

bkeryan:
> > There is a mypy error and will be fixed soon.
> 
> Fixed.
> 
> The [error](https://github.com/ni/measurement-plugin-python/actions/runs/10848059636/job/30104419153#step:9:14) was on this line `new_enum_type = Enum(new_enum_type_name, enum_values) # error: Enum() expects a string literal as the first argument [misc]`
> 
> seems like mypy static analysis checks for a string literal but we pass a variable.
> 
> To resolve this error, I've used enum `__call__` method directly to create enums dynamically at runtime. `new_enum_type = Enum.__call__(new_enum_type_name, enum_values)`

This is an indication that we are using the enum functional API wrong.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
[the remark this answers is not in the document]
bkeryan: > Please let me know your thoughts on this.

My thoughts are that we are not giving users enough guidance on naming style, and this makes our lives harder when generating client code. Please make sure your team updates the documentation to offer better guidance.

As for the details: I'll look at those on Monday.

> The generated code defines the enum classes, however, when using the client APIs, in _create_file_descriptor, we might not know the enum type for a particular enum parameter. Thus, we need to get the enum types at runtime also.
At this stage, configuration is of type measurement_service_pb2.ConfigurationParameter, so we won't be having enum_type attribute in it.

Great, we have the information but we're not using it. Please structure the code to make this information available where it's needed, instead of regenerating it by using the Python enum functional API in a way that Mypy flags as incorrect usage.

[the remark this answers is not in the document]
MounikaBattu17: > we have the information but we're not using it. 

Yes! We missed this. We already have the data in` self._configuration_metadata` and `self._output_metadata`. While `_create_file_descriptor` instead of using the above data, we try to get the data using `GetMetadata`. So, I have removed the code where we try to get the metadata and used the already available information from `self._configuration_metadata` and `self._output_metadata`

MounikaBattu17:
> If we absolutely must use the enum functional API in this way, then I would prefer to use `# type: ignore[misc]` on the line that does this and put a comment above it saying why we have to do it this way.

Seems like MyPy static analysis throws an error if we don't use a direct string literal for dynamic Enum creation. So I have suppressed the error instead of using dunner method like you suggested.

Syntax like,
```
Enum('Color', {"NONE":0, "RED": 1}) -> no mypy error

enum_name: str = "Color"
Enum(enum_name, {"NONE":0, "RED": 1}) -> throws mypy error
```


> Also I don't think the generated code should be doing this at all. It has static definitions of all of the enum types and it should use them.

Removed and used the already available metadata.

dixonjoel (approved):

bkeryan (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -284,3 +335,35 @@ def _get_python_type_as_str(type: Field.Kind.ValueType, is_array: bool) -> str:
...
+
+
+def _get_enum_class_name(name: str) -> str:
+    name = re.sub(r"[^\w\s]", "", name).replace("_", " ")
+    split_string = name.split()
+    if len(split_string) > 1:
+        name = "".join(s.capitalize() for s in split_string)
+    else:
+        name = name[0].upper() + name[1:]
+    return name + "Enum"
```
jayaseelan-james, line 369: Prefer f-strings for better readability.

packages/generator/tests/utilities/measurements/non_streaming_data_measurement/__init__.py
```diff
@@ -52,6 +62,10 @@
     "IO Array In", nims.DataType.IOResourceArray1D, ["resource1", "resource2"]
 )
 @measurement_service.configuration("Integer In", nims.DataType.Int32, 10)
+@measurement_service.configuration("Enum In", nims.DataType.Enum, Color.BLUE, enum_type=Color)
+@measurement_service.configuration(
+    "Enum Array In", nims.DataType.EnumArray1D, [1, 2], enum_type=Color
+)
```
jayaseelan-james, line 68: It might be an over-ask but because we showcase "protobuf enums" in our sample measurement, should we also test the Protobuf Enum type with our client to ensure we're not missing anything on that part?

PR #882, opened 2024-09-10
fix: Measurement Plug-In Client Generator Python code injection

Jotheeswaran-Nandagopal:

- Add `repr()` function to the string parameters in the `mako template` to omit arbitrary Python code in the output.

[Bug 2843797](https://dev.azure.com/ni/DevCentral/_workitems/edit/2843797)
- If any malicious service returns metadata with escape sequences, that results in arbitrary Python code injection in the output. This implementation handles the escape sequences with `repr()`.

- Tested manually.

### DIFF
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
@@ -163,7 +163,7 @@ def create_client(
             file_name=f"{module_name}.py",
             directory_out=directory_out_path,
             class_name=class_name,
-            display_name=repr(metadata.measurement_details.display_name).strip("'").strip('"'),
+            display_name=metadata.measurement_details.display_name,
             configuration_metadata=configuration_metadata,
             output_metadata=output_metadata,
             service_class=service_class,
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
@@ -163,7 +163,7 @@ def create_client(
             file_name=f"{module_name}.py",
             directory_out=directory_out_path,
             class_name=class_name,
-            display_name=metadata.measurement_details.display_name,
+            display_name=repr(metadata.measurement_details.display_name).strip("'").strip('"'),
             configuration_metadata=configuration_metadata,
             output_metadata=output_metadata,
             service_class=service_class,
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -43,7 +43,7 @@ class Output(NamedTuple):
 % endif
 
 
-class ${class_name}:
+class ${repr(class_name)}:
     """Client to interact with the measurement plug-in."""
      
     def __init__(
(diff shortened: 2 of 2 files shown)


Jotheeswaran-Nandagopal:
When a malicious service returns metadata with escape sequences, that leads to vulnerability. 
**E.g.**,  `"This is a malicious function.\"\n        import os; os.system(\"rm -rf \")\n        test=\"" `

In this implementation, 

- The parameters in the `mako templates`, which returns string type are represented with `repr()` function, as the escape sequences in above example are considered as lin separation and ends up in arbitrary code. 
- The parameters of 'json' type are handling escape sequences by default.

The below is the image of the escape sequence handled output of the above example:
image

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -43,7 +43,7 @@ class Output(NamedTuple):
 % endif
 
 
-class ${class_name}:
+class ${repr(class_name)}:
```
bkeryan, line 46: Doesn't repr include the quotes? That shouldn't work: `class "MyClass":`

For class_name I think it's more appropriate to call isidentifier() and return an error if the class name is not a valid identifier. Maybe this is already covered?

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -161,7 +161,7 @@ class ${class_name}:
         self,
         ${configuration_parameters_with_type_and_default_values}
     ) -> ${output_type} :
-        """Executes the ${display_name}.
+        """Executes the ${repr(display_name)}.
```
bkeryan, line 164: This should not include the quotes either.

bkeryan (changes requested):

bkeryan:
> @bkeryan , I would like to know your suggestion on this fix for this Python code injection Bug.

I think using repr() is a good solution when you want to print a string literal. When you want to print an identifier, I think it's better to use isidentifier() and throw an error if the string isn't an identifier. For strings in comments, I think it's ok to use repr() and strip the quotes.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
Jotheeswaran-Nandagopal: Yes, we have validated with `isidentifier()` already. So I am removing the `repr` here.

Jotheeswaran-Nandagopal:
> I think using repr() is a good solution when you want to print a string literal. When you want to print an identifier, I think it's better to use isidentifier() and throw an error if the string isn't an identifier. For strings in comments, I think it's ok to use repr() and strip the quotes.

Sure @bkeryan. I have updated the following,
-> Removed `repr`,
- For `service class` as we are resolving the service and throwing error if it's invalid.
- For `class name` as we are validating the class name with `isidentifier`.

-> Updated the display name with `repr` and stripped the quotes, as we are retrieving the display name from metadata.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -163,7 +163,7 @@ def create_client(
             file_name=f"{module_name}.py",
             directory_out=directory_out_path,
             class_name=class_name,
-            display_name=metadata.measurement_details.display_name,
+            display_name=repr(metadata.measurement_details.display_name).strip("'").strip('"'),
```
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: I get that. I have updated the filtering in mako file as sugeested - `${display_name | repr}`. 
And, I have updated the documentation. 
As per the updated documentation, striping quotes will not be applicable. So, I have removed that logic.

bkeryan (approved):
Looks good, thanks!

PR #883, opened 2024-09-12
Add interactive mode for client generation

MounikaBattu17:

Facilities the measurement client generation in interactive mode.

This implementation facilitates the users to generate python client in interactive mode.
- With click-option-group, made 
   - `--all`, `--interactive`, `--measurement_service_class` as individual modes and mutually exclusive to each other.
   - `--module_name`, `--class_name`, `directory_out` as optional parameters.
- Now the command can be used as follows:
  - `ni-measurement-plugin-client-generator -a`  -> creates clients for all registered measurements.
   Optional parameters `-m`, `-c` will be ignored if specified.

  - `ni-measurement-plugin-client-generator -i`  -> creates clients for any registered measurements interactively.
   Optional parameters `-m`, `-c`, `-o` will be ignored if specified.

  - `ni-measurement-plugin-client-generator -s "ni.examples.SampleMeasurement_Python"`  -> creates clients for given measurement.
   Optional parameters `-m`, `-c`, `-o` are recommended in this mode.
       - To specify multiple clients for this mode use this
`ni-measurement-plugin-client-generator -s "ni.examples.SampleMeasurement_Python" -s "ni.examples.GameOfLife_Python"`
- The version for click-opt-group is `>=0.5.61`.
- Refactored `create_client` method  in `__init.py__`.
- Existing test updates and new tests for interactive mode will be covered as a part of another PR.

Verified commands manually.
Automated tests yet to be added.


MounikaBattu17:
Here is a small gif on the client creation through interactive workflow
Client-InteractiveMode

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -89,91 +290,66 @@ def create_client(
...
-        print(
-            f"The measurement plug-in client for the service class '{service_class}' has been created successfully."
-        )
+@click.group()
+def create_client() -> None:
+    """Generates a Python Measurement Plug-In Client module for the measurement service.
+
+    You can use the generated module to interact with the corresponding measurement service.
+
+    Use command 'b' to generate the client module in batch mode, or 'i' for interactive mode.
```
bkeryan, line 349: Commands should be words, not single letters.

Also, commands are typically verbs like "generate" or "build", not adjectives like "batch" or "interactive".

packages/generator/ni_measurement_plugin_sdk_generator/client/__main__.py
bkeryan, line 1: @mshafer-NI Do you have any suggestions about adding subcommands to an existing script? I don't like where this is going.

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/__main__.py
MounikaBattu17: @bkeryan 

We proposed the approach of having two separate commands for batch and interactive mode because we want to be clear to users on the usage of both modes. Combining both modes into a single command didn't make much sense to have options like module name, class name for interactive mode also.

> Do you have any suggestions about adding subcommands to an existing script?  I don't like where this is going.

Could you please provide more details on your concerns? We chose this approach to simplify running either mode from `main.py`. We also considered using argparse, but decided on click due to its advantages. If subcommands are not preferred, would it be viable to include only the batch mode in this script?

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,8 +47,215 @@ def _create_file(
...
+
+
 @click.command()
 @click.argument("measurement_service_class", nargs=-1)
+@click.option(
+    "-a",
+    "--all",
+    is_flag=True,
+    help="Creates Python Measurement Plug-In Client for all the registered measurement services.",
+)
```
mshafer-NI, line 258: Does this need to be moved?

packages/generator/ni_measurement_plugin_sdk_generator/client/__main__.py
MounikaBattu17: @bkeryan 
As per your suggestions, I've integrated the interactive mode as a click `option` within the existing `create_client `command. With this update, the command will now operate as follows:

```
>> ni-measurement-plugin-client-generator ni.examples.NIDCPowerSourceDCVoltage_Python -m "ni_dc_power_module" -c "NIDCPowerClass" -o "D:\"
The measurement plug-in client for the service class 'ni.examples.NIDCPowerSourceDCVoltage_Python' has been created successfully.

>> ni-measurement-plugin-client-generator -i
ni-measurement-plugin-client-generator -i
Creating the Python Measurement Plug-In Client in interactive mode...

>> ni-measurement-plugin-client-generator ni.examples.NIDCPowerSourceDCVoltage_Python -m "ni_dc_power_module" -i
Error: Interactive mode does not support additional parameters. Please remove any extra parameters and try again.

>> ni-measurement-plugin-client-generator
Creating the Python Measurement Plug-In Client in interactive mode...

```
cc: @mshafer-NI

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
MounikaBattu17: I rearranged such that it matches the method signature.
```
def create_client(
    measurement_service_class: List[str],
    all: bool,
    module_name: Optional[str],
    class_name: Optional[str],
    directory_out: Optional[str],
    interactive: bool,
) -> None:
```

packages/generator/ni_measurement_plugin_sdk_generator/client/__main__.py
mshafer-NI: From a CLI perspective and a use of Click perspective, this is much better.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,8 +47,215 @@ def _create_file(
...
+        raise click.ClickException(
+            f"The {name_type} name '{name}' is not a valid Python identifier."
+        )
+
+
+def _get_interactive_module_and_class_names() -> Tuple[str, str]:
+    module_name = click.prompt("Enter a name for the python client module", type=str)
+    _validate_identifier(module_name, "module")
+    class_name = click.prompt("Enter a name for the python client class", type=str)
+    _validate_identifier(class_name, "class")
```
bkeryan, line 73: In batch mode, these are optional, but in interactive mode they are required. That means using interactive mode requires the user to do more work than batch mode.

A good way to handle this is to calculate the default value based on the service class and include it in the prompt. If the user presses the enter key without entering a value, then use the default. `click.prompt` does this automatically if you specify `default=`.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,8 +47,215 @@ def _create_file(
...
+
+def _validate_identifier(name: str, name_type: str) -> None:
+    if not is_python_identifier(name):
+        raise click.ClickException(
+            f"The {name_type} name '{name}' is not a valid Python identifier."
+        )
+
+
+def _get_interactive_module_and_class_names() -> Tuple[str, str]:
+    module_name = click.prompt("Enter a name for the python client module", type=str)
```
bkeryan, line 70: Python is a proper noun, so it should be capitalized.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -59,23 +266,28 @@ def _create_file(
...
     help="Output directory for Measurement Plug-In Client files. Default: '<current_directory>/<module_name>'",
 )
+@click.option(
+    "-i",
+    "--interactive",
+    is_flag=True,
+    help=(
+        "Enables interactive mode for creating Measurement Plug-In Client files. "
+        "If no parameters are provided, this mode will be activated by default."
+        "When using this mode, no other parameters should be specified. "
```
bkeryan, line 281: Leave a space between sentences.
```
  -i, --interactive         Enables interactive mode for creating Measurement
                            Plug-In Client files. If no parameters are
                            provided, this mode will be activated by
                            default.When using this mode, no other parameters
                            should be specified.
  --help                    Show this message and exit.
```

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -89,91 +301,48 @@ def create_client(
...
-        if not measurement_service_class:
-            raise click.ClickException(
-                "The measurement service class cannot be empty. Either provide a measurement service class or use the 'all' flag to generate clients for all registered measurements."
-            )
-
-    if directory_out is None:
-        directory_out_path = pathlib.Path.cwd()
+    has_batch_parameters = (
+        measurement_service_class or all or module_name or class_name or directory_out
+    )
```
bkeryan, line 306: https://pypi.org/project/click-option-group/ could be helpful for this.
- It allows displaying options under a subheading such as "Batch mode configuration:"
- It can enforce that a group of options are mutually exclusive (such as `--interactive`, `--service-class`, and `--all`).
- It can enforce that at least one option from the group is set.
- I'm not sure, but it may be possible to make a group conflict with another group. @mshafer-NI said he'll look this up on Monday.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -89,91 +301,48 @@ def create_client(
...
-        )
-        configuration_metadata = get_configuration_metadata_by_index(metadata, service_class)
-        output_metadata = get_output_metadata_by_index(metadata)
-
-        configuration_parameters_with_type_and_default_values, measure_api_parameters = (
-            get_configuration_parameters_with_type_and_default_values(
-                configuration_metadata, built_in_import_modules
+            selection = (
+                click.prompt(
+                    "\nEnter 'yes' or 'y' to continue, or enter any other keys to exit",
```
bkeryan, line 337: This seems clunky to me. What if we asked the user to select another measurement (x to exit)?

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
bkeryan, line 1: A related idea: it would be nice if `--list` would print all of the registered service display names and service classes.

Avinash2Suresh: @MounikaBattu17 I also feel that this command will be helpful for users to understand the list of registered measurements.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -59,23 +266,28 @@ def _create_file(
...
     "--directory-out",
     help="Output directory for Measurement Plug-In Client files. Default: '<current_directory>/<module_name>'",
 )
+@click.option(
+    "-i",
+    "--interactive",
+    is_flag=True,
+    help=(
+        "Enables interactive mode for creating Measurement Plug-In Client files. "
+        "If no parameters are provided, this mode will be activated by default."
```
Avinash2Suresh, line 280: Do we need to make the `-i` command as optional? IMO, I would recommend throwing an error if the user does not provide any subcommand along with the `ni-measurement-plugin-client-generator` command.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -89,91 +301,48 @@ def create_client(
...
-    else:
-        if not measurement_service_class:
-            raise click.ClickException(
-                "The measurement service class cannot be empty. Either provide a measurement service class or use the 'all' flag to generate clients for all registered measurements."
-            )
-
-    if directory_out is None:
-        directory_out_path = pathlib.Path.cwd()
+    has_batch_parameters = (
+        measurement_service_class or all or module_name or class_name or directory_out
```
Avinash2Suresh, line 305: It should only take `measurement_service_class` or `all` into account.

In cases where users provide only the module name, or class name, or directory name, the has_batch_parameters boolean will still be set. However, without any of the following three commands (service_class, all, or interactive), we can't proceed with client generation.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
MounikaBattu17: Create a default value and specified the same in prompt message for clarity.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,8 +47,218 @@ def _create_file(
...
+
+
+def _create_client(
+    channel_pool: GrpcChannelPool,
+    discovery_client: DiscoveryClient,
+    built_in_import_modules: List[str],
+    custom_import_modules: List[str],
+    measurement_service_class: List[str] = [],
+    all: bool = False,
+    module_name: Optional[str] = "",
```
Jotheeswaran-Nandagopal, line 225: Can we change the default values to the optional parameters as None? As the None checks will fail because of this.
```suggestion
    module_name: Optional[str] = None,
    class_name: Optional[str] = None,
    directory_out: Optional[str] = None,
```

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -89,91 +304,48 @@ def create_client(
     built_in_import_modules: List[str] = []
     custom_import_modules: List[str] = []
```
Jotheeswaran-Nandagopal, line 305: Can we create these variables in the respective methods, where it is used and pass it  to other functions from there. Because, we are just passing it to `_create_client`.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -86,94 +308,44 @@ def create_client(
...
-                    module_name = camel_to_snake_case(base_service_class) + "_client"
-                if class_name is None:
-                    class_name = base_service_class.replace("_", "") + "Client"
-
-        if not is_python_identifier(module_name):
+        if has_batch_parameters and interactive:
             raise click.ClickException(
-                f"The module name '{module_name}' is not a valid Python identifier."
+                "Interactive mode does not support additional parameters. Please remove any extra parameters and try again."
             )
```
jayaseelan-james, line 329: Should we consider following an early exit pattern and check for this exception before executing anything else?

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -86,94 +308,44 @@ def create_client(
...
-                if module_name is None:
-                    module_name = camel_to_snake_case(base_service_class) + "_client"
-                if class_name is None:
-                    class_name = base_service_class.replace("_", "") + "Client"
-
-        if not is_python_identifier(module_name):
+        if has_batch_parameters and interactive:
             raise click.ClickException(
-                f"The module name '{module_name}' is not a valid Python identifier."
+                "Interactive mode does not support additional parameters. Please remove any extra parameters and try again."
```
jayaseelan-james, line 328:
```suggestion
                "Interactive mode does not accept additional parameters. Please remove any extra parameters and try again."
```

jayaseelan-james (changes requested):
I'd like to go over the PR with the latest changes once before approving.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
MounikaBattu17: We are working on the `optgroup` approach.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
MounikaBattu17: This option will be taken in a new PR.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -89,91 +301,48 @@ def create_client(
...
-                f"The class name '{class_name}' is not a valid Python identifier."
+        print("Creating the Python Measurement Plug-In Client in interactive mode...")
+        while True:
+            _create_client(
+                channel_pool=channel_pool,
+                discovery_client=discovery_client,
+                built_in_import_modules=built_in_import_modules,
+                custom_import_modules=custom_import_modules,
+                interactive_mode=True,
             )
```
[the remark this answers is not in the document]
MounikaBattu17: Agreed! The flow of mixing interactive mode and batch mode code was to reuse whatever code possible. But now we have split the modes into three (all, interactive, service class). Thus, I have refractored the code such that we can clearly understand what each mode is trying to do.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
bkeryan: Previously in chat:
> One other approach that some command line utilities use is to enter interactive mode when no arguments or options are specified. Tools like nslookup, netsh (Windows), lldpcli (Linux), etc. do this.

FYI @MounikaBattu17 This suggestion was really an alternative to an explicit `-i` option. Using an explicit `-i` option and returning an error if no options are specified is fine with me.

MounikaBattu17: Sure! Thanks for the confirmation. With the usage of click-option-group, we will be throwing error if required option (-a/-i/-s) is not specified.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -86,94 +212,110 @@ def create_client(
     """
     channel_pool = GrpcChannelPool()
     discovery_client = DiscoveryClient(grpc_channel_pool=channel_pool)
-    built_in_import_modules: List[str] = []
-    custom_import_modules: List[str] = []
 
     if all:
```
bkeryan, line 216: This function is getting big, so please extract each case into a helper function:
- `_create_all_clients`
- `_create_clients_interactively`
- `_create_clients`

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -86,94 +212,110 @@ def create_client(
...
-        measurement_service_class = get_all_registered_measurement_service_classes(discovery_client)
+        directory_out = _resolve_output_directory(directory_out)
```
bkeryan, line 217: FYI, this is identical in all 3 cases, but it will not be needed for `--list`, so I think it's ok to continue to duplicate it.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,8 +47,225 @@ def _create_file(
...
+        built_in_import_modules=to_ordered_set(built_in_import_modules),
+        custom_import_modules=to_ordered_set(custom_import_modules),
+    )
+
+    print(
+        f"The measurement plug-in client for the service class '{service_class}' has been created successfully."
+    )
+
+
+def _create_client(
```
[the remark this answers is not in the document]
MounikaBattu17: Agreed. But with the refractored code, these methods are removed.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
MounikaBattu17: With the usage of click-option-group, interactive mode is mutually exclusive of batch mode. So, I have removed this code.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
MounikaBattu17: Removed this code in favor of click-option-group.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -59,23 +276,28 @@ def _create_file(
...
     help="Output directory for Measurement Plug-In Client files. Default: '<current_directory>/<module_name>'",
 )
+@click.option(
+    "-i",
+    "--interactive",
+    is_flag=True,
+    help=(
+        "Enables interactive mode for creating Measurement Plug-In Client files. "
+        "If no parameters are provided, this mode will be activated by default. "
+        "When using this mode, no other parameters should be specified. "
```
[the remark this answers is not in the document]
MounikaBattu17: Now, we do not have interactive mode as default. Thus this help context is not needed.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,8 +47,225 @@ def _create_file(
...
+    discovery_client: DiscoveryClient,
+) -> List[str]:
+
+    if all or interactive:
+        measurement_service_class, measurement_display_name = get_all_registered_measurement_info(
+            discovery_client
+        )
+        if len(measurement_service_class) == 0:
+            raise click.ClickException("No registered measurements.")
+        if interactive:
```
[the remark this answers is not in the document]
MounikaBattu17: Refractored the method.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,8 +47,225 @@ def _create_file(
...
+    selection: int, measurement_service_classes: List[str]
+) -> List[str]:
+    if not (1 <= selection <= len(measurement_service_classes)):
+        raise click.ClickException(
+            f"Input {selection} is not invalid. Please try again by selecting a valid measurement from the list."
+        )
+    return [measurement_service_classes[selection - 1]]
+
+
+def _get_measurement_service_class(
```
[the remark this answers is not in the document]
MounikaBattu17: With the refractored code, this method is removed.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,8 +47,218 @@ def _create_file(
...
+
+def _create_module_name(base_service_class: str) -> str:
+    return camel_to_snake_case(base_service_class) + "_client"
+
+
+def _create_class_name(base_service_class: str) -> str:
+    return base_service_class.replace("_", "") + "Client"
+
+
+def _get_class_and_module_names(
```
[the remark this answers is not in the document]
MounikaBattu17: With the refactored code, this method is mremoved.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,8 +47,218 @@ def _create_file(
...
+
+
+def _validate_identifier(name: str, name_type: str) -> None:
+    if not is_python_identifier(name):
+        raise click.ClickException(
+            f"The {name_type} name '{name}' is not a valid Python identifier."
+        )
+
+
+def _get_interactive_module_and_class_names(base_service_class: str) -> Tuple[str, str]:
```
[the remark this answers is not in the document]
MounikaBattu17: With the refactored code, this method is removed.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
MounikaBattu17: Since this issue is minor and would only result in a slight increase in time to address, I’ll let it slide this time and avoid raising a separate PR for it. Moving forward, I’ll ensure that similar changes are handled more promptly.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
MounikaBattu17: Moved to helper methods.

packages/generator/pyproject.toml
```diff
@@ -25,6 +25,7 @@ click = ">=8.1.3"
 grpcio = "^1.49.1"
 protobuf = "^4.21"
 black = ">=24.8.0"
+click-option-group = "^0.5.6"
```
bkeryan, line 28: Use >= unless you have reason to believe that 0.6 will break compatibility.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,25 +48,265 @@ def _create_file(
...
+    print(
+        f"The measurement plug-in client for the service class '{measurement_service_class}' has been created successfully."
+    )
+
+
+def _create_all_clients(directory_out: str) -> None:
+    channel_pool = GrpcChannelPool()
+    discovery_client = DiscoveryClient(grpc_channel_pool=channel_pool)
+
+    if all:
```
bkeryan, line 157: `all` is undefined

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,25 +48,265 @@ def _create_file(
...
+                class_name=class_name,
+                directory_out=directory_out,
+            )
+
+
+def _create_clients_interactively() -> None:
+    print("Creating the Python Measurement Plug-In Client in interactive mode...")
+    channel_pool = GrpcChannelPool()
+    discovery_client = DiscoveryClient(grpc_channel_pool=channel_pool)
+    directory_out = _resolve_output_directory()
```
bkeryan, line 184: This ignores `--directory-out`. Is that intentional?

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
bkeryan: I mean combine this with the number prompt. `Select a measurement to generate a client (x to exit)`

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
bkeryan, line 1: mypy

```
Run poetry run mypy
ni_measurement_plugin_sdk_generator/client/__init__.py:100: error: Incompatible return value type (got "str", expected "list[str]")  [return-value]
ni_measurement_plugin_sdk_generator/client/__init__.py:157: error: Function "all" could always be true in boolean context  [truthy-function]
ni_measurement_plugin_sdk_generator/client/__init__.py:[15](https://github.com/ni/measurement-plugin-python/actions/runs/10889964246/job/30217686496?pr=883#step:9:16)8: error: Incompatible types in assignment (expression has type "Path", variable has type "str")  [assignment]
ni_measurement_plugin_sdk_generator/client/__init__.py:176: error: Argument "directory_out" to "_create_client" has incompatible type "str"; expected "Path"  [arg-type]
ni_measurement_plugin_sdk_generator/client/__init__.py:205: error: Argument 1 to "_extract_base_service_class" has incompatible type "list[str]"; expected "str"  [arg-type]
ni_measurement_plugin_sdk_generator/client/__init__.py:224: error: Argument "measurement_service_class" to "_create_client" has incompatible type "list[str]"; expected "str"  [arg-type]
ni_measurement_plugin_sdk_generator/client/__init__.py:249: error: Incompatible types in assignment (expression has type "Path", variable has type "str")  [assignment]
ni_measurement_plugin_sdk_generator/client/__init__.py:267: error: Argument "directory_out" to "_create_client" has incompatible type "str"; expected "Path"  [arg-type]
ni_measurement_plugin_sdk_generator/client/__init__.py:330: error: Argument 1 to "_create_all_clients" has incompatible type "Optional[str]"; expected "str"  [arg-type]
ni_measurement_plugin_sdk_generator/client/__init__.py:336: error: Argument "module_name" to "_create_clients" has incompatible type "Optional[str]"; expected "str"  [arg-type]
ni_measurement_plugin_sdk_generator/client/__init__.py:337: error: Argument "class_name" to "_create_clients" has incompatible type "Optional[str]"; expected "str"  [arg-type]
ni_measurement_plugin_sdk_generator/client/__init__.py:338: error: Argument "directory_out" to "_create_clients" has incompatible type "Optional[str]"; expected "str"  [arg-type]
Found 12 errors in 1 file (checked [17](https://github.com/ni/measurement-plugin-python/actions/runs/10889964246/job/30217686496?pr=883#step:9:18) source files)
```

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
MounikaBattu17: Fixed the mypy errors

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -5,6 +5,7 @@
 
 import black
 import click
+from click_option_group import optgroup, RequiredMutuallyExclusiveOptionGroup
```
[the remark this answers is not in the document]
MounikaBattu17: Added the package to dependencies.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,8 +47,225 @@ def _create_file(
...
+                type=int,
+            )
+            measurement_service_class = _get_selected_measurement_service_class(
+                selection, measurement_service_class
+            )
+
+    else:
+        if not measurement_service_class:
+            raise click.ClickException(
+                "The measurement service class cannot be empty. Either provide a measurement service class or use the 'all' flag to generate clients for all registered measurements."
```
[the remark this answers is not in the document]
MounikaBattu17: Removed this code in favor of click-option-group.

bkeryan (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
MounikaBattu17: As the approach got changed. Now we don't need sub commands

bkeryan: @MounikaBattu17 Please update the PR description accordingly.

MounikaBattu17: Updated.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,25 +48,256 @@ def _create_file(
...
-@click.argument("measurement_service_class", nargs=-1)
-@click.option(
+@optgroup.group(
+    "all-modes",
+    cls=RequiredMutuallyExclusiveOptionGroup,
+    help="The different modes to create Python measurement client.",
+)
+@optgroup.option(
+    "-s",
+    "--measurement_service_class",
```
bkeryan, line 270: Long options should use kebab-case, not snake_case. Click will automatically convert to snake_case variable names.
```suggestion
    "--measurement-service-class",
```

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -76,6 +308,7 @@ def create_client(
     module_name: Optional[str],
     class_name: Optional[str],
     directory_out: Optional[str],
+    interactive: bool,
```
Avinash2Suresh, line 311: This should be moved just above the optional parameters

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -84,96 +317,14 @@ def create_client(
     MEASUREMENT_SERVICE_CLASS: Accepts one or more measurement service classes.
     Separate each service class with a space.
```
Avinash2Suresh, line 318: with the latest update, this needs to be updated as we should give -s for each service class name.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -82,15 +82,20 @@ def get_measurement_service_stub(
...
     """Returns the service classes of all the registered measurement services."""
     registered_measurement_services = discovery_client.enumerate_services(
         _V2_MEASUREMENT_SERVICE_INTERFACE
     )
     measurement_service_classes = [
         measurement_service.service_class for measurement_service in registered_measurement_services
     ]
-    return measurement_service_classes
+    measurement_display_names = [
+        measurement_service.display_name for measurement_service in registered_measurement_services
```
Avinash2Suresh, line 96: we are retrieving both the display name and service class from the same object. So, we can use a single for loop to fetch both the information.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -82,15 +82,20 @@ def get_measurement_service_stub(
     return v2_measurement_service_pb2_grpc.MeasurementServiceStub(channel)
 
 
-def get_all_registered_measurement_service_classes(discovery_client: DiscoveryClient) -> List[str]:
+def get_all_registered_measurement_info(
+    discovery_client: DiscoveryClient,
+) -> Tuple[List[str], List[str]]:
     """Returns the service classes of all the registered measurement services."""
```
Avinash2Suresh, line 88: doc needs to update as it is returning the display name as well along with the service class.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,25 +48,256 @@ def _create_file(
...
+    )
+
+
+def _create_all_clients(directory_out: Optional[str]) -> None:
+    channel_pool = GrpcChannelPool()
+    discovery_client = DiscoveryClient(grpc_channel_pool=channel_pool)
+
+    directory_out_path = _resolve_output_directory(directory_out)
+    measurement_service_class, _ = get_all_registered_measurement_info(discovery_client)
+    if len(measurement_service_class) == 0:
```
Avinash2Suresh, line 159: Can we please keep this as a helper method and utilize it? We are using this validation in more than one place. So, in the future, if we want to modify the error message, we would need to change it in all occurrences.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,25 +48,256 @@ def _create_file(
...
+        )
+
+        base_service_class = _extract_base_service_class(service_class)
+        default_module_name = _create_module_name(base_service_class)
+        module_name = click.prompt(
+            "Enter a name for the Python client module (or) press enter to choose the default name",
+            type=str,
+            default=default_module_name,
+        )
+        _validate_identifier(module_name, "module")
```
Avinash2Suresh, line 213: What if the `default_module_name` we are providing is not a valid identifier? We are showing it to the user, and if it is not a valid identifier, we then throw an error. This doesn't seem valid to me. The same applies to the class name. This comment is specific to interactive mode, as we are showing the default value to the user and then validating it.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,25 +48,256 @@ def _create_file(
...
+    "-a",
+    "--all",
+    is_flag=True,
+    help="Creates Python Measurement Plug-In Clients for all registered measurement services.",
+)
+@optgroup.option(
+    "-i",
+    "--interactive",
+    is_flag=True,
+    help=("Creates Python Measurement Plug-In Clients interactively."),
```
Jotheeswaran-Nandagopal, line 284:
```suggestion
    help="Creates Python Measurement Plug-In Clients interactively.",
```

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,25 +48,256 @@ def _create_file(
...
+        if selection.lower() == "x":
+            break
+        service_class = _get_selected_measurement_service_class(
+            int(selection), measurement_service_class
+        )
+
+        base_service_class = _extract_base_service_class(service_class)
+        default_module_name = _create_module_name(base_service_class)
+        module_name = click.prompt(
+            "Enter a name for the Python client module (or) press enter to choose the default name",
```
jayaseelan-james, line 209:
```suggestion
            "Enter a name for the Python client module, or press Enter to use the default name.",
```

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,25 +48,256 @@ def _create_file(
...
+            type=str,
+            default=default_module_name,
+        )
+        _validate_identifier(module_name, "module")
+        default_class_name = _create_class_name(base_service_class)
+        class_name = click.prompt(
+            "Enter a name for the Python client class (or) press enter to choose the default name",
```
jayaseelan-james, line 216:
```suggestion
            "Enter a name for the Python client class, or press Enter to use the default name.",
```

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,25 +48,256 @@ def _create_file(
...
+    measurement_service_class: List[str],
+    module_name: Optional[str],
+    class_name: Optional[str],
+    directory_out: Optional[str],
+) -> None:
+    channel_pool = GrpcChannelPool()
+    discovery_client = DiscoveryClient(grpc_channel_pool=channel_pool)
+    directory_out_path = _resolve_output_directory(directory_out)
+
+    is_multiple_measurement_client_creation = len(measurement_service_class) > 1
```
jayaseelan-james, line 242: `is_multiple_measurement_client_creation` sounds too wordy.
```suggestion
    has_multiple_service_classes = len(measurement_service_class) > 1
```

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,25 +48,256 @@ def _create_file(
...
+            discovery_client=discovery_client,
+            measurement_service_class=service_class,
+            module_name=module_name,
+            class_name=class_name,
+            directory_out=directory_out_path,
+        )
+
+
+def _create_clients(
+    measurement_service_class: List[str],
```
jayaseelan-james, line 233: Probably I overlooked this in several instances but this name doesn't emphasize the fact that it holds a list of service class.
```suggestion
    measurement_service_classes: List[str],
```

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
jayaseelan-james: Technically, having a non-optionally field after optional field(s) is completely fine.

Example: https://github.com/ni/measurement-plugin-python/blob/09b40b03980d7208343787138ee83390e0b9610e/packages/service/ni_measurement_plugin_sdk_service/_drivers/_nidcpower.py#L30-L37

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,25 +48,256 @@ def _create_file(
...
+        service_class=measurement_service_class,
+        configuration_parameters_with_type_and_default_values=configuration_parameters_with_type_and_default_values,
+        measure_api_parameters=measure_api_parameters,
+        output_parameters_with_type=output_parameters_with_type,
+        built_in_import_modules=to_ordered_set(built_in_import_modules),
+        custom_import_modules=to_ordered_set(custom_import_modules),
+    )
+
+    print(
+        f"The measurement plug-in client for the service class '{measurement_service_class}' has been created successfully."
```
jayaseelan-james, line 149:
```suggestion
        f"The measurement plug-in client for the service class '{measurement_service_class}' is successfully created."
```

jayaseelan-james (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
Avinash2Suresh: Okay, The reasons is to sync with the same order that we have given to the click options.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -47,25 +48,256 @@ def _create_file(
...
 @click.command()
-@click.argument("measurement_service_class", nargs=-1)
-@click.option(
+@optgroup.group(
+    "all-modes",
+    cls=RequiredMutuallyExclusiveOptionGroup,
+    help="The different modes to create Python measurement client.",
+)
+@optgroup.option(
+    "-s",
```
Avinash2Suresh, line 269: Please make sure to update all the related test cases to pass the service class using -s sub command.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
MounikaBattu17: We can get them as `tuple` using a single for loop. By doing so, we need to separate service class and display names using another for loop in callers. So I think returning them as separate list here should be fine.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
MounikaBattu17: Anyway, it is upto the user to choose the incorrect name. We will still going to validate what users choose. So I hope we shouldn't do the early default name validation.

PR #887, opened 2024-09-13
Update default channel pool options when creating grpc channels

jasonmreding:

Updates the channel pool so that by default it:
- Does not limit the size of messages that can be sent across a channel.
- Disables http proxy if it determines the target address is for the localhost. DNS lookups are not performed as part of this calculation for simplicity and performance for what is expected to be the most common scenarios.

This minimizes support issues with http proxies that are often difficult to diagnose.

- Added unit tests for logic which determines whether an address string is for the localhost or not. 
- I manually tested with the `http_proxy` environment variable set before and after this change. I validated that example measurements fail to start successfully before the change (can't connect to discovery when trying to register themselves), and they start successfully after this change.

### DIFF
--- a/packages/service/ni_measurement_plugin_sdk_service/grpc/channelpool.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/grpc/channelpool.py
@@ -2,16 +2,13 @@
 
 from __future__ import annotations
 
+import ipaddress
+import re
 import sys
 from threading import Lock
 from types import TracebackType
-from typing import (
-    Dict,
-    Literal,
-    Optional,
-    Type,
-    TYPE_CHECKING,
-)
+from typing import TYPE_CHECKING, Dict, Literal, Optional, Type
+from urllib.parse import urlparse
 
 import grpc
 
@@ -57,7 +54,13 @@ class GrpcChannelPool(object):
         with self._lock:
             if target not in self._channel_cache:
                 self._lock.release()
-                new_channel = grpc.insecure_channel(target)
+                options = [
+                    ("grpc.max_receive_message_length", -1),
+                    ("grpc.max_send_message_length", -1),
+                ]
+                if self._is_local(target):
+                    options.append(("grpc.enable_http_proxy", 0))
+                new_channel = grpc.insecure_channel(target, options)
                 if ClientLogger.is_enabled():
                     new_channel = grpc.intercept_channel(new_channel, ClientLogger())
                 self._lock.acquire()
@@ -78,3 +81,32 @@ class GrpcChannelPool(object):
             for channel in self._channel_cache.values():
                 channel.close()
             self._channel_cache.clear()
+
+    def _is_local(self, target: str) -> bool:
+        hostname = ""
+        # First, check if the target string is in URL format
+        parse_result = urlparse(target)
+        if parse_result.scheme and parse_result.hostname and parse_result.port:
+            hostname = parse_result.hostname
+        else:
+            # Next, check for target string in <host_name>:<port> format
+            match = re.match(r"^(.*):(\d+)$", target)
+            if match:
+                hostname = match.group(1)
+
+        if not hostname:
+            return False
+        if hostname == "localhost" or hostname == "LOCALHOST":
+            return True
+        
+        # IPv6 addresses don't support parsing with leading/trailing brackets so we need to remove them.
+        match = re.match(r"^\[(.*)\]$", hostname)
+        if match:
+            hostname = match.group(1)
+
+        try:
+            address = ipaddress.ip_address(hostname)
+            return address.is_loopback
+        except:
+            pass
+        return False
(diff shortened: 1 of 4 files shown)


packages/service/ni_measurement_plugin_sdk_service/grpc/channelpool.py
```diff
@@ -78,3 +81,32 @@ def close(self) -> None:
...
+        match = re.match(r"^\[(.*)\]$", hostname)
+        if match:
+            hostname = match.group(1)
+
+        try:
+            address = ipaddress.ip_address(hostname)
+            return address.is_loopback
+        except:
+            pass
+        return False
```
bkeryan, line 112: I think it's appropriate to catch ValueError here. Some of the urllib.parse functions can return TypeError when you pass the wrong data type, or UnicodeError if you pass non-ASCII characters, and I don't think we should treat those situations as local addresses.

Also, consider moving the return into the except block.

```suggestion
        except ValueError:
            return False
```

packages/service/tests/unit/grpc/channelpool/test_channel_pool.py
```diff
@@ -0,0 +1,32 @@
...
+
+from ni_measurement_plugin_sdk_service.grpc.channelpool import GrpcChannelPool
+
+
+@pytest.mark.parametrize(
+    "target,expected_result",
+    [
+        ("127.0.0.1", False),
+        ("[::1]", False),
+        ("localhost", False),
```
bkeryan, line 11: These first 3 addresses look local, so please add an end-of-line comment to each one saying the port number is missing.

packages/service/ni_measurement_plugin_sdk_service/grpc/channelpool.py
```diff
@@ -57,7 +54,13 @@ def get_channel(self, target: str) -> grpc.Channel:
...
-                new_channel = grpc.insecure_channel(target)
+                options = [
+                    ("grpc.max_receive_message_length", -1),
+                    ("grpc.max_send_message_length", -1),
+                ]
+                if self._is_local(target):
+                    options.append(("grpc.enable_http_proxy", 0))
+                new_channel = grpc.insecure_channel(target, options)
                 if ClientLogger.is_enabled():
                     new_channel = grpc.intercept_channel(new_channel, ClientLogger())
```
bkeryan, line 65: Consider extracting this into a `_create_channel` method.

bkeryan (changes requested):

packages/service/tests/unit/grpc/channelpool/test_channel_pool.py
```diff
@@ -0,0 +1,32 @@
...
+
+from ni_measurement_plugin_sdk_service.grpc.channelpool import GrpcChannelPool
+
+
+@pytest.mark.parametrize(
+    "target,expected_result",
+    [
+        ("127.0.0.1", False), # Port must be specified explicitly
+        ("[::1]", False),     # Port must be specified explicitly
+        ("localhost", False), # Port must be specified explicitly
```
bkeryan, line 11:
```
./tests/unit/grpc/channelpool/test_channel_pool.py:9:31: BLK100 Black would make changes.
```

Run `poetry run ni-python-styleguide fix`.

packages/service/ni_measurement_plugin_sdk_service/grpc/channelpool.py
```diff
@@ -78,3 +73,43 @@ def close(self) -> None:
...
+            match = re.match(r"^(.*):(\d+)$", target)
+            if match:
+                hostname = match.group(1)
+
+        if not hostname:
+            return False
+        if hostname == "localhost" or hostname == "LOCALHOST":
+            return True
+
+        # IPv6 addresses don't support parsing with leading/trailing brackets so we need to remove them.
```
bkeryan, line 106:
```
./ni_measurement_plugin_sdk_service/grpc/channelpool.py:106:101: W505 doc line too long (104 > 100 characters)
```
Run `poetry run ni-python-styleguide fix`. If it doesn't update this, you may need to wrap it manually.

bkeryan (approved):
Approved with suggestions

packages/service/ni_measurement_plugin_sdk_service/grpc/channelpool.py
jasonmreding: Manually fixed since `ni-python-styleguide fix` didn't correct it. Is there something I should have run that would have detected that? Or maybe I just need to configure some editor settings?

bkeryan: `poetry run ni-python-styleguide lint`

I need to update https://github.com/ni/measurement-plugin-python/blob/main/CONTRIBUTING.md .

dixonjoel (approved):

PR #892, opened 2024-09-17
Add acceptance test to interactive mode of client generation

Jotheeswaran-Nandagopal:

- Added a couple of test cases for interactive mode of client generation in `test_client_generator.py`.
- Updated the click module tests with `CliRunner`.

- This implementation contains the automation testing for the interactive mode of Measurement Plug-In Client generation workflow.
- Removed the `systemExit` catching logics from the tests for modules using click under generator and replaced with `CLIRunner`.

- Existing automated tests passed.
- Newly added acceptance tests passed.


packages/generator/tests/acceptance/test_client_generator.py
```diff
@@ -71,6 +73,45 @@ def test___command_line_args___create_client_for_all_registered_measurements___r
...
+    with patch(
+        "click.prompt",
+        side_effect=[
+            "1",
+            "non_streaming_data_measurement_client",
+            "NonStreamingDataMeasurementClient",
+            "x",
+        ],
+    ):
+        with pytest.raises(SystemExit) as exc_info:
```
mshafer-NI, line 97: rather then catching the systemExit, let's use Click's testing utils -> https://dev.azure.com/ni/DevCentral/_git/Tools?path=/eng/pipeline/azdo_cli/tests/test_agents.py&version=GC3143bc8176382ff7809218dc6f83f2ad5e69a589&line=251&lineEnd=253&lineStartColumn=1&lineEndColumn=6&lineStyle=plain&_a=contents

(docs) https://click.palletsprojects.com/en/8.1.x/testing/

I usually like to make this a  fixture:
```python
@pytest.fixture()
def cli_runner():
    """Provide helper for calling the cli in tests."""
    runner = CliRunner(mix_stderr=False)
    yield functools.partial(runner.invoke, my_module.__main__.cli, catch_exceptions=False, standalone_mode=False)
```

packages/generator/tests/acceptance/test_client_generator.py
```diff
@@ -72,6 +74,45 @@ def test___command_line_args___create_client_for_all_registered_measurements___r
     )
 
 
+@patch("pathlib.Path.cwd")
```
bkeryan, line 77: This project uses [pytest-mock](https://pypi.org/project/pytest-mock/). Don't use patch directly.

packages/generator/tests/acceptance/test_client_generator.py
```diff
@@ -72,6 +74,45 @@ def test___command_line_args___create_client_for_all_registered_measurements___r
...
+) -> None:
+    temp_directory = tmp_path_factory.mktemp("measurement_plugin_client_files")
+    mock_directory_out.return_value = temp_directory
+    golden_path = test_assets_directory / "example_renders" / "measurement_plugin_client"
+    filename = "non_streaming_data_measurement_client.py"
+
+    with patch(
+        "click.prompt",
+        side_effect=[
+            "1",
```
bkeryan, line 92: Does this fail if there are any other services?

packages/generator/tests/acceptance/test_client_generator.py
```diff
@@ -72,6 +74,45 @@ def test___command_line_args___create_client_for_all_registered_measurements___r
...
+        golden_path / filename,
+        temp_directory / filename,
+    )
+
+
+def test___command_line_args_without_registering_any_measurement___create_client_using_interactive_mode___raises_exception() -> (
+    None
+):
+    with pytest.raises(SystemExit):
+        with pytest.raises(click.ClickException):
```
bkeryan, line 112: Why are you nesting pytest.raises?

bkeryan (changes requested):

packages/generator/tests/acceptance/test_client_generator.py
```diff
@@ -3,7 +3,9 @@
 import re
 import sys
 from typing import Generator
+from unittest.mock import Mock, patch
```
bkeryan, line 6: Mocks are for unit tests, not acceptance tests. Make this a unit test.

packages/generator/tests/acceptance/test_client_generator.py
Jotheeswaran-Nandagopal: Updated the tests with cli_runner.

packages/generator/tests/acceptance/test_client_generator.py
Jotheeswaran-Nandagopal: I have avoided patch, as this is acceptance test.

packages/generator/tests/acceptance/test_client_generator.py
Jotheeswaran-Nandagopal: I have nested, to capture 2 exceptions. Now, I have updated the `systemExit` captures with `CLIRunner`, so these exceptions can be caught with that.

packages/generator/tests/acceptance/test_client_generator.py
Jotheeswaran-Nandagopal: I understand that. As, this client generation requires `enumerate services`, `resolve service` and `get metadata` from running measurements. Here, mocking is adapted to set the temp directory as directory out. So, instead of mocking all the services for getting desired path, I have used used `os.chdir` to set temp directory as directory out for testing purpose.

packages/generator/tests/conftest.py
```diff
@@ -37,3 +41,19 @@ def discovery_service_process() -> Generator[DiscoveryServiceProcess, None, None
 def pin_map_directory(test_assets_directory: pathlib.Path) -> pathlib.Path:
     """Test fixture that returns the pin map directory."""
     return test_assets_directory / "pin_map"
+
+
+@pytest.fixture(scope="session")
+def create_client() -> Generator[functools.partial[Result], Any, None]:
```
bkeryan, line 47: `functools.partial[Result]` is implementation-specific and it's kind of a weak type hint, because it doesn't describe the allowed arguments. I would prefer to annotate this as `Callable[[Sequence[str]],Result]` or whatever.

bkeryan, line 47: The middle type parameter for `Generator` is the "send type". This is only used for coroutines. You don't "send" to this generator, so use `None` for the 2nd and 3rd type parameters.

packages/generator/tests/acceptance/test_plugin_generator.py
```diff
@@ -1,35 +1,36 @@
...
-                "MeasurementUI.measui",
-                "--service-class",
-                "SampleMeasurement_Python",
-                "--description-url",
-                "https://www.example.com/SampleMeasurement.html",
-                "--directory-out",
-                temp_directory,
-            ]
-        )
+    result: Result = create_measurement(
```
bkeryan, line 17: This type hint should not be needed. Type checkers should infer it based on the return type of `create_measurement`, and this isn't a documented attribute/constant.

packages/generator/tests/acceptance/test_plugin_generator.py
```diff
@@ -38,36 +39,37 @@ def test___command_line_args___create_measurement___render_without_error(
 
 
 def test___command_line_args___create_measurement_with_annotations___render_without_exception(
```
bkeryan, line 41: change "render_without_exception" to "render_without_error"

packages/generator/tests/acceptance/test_plugin_generator.py
```diff
@@ -38,36 +39,37 @@ def test___command_line_args___create_measurement___render_without_error(
...
-                "Measurement.Collection",
-                "--tags",
-                "M1",
-                "--tags",
-                "M2",
-                "--tags",
-                "M3",
-            ]
-        )
+    create_measurement(
```
bkeryan, line 48: Save `result` and check the exit code to verify it rendered without errors.

packages/generator/tests/acceptance/test_client_generator.py
```diff
@@ -22,41 +25,40 @@ def test___command_line_args___create_client___render_without_error(
...
-                module_name,
-                "--class-name",
-                "NonStreamingDataMeasurementClient",
-                "--directory-out",
-                temp_directory,
-            ]
-        )
-
-    assert not exc_info.value.code
+    result: Result = create_client(
```
bkeryan, line 28: ditto

packages/generator/tests/acceptance/test_client_generator.py
bkeryan: You are still using `unittest.mock` to patch the `click.prompt` function.

Read https://click.palletsprojects.com/en/8.1.x/testing/#input-streams for an alternative that is acceptable in acceptance/component tests.

packages/generator/tests/acceptance/test_plugin_generator.py
Jotheeswaran-Nandagopal: Updated to "render_without_error".

packages/generator/tests/acceptance/test_client_generator.py
Jotheeswaran-Nandagopal: removed the similar type hints throughout the file.

packages/generator/tests/conftest.py
Jotheeswaran-Nandagopal: Sure, I have updated to `Callable[[Sequence[str]],Result]`.

packages/generator/tests/acceptance/test_client_generator.py
```diff
@@ -65,36 +65,70 @@ def test___command_line_args___create_client_for_all_registered_measurements___r
...
+    filename = "non_streaming_data_measurement_client.py"
+    inputs = [
+        "1",
+        "non_streaming_data_measurement_client",
+        "NonStreamingDataMeasurementClient",
+        "x",
+    ]
+    os.chdir(temp_directory)
+
+    result = create_client(["--interactive"], input="\n".join(inputs))
```
bkeryan, line 92:
```
tests/acceptance/test_client_generator.py:92: error: Unexpected keyword argument "input"  [call-arg]
```

The correct way to type hint this is with a protocol:
```
class CreateClientFunction(Protocol):
    def __call__(self, args: Sequence[str], input: Optional[str] = None): ...
```

```
create_client: CreateClientFunction,
```

packages/generator/tests/acceptance/test_client_generator.py
Jotheeswaran-Nandagopal: I have removed the unittest.mock and updated with `input` to pass the inputs.

packages/generator/tests/acceptance/test_client_generator.py
Jotheeswaran-Nandagopal: Thanks for mentioning, I have added updated this by giving a common name `CliRunnerFunction` and used the same for `create_measurement` too.

bkeryan (approved):

PR #889, opened 2024-09-16
Update measurement client generator to interact with C# measurement services

Jotheeswaran-Nandagopal:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/measurement-services-python/blob/main/CONTRIBUTING.md).

### What does this Pull Request accomplish?

- Add `type_url` to `MeasureRequest.configuration_parameters`, while creating measure request, as the C# measurement services validates `type_url`. 

### Why should this Pull Request be merged?

- As, the C# measurement services validates `type_url`, the generated measurement client throws error while calling measure API  ([Bug #851](https://github.com/ni/measurement-plugin-python/issues/851)). 
- This implementation contains the fix to make generated measurement client to interact with C# measurement.

### What testing has been done?

- Existing automated test cases passed.
- Done manual testing.

PR #900, opened 2024-09-19
Add test case for measurements with no output configuration in client generation acceptance test

Jotheeswaran-Nandagopal:

- Added a measurement with no output configurations to test utilities.
- Added a test case for measurements with no output configuration in `test_client_generator.py`.
- Added a if condition in `mako`, to remove missing imports during the generation of measurements with no output configuration.
- Remove duplicate imports in `mako`.

- Added a new test case to verify the client generation for the measurements with no output configuration.
- Handle unused imports in `mako`. As, some imports are unused in the client module for the measurements with no output configuration.
- Remove duplicate imports i.e., Path is imported twice, as `pathlib` is already added in `mako`, also, we are adding `pathlib`, during client creation of measurement with path configurations.

### DIFF
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -8,7 +8,12 @@ import threading
 from enum import Enum
 % endif
 from pathlib import Path
-from typing import Any, Generator${", Iterable" if output_metadata else ""}, List${", NamedTuple" if output_metadata else ""}, Optional
+<%
+    typing_imports = ["Any", "Generator", "List", "Optional"]
+    if output_metadata:
+        typing_imports += ["Iterable", "NamedTuple"]
+%>\
+from typing import ${", ".join(sorted(typing_imports))}
 
 import grpc
 from google.protobuf import any_pb2, descriptor_pool
--- a/packages/generator/tests/acceptance/test_client_generator.py
+++ b/packages/generator/tests/acceptance/test_client_generator.py
@@ -67,7 +67,7 @@ def test___command_line_args_for_void_measurement___create_client___render_witho
             "--class-name",
             "VoidMeasurementClient",
             "--directory-out",
-            temp_directory,
+            str(temp_directory),
         ]
     )
 
--- a/packages/generator/tests/acceptance/test_client_generator.py
+++ b/packages/generator/tests/acceptance/test_client_generator.py
@@ -9,14 +9,18 @@ from ni_measurement_plugin_sdk_service.measurement.service import MeasurementSer
 
 from tests.conftest import CliRunnerFunction
 from tests.utilities.discovery_service_process import DiscoveryServiceProcess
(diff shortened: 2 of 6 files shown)


packages/generator/tests/test_assets/example_renders/measurement_plugin_client/void_measurement_client.py
```diff
@@ -0,0 +1,221 @@
...
+            input_metadata=list(self._configuration_metadata.values()),
+            output_metadata=list(self._output_metadata.values()),
+            service_name=self._service_class,
+            pool=descriptor_pool.Default(),
+        )
+
+    def _create_measure_request(
+        self, parameter_values: List[Any]
+    ) -> v2_measurement_service_pb2.MeasureRequest:
+        serialized_configuration = any_pb2.Any(
```
Avinash2Suresh, line 156: we would need to add type url here once the other PR for [C# service support](https://github.com/ni/measurement-plugin-python/pull/889/) goes in.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -3,19 +3,15 @@
...
 import threading
 % if len(enum_by_class_name):
 from enum import Enum
 % endif
-% for module in built_in_import_modules:
-${module}
-% endfor
-from typing import Any, Generator, Iterable, List, NamedTuple, Optional
+from pathlib import Path
+from typing import Any, Generator${", Iterable" if output_metadata else ""}, List${", NamedTuple" if output_metadata else ""}, Optional
```
bkeryan, line 11: Let's use a Python block to make this easier to maintain.
```suggestion
<%
    typing_imports = ["Any", "Generator", "List", "Optional"]
    if output_metadata:
        typing_imports += ["Iterable", "NamedTuple"]
%>\
from typing import ${", ".join(sorted(typing_imports))}
```

packages/generator/tests/acceptance/test_client_generator.py
```diff
@@ -9,14 +9,18 @@
...
-from tests.utilities.measurements import non_streaming_data_measurement, streaming_data_measurement
+from tests.utilities.measurements import (
+    non_streaming_data_measurement,
+    streaming_data_measurement,
+    void_measurement,
+)
 
 
-def test___command_line_args___create_client___render_without_error(
+def test___command_line_args_for_non_streaming_measurement___create_client___render_without_error(
```
bkeryan, line 19: If every precondition says "command line args" then you can remove it from all of them. Just describe what's different.

- non_streaming_measurement
- void_measurement
- all_registered_measurements
- interactive_mode
- etc.

bkeryan (changes requested):

packages/generator/tests/acceptance/test_client_generator.py
Jotheeswaran-Nandagopal: I have updated the names of the tests by removing `command_line_args`.

bkeryan (approved):

PR #901, opened 2024-09-19
Code quality improvements to the Measurement Plug-In Client generator

Jotheeswaran-Nandagopal:

- Update docstring for `pin_map_context` in `mako` file.
- Move helper method from `__init__.py` to `_support.py` in client generator.
- Add `protobuf enum` to `non_streaming_data_measurement` for testing.
- Remove `os.linesep` and `tab spaces` for output parameters, and handle the formatting in mako template using for loop.
- Update the `docstring` for `create_client` method.

This Pull Request focuses on improving the overall quality and performance of the client generator code.
- Refactoring of the code.
- Adding an extra test case `i.e., Protobuf enums`.
- Correcting link text in README files.
- The `create_client` method's docstring is outdated. So, I have updated it based on the new change of moving the `measurement-service-class` to option from argument.

- Existing test cases are passed.


jayaseelan-james (approved):

packages/generator/tests/utilities/measurements/non_streaming_data_measurement/_stubs/color_pb2.py
```diff
@@ -0,0 +1,29 @@
...
+from google.protobuf import descriptor_pool as _descriptor_pool
+from google.protobuf import symbol_database as _symbol_database
+
+# @@protoc_insertion_point(imports)
+
+_sym_db = _symbol_database.Default()
+
+
+DESCRIPTOR = _descriptor_pool.Default().AddSerializedFile(
+    b"\n\x0b\x63olor.proto\x12\x34ni.measurementlink.measurement.sample_measurement.v1*9\n\rProtobufColor\x12\x08\n\x04NONE\x10\x00\x12\x08\n\x04PINK\x10\x01\x12\t\n\x05WHITE\x10\x02\x12\t\n\x05\x42LACK\x10\x03\x42\x37\xaa\x02\x34NationalInstruments.MeasurementServices.Measurementsb\x06proto3"
```
jayaseelan-james, line 16: Don't copy the stubs from the other directories. Please make a copy of the `color.proto` file here and regenerate the stubs.

jayaseelan-james (changes requested):
Regenerate the stubs for the non-streaming measurement.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -244,10 +243,8 @@ def get_configuration_parameters_with_type_and_default_values(
         configuration_parameters.append(f"{parameter_name}: {parameter_type} = {default_value}")
 
     # Use line separator and spaces to align the parameters appropriately in the generated file.
-    configuration_parameters_with_type_and_value = f",{os.linesep}        ".join(
-        configuration_parameters
-    )
-    parameter_names_as_str = ", ".join(parameter_names)
+    configuration_parameters_with_type_and_value = f",".join(configuration_parameters)
+    parameter_names_as_str = ",".join(parameter_names)
```
bkeryan, line 247: I don't think generating worse output prior to reformatting with Black is an improvement. Revert this.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -284,7 +281,7 @@ def get_output_parameters_with_type(
 
         output_parameters_with_type.append(f"{parameter_name}: {parameter_type}")
 
-    return f"{os.linesep}    ".join(output_parameters_with_type)
+    return f";".join(output_parameters_with_type)
```
bkeryan, line 284: I don't think generating worse output prior to reformatting with Black is an improvement. Revert this.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -39,18 +39,13 @@ def _render_template(template_name: str, **template_args: Any) -> bytes:
...
-
 def _create_file(
     template_name: str, file_name: str, directory_out: pathlib.Path, **template_args: Any
 ) -> None:
     output_file = directory_out / file_name
 
     output = _render_template(template_name, **template_args).decode("utf-8")
-    output = _replace_enum_class_type(output)
+    output = replace_enum_class_type(output)
     formatted_output = black.format_str(
```
bkeryan, line 49: I overlooked this in the previous reviews. Why are you replacing the type at this level? Doesn't the generator know when it's generating a type?

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -356,6 +353,12 @@ def validate_measurement_service_classes(measurement_service_classes: List[str])
         raise click.ClickException("No registered measurements.")
 
 
+def replace_enum_class_type(output: str) -> str:
+    """Formats the data with enum type."""
```
bkeryan, line 357: This docstring is vague and having an input named "output" doesn't help.

packages/generator/pyproject.toml
```diff
@@ -75,3 +75,6 @@ skips = [
   "B101", # assert_used
   "B702", # use_of_mako_templates
 ]
+
+[tool.ni-python-styleguide]
```
bkeryan, line 79: Keep tool.black and tool.ni-python-styleguide together and keep their excludes in sync. Note that black uses regex instead of comma-separated list.

Example: https://github.com/ni/measurement-plugin-python/blob/main/packages/service/pyproject.toml#L1

packages/generator/tests/utilities/measurements/non_streaming_data_measurement/_stubs/color_pb2.pyi
bkeryan, line 1: Automate the protobuf codegen and update the build to check whether the protobuf codegen is out of date.

I don't want to have to guess what command line options you used to generate this, and I want to be able to easily regenerate it when upgrading grpcio-tools.

It would be nice if this also covered the example that you copied from.

See
- https://github.com/ni/measurement-plugin-python/blob/main/packages/service/scripts/generate_grpc_stubs.py
- https://github.com/ni/measurement-plugin-python/blob/main/.github/workflows/check_nims.yml#L82

bkeryan, line 1: Where is color.proto?

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
MounikaBattu17: Yes, the generator knows the enum type. The enum_type will be like this
`configuration_metadata = ParameterMetadata.Initialize(...,enum_type=Color)`

But when after rendering we see the output like this
```
self._configuration_metadata = {
    1: ParameterMetadata(
        display_name="Float In",
        type=2,
        repeated=False,
        default_value=0.05999999865889549,
        annotations={},
        message_type="",
        field_name="Float_In",
        enum_type=None,
    ),
    6: ParameterMetadata(
        display_name="Enum Array In",
        type=14,
        repeated=True,
        default_value=[3, 2],
        annotations={
            "ni/enum.values": '{"NONE": 0, "RED": 1, "GREEN": 2, "BLUE": 3}',
            "ni/type_specialization": "enum",
        },
        message_type="",
        field_name="Enum_Array_In",
        enum_type=<enum 'EnumInEnum'>,     # This representation instead of just Enum name.
    ),
}
```
Black formatting fails because of `enum_type=<enum 'EnumInEnum'>` where it has to be `enum_type=EnumInEnum`. So we are fixing this error here.

bkeryan: Generating wrong output and then fixing it up later by running a regex over the whole output file is a big hack.

At the very least, reduce the scope of the hack by moving it to the code that generates the metadata: https://github.com/ni/measurement-plugin-python/blob/main/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako#L93

A more robust fix would be to add code to the template to generate the metadata namedtuple, rather than relying on Python's str/repr implementation.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: Updated to send `List[str]`, suggested the following comment - [Link](https://github.com/ni/measurement-plugin-python/pull/901#discussion_r1766999990).

packages/generator/tests/utilities/measurements/non_streaming_data_measurement/_stubs/color_pb2.py
Jotheeswaran-Nandagopal: I added a copy of `color.proto` and generated the stubs using the scripts.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Jotheeswaran-Nandagopal: Updated the separator to ", ".

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Jotheeswaran-Nandagopal: Updated the method to return a list and used for loop in mako to generate the Output class.

packages/generator/tests/utilities/measurements/non_streaming_data_measurement/_stubs/color_pb2.pyi
Jotheeswaran-Nandagopal: I have copied it to this directory and updated the stubs now.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -248,10 +247,8 @@ def get_configuration_parameters_with_type_and_default_values(
 
         configuration_parameters.append(f"{parameter_name}: {parameter_type} = {default_value}")
 
-    # Use line separator and spaces to align the parameters appropriately in the generated file.
-    configuration_parameters_with_type_and_value = f",{os.linesep}        ".join(
-        configuration_parameters
-    )
+    # Join the list of configuration parameters into a comma-separated string.
```
jayaseelan-james, line 250: Probably this comment is redundant now as it simply restates the obvious.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -262,7 +259,7 @@ def get_output_parameters_with_type(
     built_in_import_modules: List[str],
     custom_import_modules: List[str],
     enum_values_by_type: Dict[Type[Enum], Dict[str, int]] = {},
-) -> str:
+) -> List[str]:
     """Returns the output parameters of the measurement with type."""
     output_parameters_with_type = []
```
jayaseelan-james, line 264:
```suggestion
    output_parameters_with_type: List[str] = []
```

packages/generator/tests/utilities/measurements/non_streaming_data_measurement/_stubs/color.proto
```diff
@@ -0,0 +1,21 @@
+//---------------------------------------------------------------------
+//---------------------------------------------------------------------
+syntax = "proto3";
+
+//---------------------------------------------------------------------
+//---------------------------------------------------------------------
+package ni.measurementlink.measurement.sample_measurement.v1;
```
jayaseelan-james, line 7: Update this package with the actual measurement name.
```suggestion
package ni.measurementlink.measurement.non_streaming_measurement.v1;
```

packages/generator/tests/utilities/measurements/non_streaming_data_measurement/_stubs/color.proto
```diff
@@ -0,0 +1,21 @@
...
+
+//---------------------------------------------------------------------
+//---------------------------------------------------------------------
+option csharp_namespace = "NationalInstruments.MeasurementServices.Measurements";
```
jayaseelan-james, line 11: Given that this proto will be used only for Python tests, should we remove this?

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Jotheeswaran-Nandagopal: Removed the comment.

packages/generator/tests/utilities/measurements/non_streaming_data_measurement/_stubs/color.proto
Jotheeswaran-Nandagopal: Sure removing it.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Jotheeswaran-Nandagopal: I have updated the the input parameter name and docstring. Also, I have moved this method to `mako` as per the suggestion from other [comment](https://github.com/ni/measurement-plugin-python/pull/901#discussion_r1767497806)

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
Jotheeswaran-Nandagopal: I have updated by moving this conversion to `mako` template to update the `enum_type` of `configuration_metadata` and `output_metadata`.

jayaseelan-james (approved):

packages/generator/tests/utilities/measurements/non_streaming_data_measurement/_stubs/color_pb2.pyi
Jotheeswaran-Nandagopal: I have implemented the script for automation of the stubs in generator and added the build check in pipeline.
I copied the proto from `sample_measurement` in the examples, the automation of stub generation for that will be tracked in this [issue](https://github.com/ni/measurement-plugin-python/issues/903).

dixonjoel (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -1,5 +1,20 @@
...
+    def _replace_enum_class_type(input_string: str) -> str:
+        """Replace enum class type representation with the enum name."""
+        pattern = r"<enum '([^']+)'>"
+        return re.sub(pattern, r"\1", input_string)
+
+    configuration_metadata = _replace_enum_class_type(str(configuration_metadata))
+    if output_metadata:
+        output_metadata = _replace_enum_class_type(str(output_metadata))
+%>
+
```
bkeryan, line 17: I don't think you should throw normal mako practices out the window just because you're reformatting the output with black.

```suggestion
%>\
\
<%page args="class_name, display_name, configuration_metadata, output_metadata, service_class, configuration_parameters_with_type_and_default_values, measure_api_parameters, output_parameters_with_type, built_in_import_modules, custom_import_modules, enum_by_class_name, configuration_parameters_type_url"/>\
\
<%
    def _replace_enum_class_type(input_string: str) -> str:
        """Replace enum class type representation with the enum name."""
        pattern = r"<enum '([^']+)'>"
        return re.sub(pattern, r"\1", input_string)

    configuration_metadata = _replace_enum_class_type(str(configuration_metadata))
    if output_metadata:
        output_metadata = _replace_enum_class_type(str(output_metadata))
%>\
\
```

bkeryan (approved):
Approved with suggestions

PR #902, opened 2024-09-19
[Minor] Raise exception during Client creation with Double XY datatype as input configuration

Jotheeswaran-Nandagopal:

- Raise an exception to show users that Double XY datatype is not supported, when creating a client for measurements configured with Double XY datatype as inputs.

- Currently, Measurement Plug-In client generator doesn't support Double XY datatype. So, this implementation will state a exact issue of the client generation failure, instead of error stack trace.

- Done manual testing.
- Existing test cases passed.

### DIFF
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
@@ -226,7 +226,7 @@ def get_configuration_parameters_with_type_and_default_values(
 
         if metadata.message_type and metadata.message_type == "ni.protobuf.types.DoubleXYData":
             raise click.ClickException(
-                "Measurement configuration(s) with Double XY datatype is not supported."
+                "Measurement configuration with Double XY datatype is not supported."
             )
 
         if metadata.annotations and metadata.annotations.get("ni/type_specialization") == "enum":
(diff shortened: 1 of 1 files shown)


packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -224,6 +224,11 @@ def get_configuration_parameters_with_type_and_default_values(
             else:
                 default_value = f"Path({default_value})"
 
+        if metadata.message_type and metadata.message_type == "ni.protobuf.types.DoubleXYData":
```
bkeryan, line 227: What message types are supported? Should this error for any "truthy" message_type?

Jotheeswaran-Nandagopal: No message types are supported, as only `DoubleXYData` has the type `type_pb2.Field.TYPE_MESSAGE`. To be more specific for `DoubleXYData` type, I added the second check. I will remove that.

Jotheeswaran-Nandagopal: I have updated the if condition to check "truthy" message_type.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -224,6 +224,11 @@ def get_configuration_parameters_with_type_and_default_values(
             else:
                 default_value = f"Path({default_value})"
 
+        if metadata.message_type:
+            raise click.ClickException(
+                "Measurement configuration with Double XY datatype is not supported."
```
bkeryan, line 229:
```suggestion
                "Measurement configuration with message datatype is not supported."
```

bkeryan (approved):
Approved with suggestions

PR #907, opened 2024-09-20
Do not specify version in the MeasurementService constructor

dixonjoel:

Specified the same version 1.0.0.0 for all examples in measurement.py and .serviceconfig file.

With 2.1.0-dev0 or greater, this version will be used to register the service. Many of the examples were using '0.1.0.0'. When TestStand tries to access a measurement, it uses '1.0.0' for the default version. So we could specify 1.0.0.0 in the .serviceconfig file and / or we could do  this.

Testing the installed examples locally.

### DIFF
--- a/examples/game_of_life/pyproject.toml
+++ b/examples/game_of_life/pyproject.toml
@@ -7,7 +7,7 @@ authors = ["National Instruments"]
 
 [tool.poetry.dependencies]
 python = "^3.8"
-ni-measurement-plugin-sdk-service = {version = "^2.1.0-dev0", allow_prereleases=true}
+ni-measurement-plugin-sdk-service = {version = "^2.1.0-dev0", allow-prereleases = true}
 click = ">=7.1.2, !=8.1.4" # mypy fails with click 8.1.4: https://github.com/pallets/click/issues/2558
 
 [tool.poetry.group.dev.dependencies]
--- a/examples/game_of_life/pyproject.toml
+++ b/examples/game_of_life/pyproject.toml
@@ -7,7 +7,7 @@ authors = ["National Instruments"]
 
 [tool.poetry.dependencies]
 python = "^3.8"
-ni-measurement-plugin-sdk-service = {version = "^2.0.0"}
+ni-measurement-plugin-sdk-service = {version = "^2.1.0-dev0", allow_prereleases=true}
 click = ">=7.1.2, !=8.1.4" # mypy fails with click 8.1.4: https://github.com/pallets/click/issues/2558
 
 [tool.poetry.group.dev.dependencies]
--- a/examples/nidaqmx_analog_input/measurement.py
+++ b/examples/nidaqmx_analog_input/measurement.py
@@ -17,7 +17,7 @@ script_or_exe = sys.executable if getattr(sys, "frozen", False) else __file__
 service_directory = pathlib.Path(script_or_exe).resolve().parent
 measurement_service = nims.MeasurementService(
     service_config_path=service_directory / "NIDAQmxAnalogInput.serviceconfig",
-    version="1.0.0.0",
+    version="1.0.0",
     ui_file_paths=[service_directory / "NIDAQmxAnalogInput.measui"],
 )
 
--- a/examples/nidaqmx_analog_input/measurement.py
+++ b/examples/nidaqmx_analog_input/measurement.py
@@ -17,6 +17,7 @@ script_or_exe = sys.executable if getattr(sys, "frozen", False) else __file__
 service_directory = pathlib.Path(script_or_exe).resolve().parent
 measurement_service = nims.MeasurementService(
     service_config_path=service_directory / "NIDAQmxAnalogInput.serviceconfig",
+    version="1.0.0.0",
     ui_file_paths=[service_directory / "NIDAQmxAnalogInput.measui"],
 )
 
--- a/examples/nidaqmx_analog_input/measurement.py
+++ b/examples/nidaqmx_analog_input/measurement.py
@@ -17,7 +17,6 @@ script_or_exe = sys.executable if getattr(sys, "frozen", False) else __file__
 service_directory = pathlib.Path(script_or_exe).resolve().parent
 measurement_service = nims.MeasurementService(
     service_config_path=service_directory / "NIDAQmxAnalogInput.serviceconfig",
-    version="0.1.0.0",
     ui_file_paths=[service_directory / "NIDAQmxAnalogInput.measui"],
 )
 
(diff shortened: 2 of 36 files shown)


examples/nidaqmx_analog_input/measurement.py
```diff
@@ -17,7 +17,7 @@
 service_directory = pathlib.Path(script_or_exe).resolve().parent
 measurement_service = nims.MeasurementService(
     service_config_path=service_directory / "NIDAQmxAnalogInput.serviceconfig",
-    version="0.1.0.0",
+    version="1.0.0.0",
```
bkeryan, line 20: You can delete this parameter entirely from the `measurement.py` files. @nick-beer made it optional in https://github.com/ni/measurement-plugin-python/pull/827 . If not specified, it will be read from the `.serviceconfig` file.

bkeryan (changes requested):

examples/nidaqmx_analog_input/measurement.py
nick-beer: Yeah - in fact, passing the a version to the constructor is problematic _if_ it is not also found in the .serviceconfig file.  We should probably update the service.py code to enforce this.  The problem is as follows:

- You specify a version in your constructor, but _not_ in your .serviceconfig file.
- When the service is published to plugin library or statically installed, the DiscoveryService thinks the service is _unversioned_ - because there's no version in the .serviceconfig file.
- As a result, it registers it internally with a default version (used to be `1.0.0`, but as of last night, it's now `0.0.1`)
- However, when the service starts itself and registers itself with discovery service, it uses this version that's passed in the constructor, and _those versions don't match_.  This means DiscoveryService fails to see this as a valid registration/resolution, and the client is left hanging, waiting for that valid registration.

bkeryan: We previously discussed emitting a warning if you specify the version when constructing MeasurementService: https://github.com/ni/measurement-plugin-python/pull/827#discussion_r1704562044

Maybe that would be appropriate now, but this seems like a regression for existing services.

dixonjoel: > @nick-beer made it optional in https://github.com/ni/measurement-plugin-python/pull/827

The version parameter isn't optional yet. It would be a breaking change to take it out, wouldn't it?
```
  File "C:\dev\measurementlink-python\examples\game_of_life\measurement.py", line 16, in <module>
    measurement_service = nims.MeasurementService(
TypeError: __init__() missing 1 required positional argument: 'version'
```

nick-beer: Pretty sure it is optional - see [here](https://github.com/ni/measurement-plugin-python/blob/82ab5530e68e40e1b86ca6ff4e20761f2c4d5788/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py#L180).

```python
class MeasurementService:
    """Class that supports registering and hosting a python function as a gRPC service."""

    def __init__(
        self,
        service_config_path: Path,
        version: str = "",
        ui_file_paths: List[Path] = [],
        service_class: Optional[str] = None,
    ) -> None:
```

bkeryan: It's optional in the 2.1 pre-release:
```python
class MeasurementService:
    """Class that supports registering and hosting a python function as a gRPC service."""

    def __init__(
        self,
        service_config_path: Path,
        version: str = "",
        ui_file_paths: List[Path] = [],
        service_class: Optional[str] = None,
    ) -> None:
```
But game_of_life is still using the released 2.0 package, not the 2.1 pre-release.

examples/game_of_life/pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = ["National Instruments"]
 
 [tool.poetry.dependencies]
 python = "^3.8"
-ni-measurement-plugin-sdk-service = {version = "^2.0.0"}
```
bkeryan, line 10:
```
examples/game_of_life/
  ~/work/measurement-plugin-python/measurement-plugin-python/examples/game_of_life ~/work/measurement-plugin-python/measurement-plugin-python
  
  The Poetry configuration is invalid:
    - data.dependencies.ni-measurement-plugin-sdk-service must be valid exactly by one definition (0 matches found)
  
```
allow-prereleases needs a dash, not an underscore.

examples/nidaqmx_analog_input/measurement.py
bkeryan: > Yeah - it is a bit of a regression. I think an appropriate fix is to allow the following and raise an error for anything else:
> 
> * No version specified in either the constructor or serviceconfig.  This will just get the default version
> * No version specified in the constructor, but a version specified in the serviceconfig.
> * A version specified in the constructor, and that same version is specified in the serviceconfig
> 
> If someone has an existing service with a version in the constructor (effectively all services), they'll be required to make some kind of change when updating SDK versions. This is unfortunate, but because DiscoveryService reads the .serviceconfig file to get version information, I'm not sure how else to get around it.

The current state of this PR should fix the examples. Updating the error logic in a follow-up PR is fine with me.

bkeryan (approved):

examples/nidaqmx_analog_input/measurement.py
bkeryan: https://github.com/ni/measurement-plugin-python/issues/908

bkeryan:
The test failure looks like flakiness in one of the new client cancellation tests:
```
================================== FAILURES ===================================
____ test___streaming_measurement_execution___cancel___cancels_measurement ____

measurement_plugin_client_module = <module 'test_measurement_client.py' from 'C:\\Users\\ec2-user\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-5\\measurement_plugin_client_files7\\test_measurement_client.py'>

    def test___streaming_measurement_execution___cancel___cancels_measurement(
        measurement_plugin_client_module: ModuleType,
    ) -> None:
        test_measurement_client_type = getattr(measurement_plugin_client_module, "TestMeasurement")
        measurement_plugin_client = test_measurement_client_type()
    
            with concurrent.futures.ThreadPoolExecutor() as executor:
                measure = executor.submit(lambda: list(measurement_plugin_client.stream_measure()))
                measurement_plugin_client.cancel()
>               measure.result()
E               Failed: DID NOT RAISE <class 'grpc.RpcError'>

C:\actions-runner\1\_work\measurement-plugin-python\measurement-plugin-python\packages\generator\tests\acceptance\test_streaming_measurement_client.py:96: Failed
```
I filed #910 and reran the PR checks.

PR #911, opened 2024-09-22
Update version checking to warn that version is deprecated and error if there is no .serviceconfig version

### LINKED ISSUE #908: MeasurementService should provide more guidance about moving version number to the .serviceconfig file
# Bug Report

If the service specifies the version number in code but not in the serviceconfig file, the discovery service assumes that the service is unversioned and assigns it a default version number. If this doesn't match, calling the measurement service from TestStand will fail.

From @nick-beer 

> Yeah - in fact, passing the a version to the constructor is problematic _if_ it is not also found in the .serviceconfig file.  We should probably update the service.py code to enforce this.  The problem is as follows:
>
>- You specify a version in your constructor, but _not_ in your .serviceconfig file.
>- When the service is published to plugin library or statically installed, the DiscoveryService thinks the service is _unversioned_ - because there's no version in the .serviceconfig file.
>- As a result, it registers it internally with a default version (used to be `1.0.0`, but as of last night, it's now `0.0.1`)
>- However, when the service starts itself and registers itself with discovery service, it uses this version that's passed in the constructor, and _those versions don't match_.  This means DiscoveryService fails to see this as a valid registration/resolution, and the client is left hanging, waiting for that valid registration.
>
> ...
> I think an appropriate fix is to allow the following and raise an error for anything else:
>- No version specified in either the constructor or serviceconfig.  This will just get the default version
>- No version specified in the constructor, but a version specified in the serviceconfig.
>- A version specified in the constructor, and that same version is specified in the serviceconfig
>
>If someone has an existing service with a version in the constructor (effectively all services), they'll be required to make some kind of change when updating SDK versions.  This is unfortunate, but because DiscoveryService reads the .serviceconfig file to get version information, I'm not sure how else to get around it.

See this PR discussion for more details:
https://github.com/ni/measurement-plugin-python/pull/907#discussion_r1769231913

[AB#2863530](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2863530)

dixonjoel:

Fixes #908 
This gives a warning if the 'version' parameter of `MeasurementService` is used. It also errors if there is a 'version' parameter, but no 'version' in the .serviceconfig file. The rule is if you pass a 'version' parameter, you must have the same 'version' in the .serviceconfig file.

Gives clearer guidance on the latest state of the version parameters.

Existing tests plus added one test to verify a RuntimeError in the case where there is a 'version' parameter passed but no version in the .serviceconfig file.

### DIFF
--- a/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
@@ -219,7 +219,10 @@ class MeasurementService:
             )
 
         if version:
-            warnings.warn("The 'version' parameter is deprecated. Specify the version in the .serviceconfig file instead.", DeprecationWarning)
+            warnings.warn(
+                "The 'version' parameter is deprecated. Specify the version in the .serviceconfig file instead.",
+                DeprecationWarning,
+            )
         config_version = service.get("version")
         if version and config_version is None:
             raise RuntimeError(
--- a/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
@@ -225,9 +225,7 @@ class MeasurementService:
             )
         config_version = service.get("version")
         if version and config_version is None:
-            raise RuntimeError(
-                f"Version error: version parameter is '{version}'. Specify the version using the 'version' field in the .serviceconfig file instead.",
-            )
+            version = None
         if config_version is not None:
             if version and version != config_version:
                 raise RuntimeError(
--- a/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
@@ -218,7 +218,13 @@ class MeasurementService:
                 f"Service class '{service_class}' not found in '{service_config_file}'"
             )
 
+        if version:
+            warnings.warn("The 'version' parameter is deprecated. Specify the version in the .serviceconfig file instead.", DeprecationWarning)
         config_version = service.get("version")
+        if version and config_version is None:
+            raise RuntimeError(
+                f"Version error: version parameter is '{version}'. Specify the version in the .serviceconfig file instead.",
+            )
         if config_version is not None:
             if version and version != config_version:
                 raise RuntimeError(
--- a/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
@@ -220,13 +220,13 @@ class MeasurementService:
 
         if version:
             warnings.warn(
-                "The 'version' parameter is deprecated. Specify the version in the .serviceconfig file instead.",
+                "The 'version' constructor parameter is deprecated. Specify the version using the 'version' field in the .serviceconfig file instead.",
                 DeprecationWarning,
             )
         config_version = service.get("version")
         if version and config_version is None:
             raise RuntimeError(
-                f"Version error: version parameter is '{version}'. Specify the version in the .serviceconfig file instead.",
+                f"Version error: version parameter is '{version}'. Specify the version using the 'version' field in the .serviceconfig file instead.",
             )
         if config_version is not None:
             if version and version != config_version:
--- a/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
@@ -225,7 +225,7 @@ class MeasurementService:
             )
         config_version = service.get("version")
         if version and config_version is None:
-            version = None
+            version = ""
         if config_version is not None:
             if version and version != config_version:
                 raise RuntimeError(
--- a/packages/service/tests/unit/test_service.py
+++ b/packages/service/tests/unit/test_service.py
@@ -450,9 +450,9 @@ def test___service_config___create_measurement_service_with_version___service_co
     test_assets_directory: pathlib.Path,
 ):
     measurement_service = MeasurementService(
-            service_config_path=test_assets_directory / "example.NoVersion.serviceconfig",
-            version="2.0.1",
-        )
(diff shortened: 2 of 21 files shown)


packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
```diff
@@ -218,7 +218,16 @@ def __init__(
                 f"Service class '{service_class}' not found in '{service_config_file}'"
             )
 
+        if version:
+            warnings.warn(
+                "The 'version' parameter is deprecated. Specify the version in the .serviceconfig file instead.",
```
nick-beer, line 223: Add a little more detail here? Something like `The 'version' constructor parameter is deprecated.  Specify the version using the 'version' field in the .serviceconfig file instead.`

Same in the message below: tell them which json field to add to the .serviceconfig.

nick-beer (commented):
LGTM - assuming someone else will grant explicit approval for these changes.

packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
bkeryan, line 1: This breaks 100% of existing services. Would it be possible to ignore the version parameter and make the service work?

```
ni-measurement-plugin-sdk-service = {path = "D:/Dev/measurement-plugin-python/packages/service"}
```

```
PS C:\Users\Administrator\Downloads\measurement-plugin-python-examples-2.0.0\sample_measurement>poetry run python .\measurement.py -v                                                             Traceback (most recent call last):
  File "C:\Users\Administrator\Downloads\measurement-plugin-python-examples-2.0.0\sample_measurement\measurement.py", line 20, in <module>
    measurement_service = nims.MeasurementService(
  File "C:\Users\Administrator\Downloads\measurement-plugin-python-examples-2.0.0\sample_measurement\.venv\lib\site-packages\ni_measurement_plugin_sdk_service\measurement\service.py", line 228, in __init__
    raise RuntimeError(
RuntimeError: Version error: version parameter is '0.1.0.0'. Specify the version using the 'version' field in the .serviceconfig file instead.
```

nick-beer: Yeah, you could just set the `version` parameter to `None` and not pass it along to the `RegisterService` call on DiscoveryService.  You'd lose the version reporting from the measurement API, but that's probably fine given that it's not used for anything.

nick-beer: Given that we have a warning for the constructor parameter `version`, this seems like a reasonable approach to take.

dixonjoel: So basically instead of throwing a RuntimeError, just set `version = None` and carry on? And then I can revert all of the __init__.py changes that remove the version?

bkeryan: > So basically instead of throwing a RuntimeError, just set version = None and carry on? 

That's a choice. Either we warn-and-ignore, or we warn-validate-and-ignore.

> And then I can revert all of the init.py changes that remove the version?

No.

```
============================== warnings summary ===============================
tests/unit/test_service.py::test___service_config___create_measurement_service_with_version___version_differs_from_service_config
tests/unit/test_service.py::test___service_config___create_measurement_service_with_version___service_config_has_no_version
tests/unit/test_service.py::test___service_config___create_measurement_service_with_version___version_is_used[example.serviceconfig-1.0.1]
tests/unit/test_service.py::test___service_config___create_measurement_service_with_version___version_is_used[example.v1.serviceconfig-1.0.1]
  D:\a\measurement-plugin-python\measurement-plugin-python\packages\service\ni_measurement_plugin_sdk_service\measurement\service.py:222: DeprecationWarning: The 'version' constructor parameter is deprecated. Specify the version using the 'version' field in the .serviceconfig file instead.
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
- generated xml file: D:\a\measurement-plugin-python\measurement-plugin-python\packages\service\test_results\nims-windows-latest-py3.11-no-extras.xml -
```

> Or do we go ahead and leave the init.py changes - make the test services we have not specify a version and also don't break any existing services that may be specifying a version in the constructor and not in the .serviceconfig.

I think we should have one test service that specifies `version` at runtime. The test for that service should use `pytest.deprecated_call`, which will validate that the warning is thrown and prevent it from showing up in the test results.

bkeryan (approved):

packages/service/tests/unit/test_service.py
```diff
@@ -446,11 +446,22 @@ def test___service_config___create_measurement_service_with_version___version_di
...
 
 
+def test___service_config___create_measurement_service_with_version___service_config_has_no_version(
+    test_assets_directory: pathlib.Path,
+):
+    measurement_service = MeasurementService(
+        service_config_path=test_assets_directory / "example.NoVersion.serviceconfig",
+        version="2.0.1",
+    )
+
```
bkeryan, line 456: These tests are generating warnings: 

```
============================== warnings summary ===============================
tests/unit/test_service.py::test___service_config___create_measurement_service_with_version___version_differs_from_service_config
tests/unit/test_service.py::test___service_config___create_measurement_service_with_version___service_config_has_no_version
tests/unit/test_service.py::test___service_config___create_measurement_service_with_version___version_is_used[example.serviceconfig-1.0.1]
tests/unit/test_service.py::test___service_config___create_measurement_service_with_version___version_is_used[example.v1.serviceconfig-1.0.1]
  D:\a\measurement-plugin-python\measurement-plugin-python\packages\service\ni_measurement_plugin_sdk_service\measurement\service.py:222: DeprecationWarning: The 'version' constructor parameter is deprecated. Specify the version using the 'version' field in the .serviceconfig file instead.
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
- generated xml file: D:\a\measurement-plugin-python\measurement-plugin-python\packages\service\test_results\nims-windows-latest-py3.9-all-extras.xml -
```

Fixed in https://github.com/ni/measurement-plugin-python/pull/942

PR #914, opened 2024-09-23
Add test case to check mypy error in generated client and fix existing mypy errors

Jotheeswaran-Nandagopal:

- Add test case for checking the mypy issues in the generated Measurement Plug-In Client.
- Fix the existing type hints and mypy issues in `Mako` template.

- Fix the existing type hints in the `mako` template.
- Add mypy check as a test case to check the type hints in the generated Measurement Plug-In Client.

- Done manual testing.

### DIFF
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -121,7 +121,7 @@ class ${class_name}:
                 type=Field.Kind.ValueType(${value.type}),
                 repeated=${value.repeated},
                 default_value=${_format_default_value(value.default_value)},
-                annotations=${value.annotations | n,repr},
+                annotations=${value.annotations | n},
                 message_type=${value.message_type | repr},
                 field_name=${value.field_name | repr},
                 % if value.enum_type:
@@ -139,7 +139,7 @@ class ${class_name}:
                 type=Field.Kind.ValueType(${value.type}),
                 repeated=${value.repeated},
                 default_value=${value.default_value},
-                annotations=${value.annotations | n,repr},
+                annotations=${value.annotations | n},
                 message_type=${value.message_type | repr},
                 field_name=${value.field_name | repr},
                 % if value.enum_type:
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -21,6 +21,8 @@ from typing import Any
 
 """Generated client API for the ${display_name | repr} measurement plug-in."""
 
+from __future__ import annotations
+
 import logging
 import threading
 % if len(enum_by_class_name):
@@ -120,13 +122,13 @@ class ${class_name}:
         self._configuration_metadata = {
             % for key, value in configuration_metadata.items():
             ${key}: ParameterMetadata(
-                display_name="${value.display_name}",
+                display_name=${value.display_name | repr},
                 type=Field.Kind.ValueType(${value.type}),
                 repeated=${value.repeated},
                 default_value=${_format_default_value(value.default_value)},
-                annotations=${value.annotations},
-                message_type="${value.message_type}",
-                field_name="${value.field_name}",
+                annotations=${value.annotations | n,repr},
+                message_type=${value.message_type | repr},
+                field_name=${value.field_name | repr},
                 enum_type=${_replace_enum_class_type(str(value.enum_type))}
             ),  
             % endfor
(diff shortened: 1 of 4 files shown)


dixonjoel (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -111,17 +115,43 @@ class ${class_name}:
...
         self._measure_response: Optional[
-            Generator[v2_measurement_service_pb2.MeasureResponse, None, None]
+            grpc.CallIterator[v2_measurement_service_pb2.MeasureResponse]
         ] = None
-        self._configuration_metadata = ${configuration_metadata}
-        self._output_metadata = ${output_metadata}
+        self._configuration_metadata = {
+            % for key, value in configuration_metadata.items():
+            ${key}: ParameterMetadata(
+                display_name="${value.display_name}",
```
bkeryan, line 123: Use repr to ensure proper quoting and avoid code injection.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -111,17 +115,43 @@ class ${class_name}:
         self._pin_map_client = pin_map_client
         self._stub: Optional[v2_measurement_service_pb2_grpc.MeasurementServiceStub] = None
         self._measure_response: Optional[
-            Generator[v2_measurement_service_pb2.MeasureResponse, None, None]
+            grpc.CallIterator[v2_measurement_service_pb2.MeasureResponse]
```
bkeryan, line 118: This type is defined by https://github.com/shabbyrobe/grpc-stubs/ , not by `grpcio`:

```
>>> import grpc
>>> grpc.CallIterator
AttributeError: module 'grpc' has no attribute 'CallIterator'
```

Somehow, this particular annotation is not causing a run time error, but the same annotation in a function signature would cause a run time error:

```
>>> def f() -> grpc.CallIterator: pass
...
AttributeError: module 'grpc' has no attribute 'CallIterator'
```

Please add `from __future__ import annotations` to the generated code, just to be safe.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -111,17 +115,43 @@ class ${class_name}:
...
+        self._configuration_metadata = {
+            % for key, value in configuration_metadata.items():
+            ${key}: ParameterMetadata(
+                display_name="${value.display_name}",
+                type=Field.Kind.ValueType(${value.type}),
+                repeated=${value.repeated},
+                default_value=${_format_default_value(value.default_value)},
+                annotations=${value.annotations},
+                message_type="${value.message_type}",
+                field_name="${value.field_name}",
```
bkeryan, line 129: Use repr to ensure proper quoting and avoid code injection.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -111,17 +115,43 @@ class ${class_name}:
...
+                enum_type=${_replace_enum_class_type(str(value.enum_type))}
```
bkeryan, line 130: This is https://github.com/ni/measurement-plugin-python/pull/901#discussion_r1766965181 again.

Now you have the actual type object, so remove the regex hack.

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -111,17 +115,43 @@ class ${class_name}:
...
-        self._configuration_metadata = ${configuration_metadata}
-        self._output_metadata = ${output_metadata}
+        self._configuration_metadata = {
+            % for key, value in configuration_metadata.items():
+            ${key}: ParameterMetadata(
+                display_name="${value.display_name}",
+                type=Field.Kind.ValueType(${value.type}),
+                repeated=${value.repeated},
+                default_value=${_format_default_value(value.default_value)},
+                annotations=${value.annotations},
```
bkeryan, line 127: This uses the default filter of `str`. Try using `${value.annotations | repr}` (or `${value.annotations | n,repr}`) to be more explicit about it.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
Jotheeswaran-Nandagopal: I have added `from __future__ import annotations` import to the generated code.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
Jotheeswaran-Nandagopal: Sure, I removed the helper method with a regex approach and updated the code to use `value.enum_type.__name__` to extract the enum type from the type object (for example, `<enum 'EnumInEnum'>`).

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
Jotheeswaran-Nandagopal: Sure, I have updated with `${value.annotations | n,repr}`.

bkeryan (approved):

packages/generator/tests/test_assets/example_renders/measurement_plugin_client/non_streaming_data_measurement_client.py
bkeryan, line 1: These type hints are still wrong. They don't match the runtime type. None of the list outputs are actually lists.

bkeryan: #930

PR #915, opened 2024-09-23
[High Priority] Update default value for the pin map sites

Jotheeswaran-Nandagopal:

- Updated the Measurement Plug-In Client template to set a default value for the `PinMapContext.sites`.

- When the measurement client is run for a pin-based measurement using IOResource name, sites value of the pin map context input should be explicitly set to 0.  If there are no sites in the pin map context, some measurement would fail with an exception.
**Context**: https://dev.azure.com/ni/DevCentral/_git/983dfe62-3b16-436f-b2ef-d98a72840b64/pullrequest/698395

- Done manual testing.

### DIFF
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
@@ -118,7 +118,7 @@ class ${class_name}:
         if grpc_channel is not None:
             self._stub = v2_measurement_service_pb2_grpc.MeasurementServiceStub(grpc_channel)
         self._create_file_descriptor()
-        self._pin_map_context: Optional[PinMapContext] = None
+        self._pin_map_context: PinMapContext = PinMapContext(pin_map_id="", sites=[0])
 
     @property
     def pin_map_context(self) -> PinMapContext:
@@ -136,8 +136,7 @@ class ${class_name}:
     @property
     def sites(self) -> List[int]:
         """The sites where the measurement must be executed."""
-        if self._pin_map_context is not None:
-            return self._pin_map_context.sites
+        return self._pin_map_context.sites
 
     @sites.setter
     def sites(self, val: List[int]) -> None:
@@ -206,7 +205,7 @@ class ${class_name}:
         )
         return v2_measurement_service_pb2.MeasureRequest(
             configuration_parameters=serialized_configuration,
-            pin_map_context=self._pin_map_context._to_grpc() if self._pin_map_context else None,
+            pin_map_context=self._pin_map_context._to_grpc(),
         )
 
     % if output_metadata:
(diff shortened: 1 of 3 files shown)


packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -118,7 +118,7 @@ class ${class_name}:
         if grpc_channel is not None:
             self._stub = v2_measurement_service_pb2_grpc.MeasurementServiceStub(grpc_channel)
         self._create_file_descriptor()
-        self._pin_map_context: Optional[PinMapContext] = None
+        self._pin_map_context: PinMapContext = PinMapContext(pin_map_id="", sites=[0])
```
jasonmreding, line 121: This seems like it is working around a bug in our python libraries. Has the underlying bug been fixed and we are just trying to provide compatibility with older versions of the library that have been deployed? We should at least fix the bug so clients don't have to keep implementing this hack if they care about backward compatibility with already deployed services.

dixonjoel:
Can you clarify what the problem was a bit more?
> If there are no sites in the pin map context, some measurement would fail with an exception.

What is the scenario here? The steps to hit this and the result?

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
Avinash2Suresh: > Has the underlying bug been fixed and we are just trying to provide compatibility with older versions of the library that have been deployed?

I see that we need to update the GetConnections implementation in both Python and LabVIEW to handle empty sites request properly. I guess it will take some considerable time to update the GetConnections workflow. Until then, I would prefer to live with this workaround and raise a work item to track the necessary changes in GetConnections for a later time.

bkeryan: I think the fix is probably to specify `reserved_sites=context.sites if context.sites else None` in a couple places here: https://github.com/ni/measurement-plugin-python/blob/main/packages/service/ni_measurement_plugin_sdk_service/session_management/_client.py#L148

bkeryan: also, file a bug, regardless of whether you are fixing it or working around it

Avinash2Suresh: Sure, we will raise a bug to track this.

We need to address the same site issue in both the Python and LabVIEW libraries so that the Python client functions well with both the Python and LabVIEW measurement services. Until the issue is resolved in both services, we can proceed with the workaround proposed in this PR. Please help review the PR if you’re okay with moving forward with the client workaround.

bkeryan (approved):

PR #924, opened 2024-09-23
examples: Don't assume array parameters are implemented as lists

bkeryan:

`niscope_acquire_waveform`: 
- Explicitly convert `measurement_pins` to a `list` before concatenating it to another `list`. 
- Update the type hint to `Sequence[str]`. `Iterable[str]` is also okay.

service/tests:
- Change configuration parameter type hints from `List[T]` to `Iterable[T]`. `Sequence[T]` is also okay.
- Fix incorrect type hints and weird conversions in `nidcpower_measurement`
- Fix incorrect type hints in `nidigital_measurement`
- Fix some test services that didn't have type hints and were returning a float instead of a tuple

After #767, array parameters are represented as a `protobuf`-specific `Sequence[T]` such as `RepeatedScalarFieldContainer[T]`.

Ran updated examples.

### DIFF
--- a/examples/niscope_acquire_waveform/measurement.py
+++ b/examples/niscope_acquire_waveform/measurement.py
@@ -5,7 +5,7 @@ import pathlib
 import sys
 import threading
 import time
-from typing import List, Tuple
+from typing import List, Sequence, Tuple
 
 import click
 import grpc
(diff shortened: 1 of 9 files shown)


dixonjoel (approved):

examples/niscope_acquire_waveform/measurement.py
bkeryan, line 1: We should update the tests too.

bkeryan: done

PR #925, opened 2024-09-24
Update Measurement Plug-In Generated Client to Resolve with the Corresponding Measurement Service Version

Jotheeswaran-Nandagopal:

- Add `version` as a parameter in the `resolve_service` API of the discovery client with respect to the latest change in [ni-apis](https://github.com/ni/ni-apis/blob/main/ni/measurementlink/discovery/v1/discovery_service.proto#L173).
- Update the unit tests for the `resolve_service` API with version parameter.
- Update the `Mako` template of Measurement Plug-In Client generator to resolve the service with the version of the measurement.

- The `ResolveServiceRequest` has been updated in the [ni-apis](https://github.com/ni/ni-apis/blob/main/ni/measurementlink/discovery/v1/discovery_service.proto#L173), with `version`.

- As, multiple versions of same measurement can be registered simultaneously, the resolve service will fetch the service with latest version, if the version is not specified while resolving. So, when a user creates a client with version `1.0.0.0` and later, run the `measure()` API in client with higher version `2.0.0.0` of registered measurement, the service location will be resolved without any error, as the version is not checked. With this, if the configurations are updated in new version of measurement, this will result in error during measure call.
- With the implementation in this Pull Request (adding version), client will be generated for the latest version (as multiple versions are not supported at present). During the measurement process in the generated client, measurement service will be resolved with the version of the generated client. If the version mismatch is captured, throw error appropriately.

- Done manual testing
- Automated test cases passed


jayaseelan-james (approved):
Following https://github.com/ni/measurement-plugin-python/pull/907, please consider moving the version of the test measurements from \_\_init__.py to its corresponding `.serviceconfig`.

packages/service/ni_measurement_plugin_sdk_service/discovery/_client.py
```diff
@@ -206,7 +206,9 @@ def unregister_service(self, registration_id: str = "") -> bool:
             _logger.exception("Error in unregistering with discovery service.")
             raise
 
-    def resolve_service(self, provided_interface: str, service_class: str = "") -> ServiceLocation:
+    def resolve_service(
+        self, provided_interface: str, service_class: str = "", version: str = ""
```
bkeryan, line 210: Skipping deployment_target means that if we add it later, the parameters are out of order compared to the .proto file.

https://github.com/ni/ni-apis/blob/main/ni/measurementlink/discovery/v1/discovery_service.proto#L173

dixonjoel (approved):

packages/service/tests/unit/test_discovery_client.py
```diff
@@ -128,13 +128,16 @@ def test___service_registered___resolve_service___sends_request(
     )
 
     service_location = discovery_client.resolve_service(
-        _TEST_SERVICE_INFO.provided_interfaces[0], _TEST_SERVICE_INFO.service_class
+        provided_interface=_TEST_SERVICE_INFO.provided_interfaces[0],
+        service_class=_TEST_SERVICE_INFO.service_class,
+        version=_TEST_SERVICE_INFO.versions[0],
```
bkeryan, line 133: You are adding a new optional parameter, so you should keep the existing test cases to ensure you didn't break existing usage. Add a new test case that exercises the new parameter.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -88,6 +88,7 @@ def _create_client(
         directory_out=directory_out,
         class_name=class_name,
         display_name=metadata.measurement_details.display_name,
+        version=metadata.measurement_details.version,
```
bkeryan, line 91: Questions:
- Do we really need to add this now?
- Is this really the behavior we want? Do saved `.sfp` and TestStand `.seq` files hardcode the version in this way?
- What should `--all` do with versioned services?

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
bkeryan, line 1: FYI: `get_measurement_service_stub` doesn't pass a version to `resolve_service`, so it always resolves the latest version of the service. I think that's fine for now and filed https://github.com/ni/measurement-plugin-python/issues/928 about potentially addressing it later.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
Jotheeswaran-Nandagopal: > On second thought, I think `--all` will generate a client for the latest version of each service, which seems acceptable.

Yes, `-all` will generate latest version of each service.

Avinash2Suresh: I discussed this with @cfuentes-ni a few hours ago, as we had not considered version handling earlier. If we want to allow the user to choose or provide the version when generating the client, we would need a few design changes to accommodate that. For example, we would need to get the version from the user in batch mode, display the measurement with the version in interactive mode, etc.

The workflow we have aligned on now is to create a client for the latest available version for the given service class and store the version in the generated client. This stored version will then be used for further interactions with the corresponding measurement service of the same version. The current LabVIEW implementation follows the same flow.

To support client generation with an explicit version provided by the user, we have decided to move it out of the scope of this feature and track it as a separate work item.

bkeryan (approved):

nick-beer (approved):

bkeryan:
@Jotheeswaran-Nandagopal @Avinash2Suresh I think part of what's going on with the build failures is that our virtualenv caching uses the poetry.lock as a key, and this doesn't change when you modify files in `../../packages/service`.

https://github.com/ni/measurement-plugin-python/blob/main/.github/workflows/run_system_tests.yml

packages/service/pyproject.toml
```diff
@@ -7,7 +7,7 @@ extend_exclude = '.tox/,*_pb2_grpc.py,*_pb2_grpc.pyi,*_pb2.py,*_pb2.pyi,ni_measu
 
 [tool.poetry]
 name = "ni_measurement_plugin_sdk_service"
-version = "2.1.0-dev0"
+version = "2.1.0-dev1"
```
jayaseelan-james, line 10: The "2.1.0-dev1" is already pre-released. This should be `dev2` now.
```suggestion
version = "2.1.0-dev2"
```

jayaseelan-james (changes requested):
If you're updating the version of the service pkg, then please update the version of the generator and SDK packages as well.

packages/service/tests/unit/test_discovery_client.py
```diff
@@ -147,7 +171,9 @@ def test___service_not_registered___resolve_service___raises_not_found_error(
 
     with pytest.raises(grpc.RpcError) as exc_info:
         _ = discovery_client.resolve_service(
-            _TEST_SERVICE_INFO.provided_interfaces[0], _TEST_SERVICE_INFO.service_class
+            provided_interface=_TEST_SERVICE_INFO.provided_interfaces[0],
+            service_class=_TEST_SERVICE_INFO.service_class,
+            version=_TEST_SERVICE_INFO.versions[0],
```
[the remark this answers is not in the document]
Jotheeswaran-Nandagopal: Reverted the change.

Jotheeswaran-Nandagopal:
As per the updates in #911, I have moved the `version` from the constructor to `.serviceconfig` file, as the pipelines are failing due to the requirement to include the version in the generated client.

jayaseelan-james (approved):

Jotheeswaran-Nandagopal:
> @Jotheeswaran-Nandagopal @Avinash2Suresh I think part of what's going on with the build failures is that our virtualenv caching uses the poetry.lock as a key, and this doesn't change when you modify files in `../../packages/service`.
> 
> https://github.com/ni/measurement-plugin-python/blob/main/.github/workflows/run_system_tests.yml

@bkeryan I have updated the package versions as service package is updated, in order to resolve the pipeline failure caused by our virtualenv caching.

bkeryan:
> > @Jotheeswaran-Nandagopal @Avinash2Suresh I think part of what's going on with the build failures is that our virtualenv caching uses the poetry.lock as a key, and this doesn't change when you modify files in `../../packages/service`.
> > https://github.com/ni/measurement-plugin-python/blob/main/.github/workflows/run_system_tests.yml
> 
> @bkeryan I have updated the package versions as service package is updated, in order to resolve the pipeline failure caused by our virtualenv caching.

https://github.com/ni/measurement-plugin-python/issues/941

PR #931, opened 2024-09-24
[High Priority] Fix Path type conversion and type hints for client codegen

### LINKED ISSUE #930: Client generator type hints say array outputs are `List[T]` but the runtime type is `google._upb._message.RepeatedScalarContainer`
# Bug Report

The client generator directly returns protobuf fields, which are not lists, but the type hints say that the repeated fields are lists. This is wrong and some list operations such as concatenation with another list will fail.

## Repro or Code Sample

```python
def test___measurement_plugin_client___measure___correct_output_types(
    measurement_plugin_client_module: ModuleType,
) -> None:
    test_measurement_client_type = getattr(measurement_plugin_client_module, "TestMeasurement")
    output_type = getattr(measurement_plugin_client_module, "Outputs")
    measurement_plugin_client = test_measurement_client_type()

    response = measurement_plugin_client.measure()

    assert isinstance(response, output_type)
    assert isinstance(response.float_out, float)
    print(type(response.double_array_out))
    assert isinstance(response.double_array_out, list)
    assert all([isinstance(x, float) for x in response.double_array_out])
    assert isinstance(response.bool_out, bool)
    assert isinstance(response.string_out, str)
    assert isinstance(response.string_array_out, list)
    assert all([isinstance(x, str) for x in response.string_array_out])
    assert isinstance(response.path_out, pathlib.Path)
    assert all([isinstance(x, pathlib.Path) for x in response.path_array_out])
    assert isinstance(response.io_out, str)
    assert isinstance(response.io_array_out, list)
    assert all([isinstance(x, str) for x in response.io_array_out])
    assert isinstance(response.integer_out, int)
    assert isinstance(response.xy_data_out, None)
    assert isinstance(response.io_out, str)
    assert isinstance(response.io_array_out, list)
    assert all([isinstance(x, str) for x in response.io_array_out])
    assert isinstance(response.enum_out, EnumInEnum)
    assert isinstance(response.enum_array_out, list)
    assert all([isinstance(x, EnumInEnum) for x in response.enum_array_out])
    assert isinstance(response.protobuf_enum_out, ProtobufEnumInEnum)
```

## Expected Behavior

It passes.

## Current Behavior

```
tests/acceptance/test_non_streaming_measurement_client.py::test___measurement_plugin_client___measure___correct_output_types <class 'google._upb._message.RepeatedScalarContainer'>
FAILED

====================================================== FAILURES =======================================================
__________________________ test___measurement_plugin_client___measure___correct_output_types __________________________

measurement_plugin_client_module = <module 'test_measurement_client.py' from 'C:\\Users\\Administrator\\AppData\\Local\\Temp\\pytest-of-Administrator\\pytest-2\\measurement_plugin_client_files0\\test_measurement_client.py'>

    def test___measurement_plugin_client___measure___correct_output_types(
        measurement_plugin_client_module: ModuleType,
    ) -> None:
        test_measurement_client_type = getattr(measurement_plugin_client_module, "TestMeasurement")
        output_type = getattr(measurement_plugin_client_module, "Outputs")
        measurement_plugin_client = test_measurement_client_type()

        response = measurement_plugin_client.measure()

        assert isinstance(response, output_type)
        assert isinstance(response.float_out, float)
        print(type(response.double_array_out))
>       assert isinstance(response.double_array_out, list)
E       assert False
E        +  where False = isinstance([0.1, 0.2, 0.3], list)
E        +    where [0.1, 0.2, 0.3] = Outputs(float_out=0.05999999865889549, double_array_out=[0.1, 0.2, 0.3], bool_out=False, string_out='sample string', string_array_out=['string with /forwardslash', 'string with \\backslash', "string with 'single quotes'", 'string with "double quotes"', 'string with \ttabspace', 'string with \nnewline'], path_out='sample\\path\\for\\test', path_array_out=['path\\with\\forward\\slash', 'path\\with\\backslash', "path with 'single quotes'", 'path with "double quotes"', 'path\twith\ttabs', 'path\nwith\nnewlines'], io_out='resource', io_array_out=['resource1', 'resource2'], integer_out=10, xy_data_out=None, enum_out=<EnumInEnum.BLUE: 3>, enum_array_out=[<EnumInEnum.RED: 1>, <EnumInEnum.GREEN: 2>], protobuf_enum_out=<ProtobufEnumInEnum.BLACK: 3>).double_array_out

tests\acceptance\test_non_streaming_measurement_client.py:89: AssertionError
```

## Possible Solution

Use correct type hints, like Sequence[T]

[AB#2865989](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2865989)

bkeryan:

Attempt to make the client codegen's static type hints and runtime types agree. Because so much of the serialization code uses values typed as `Any`, mypy doesn't catch blatant errors such as assigning a `str` to a field that is annotated as `Path`, or assigning a protobuf `RepeatedScalarContainer[int]` object to a field that is annotated as `List[int]`.

- Move `_convert_paths_to_strings` out of the codegen template and into the `client_support` version of `serialize_parameters`.
- Add string-to-path version to the `client_support` version of `deserialize_parameters`.
- Add `dict[int, Any]` -> `Sequence[Any]` conversion to the `client_support` version of `deserialize_parameters` and retrofit callers.
- Add test cases verifying the runtime type of each output parameter.
- Change parameter type hints to use `Iterable[T]` for array inputs and `Sequence[T]` for array outputs.

Fixes #930 

Ran acceptance tests.


packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
```diff
@@ -48,6 +48,7 @@ def serialize_parameters(
             elif parameter_metadata.type == FieldDescriptorProto.TYPE_MESSAGE:
                 getattr(message_instance, field_name).CopyFrom(parameter)
             else:
+                print(field_name, repr(parameter))
```
jayaseelan-james, line 51: Is there a particular reason to print the field name and the parameter, here? Or was it included for debugging?

packages/service/ni_measurement_plugin_sdk_service/measurement/client_support.py
```diff
@@ -1,15 +1,96 @@
...
+
+from ni_measurement_plugin_sdk_service._internal.parameter.decoder import (
+    deserialize_parameters,
+)
+from ni_measurement_plugin_sdk_service._internal.parameter.encoder import (
+    serialize_parameters,
+)
+from ni_measurement_plugin_sdk_service._internal.parameter.metadata import (
+    ParameterMetadata,
+)
```
jayaseelan-james, line 14: [Optional] The character count of all these lines is less than 100 characters. So, should we make this change?

packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py
bkeryan: Thanks for catching that. That was debugging code.

packages/service/ni_measurement_plugin_sdk_service/measurement/client_support.py
bkeryan: ni-python-styleguide lint complained about this file, and ni-python-styleguide fix did this.

mshafer-NI: this is `black`s style when the line approaches 100 (to make room for growth with a smaller diff)

packages/service/ni_measurement_plugin_sdk_service/measurement/client_support.py
```diff
@@ -1,15 +1,96 @@
...
+
+    return result
+
+
+def _path_to_string(value: Any) -> Any:
+    if isinstance(value, Path):
+        return str(value)
+    else:
+        return value
+
```
mshafer-NI, line 90: hmmm, so this method has type names in it..., 

it says it accepts "Any", but then this block will not change the value if an `int` is passed, ...

I would tend towards accepting the small copy cost of making this just `return str(value)` (without a check).

But then this is an exremely simple method that raises the question of if this method is needed.

packages/service/ni_measurement_plugin_sdk_service/measurement/client_support.py
```diff
@@ -1,15 +1,96 @@
...
+                result.append([_path_to_string(value) for value in parameter_value])
+            else:
+                result.append(_path_to_string(parameter_value))
+        else:
+            result.append(parameter_value)
+
+    return result
+
+
+def convert_strings_to_paths(
```
mshafer-NI, line 57: This method seems overly complicated.

On initial read, I come away with a few questions:
- `parameter_metadata_dict` is 1-indexed??
- so..., parameter values must be in the order of the id's in the dictionary?, why not have it also be a dict?

bkeryan: > parameter_metadata_dict is 1-indexed??

Yes. The keys are protobuf field ids, which are 1-indexed: https://protobuf.dev/programming-guides/proto3/

> so..., parameter values must be in the order of the id's in the dictionary?, why not have it also be a dict?

Good point. That's due to how the underlying serialization code handles parameter values, which is not 100% consistent between configurations (inputs) and outputs.

For configurations (inputs), we do this:
```python
    def _create_measure_request(
        self, parameter_values: List[Any]
    ) -> v2_measurement_service_pb2.MeasureRequest:
        serialized_configuration = any_pb2.Any(
            type_url=${configuration_parameters_type_url | repr},
            value=serialize_parameters(
                parameter_metadata_dict=self._configuration_metadata,
                parameter_values=parameter_values,
                service_name=f"{self._service_class}.Configurations",
            )
        )
        return v2_measurement_service_pb2.MeasureRequest(
            configuration_parameters=serialized_configuration,
            pin_map_context=self._pin_map_context._to_grpc(),
        )
```

This passes the parameters to https://github.com/ni/measurement-plugin-python/blob/main/packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py#L17 which expects a `Sequence[Any]`. Originally, this was used on the server to take a `list` or `tuple` returned from the service's `measure()` function and convert it to a gRPC message.

For outputs, we do this (prior to this PR):
```python
    def _deserialize_response(
        self, response: v2_measurement_service_pb2.MeasureResponse
    ) -> Outputs:
        if self._output_metadata:
            result = [None] * max(self._output_metadata.keys())
        else:
            result = []
        output_values = deserialize_parameters(
            self._output_metadata, response.outputs.value, f"{self._service_class}.Outputs"
        )

        for k, v in output_values.items():
            result[k - 1] = v
        return Outputs._make(result)
```

`output_values` is a `dict` indexed by protobuf field id, but we flatten it into a `list` and use the named tuple's `_make` function (which takes a sequence or iterable of named tuple fields).

I could change `convert_strings_to_paths` to operate on a `dict[int, Any]`, but then it would be inconsistent with `convert_paths_to_strings`.

https://github.com/ni/measurement-plugin-python/blob/main/packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/encoder.py#L38 highlights a way to slightly simplify these functions: use `enumerate(parameter_values, start=1)` to let `enumerate` handle the 1-based indexing.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -208,7 +208,7 @@ def get_configuration_parameters_with_type_and_default_values(
             if metadata.repeated:
                 formatted_value = ", ".join(f"Path({repr(value)})" for value in default_value)
                 default_value = f"[{formatted_value}]"
-                parameter_type = f"List[{parameter_type}]"
+                parameter_type = f"Iterable[{parameter_type}]"
```
bkeryan, line 211: Similar to Iterable for input vs. Sequence for output, we really ought to accept PurePath for input and return Path for output.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -27,7 +27,7 @@ from pathlib import Path
 <%
     typing_imports = ["Any", "Generator", "List", "Optional"]
     if output_metadata:
-        typing_imports += ["NamedTuple"]
+        typing_imports += ["NamedTuple", "Sequence"]
     if "from pathlib import Path" in built_in_import_modules:
         typing_imports += ["Iterable"]
```
bkeryan, line 32: This is wrong. Using paths does not necessary require Iterable. Using array input parameters requires Iterable.

bkeryan: @Jotheeswaran-Nandagopal @Avinash2Suresh This is too complicated. I'm going to replace it with `import typing` and `import pathlib`.

packages/service/ni_measurement_plugin_sdk_service/measurement/client_support.py
bkeryan: I deleted these helper functions and changed the conversions to be unconditional.

packages/service/ni_measurement_plugin_sdk_service/measurement/client_support.py
bkeryan: I changed the 1-based indexing to use `enumerate(..., start=1)` and renamed the `index` variable to `id`.

I filed https://github.com/ni/measurement-plugin-python/issues/939 about the inconsistency between `Sequence[Any]` and `Dict[int, Any]`.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
bkeryan: Done.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
bkeryan: I changed the `measure()` and `stream_measure()` methods to accept `pathlib.PurePath` and return `pathlib.Path`. These methods don't actually access the filesystem, they just send the paths across the network.

I didn't change `register_pin_map` because it does actually access the filesystem. (It reads the pin map contents and then sends the contents across the network.)

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -13,26 +13,27 @@
...
 )
 
-
 _V2_MEASUREMENT_SERVICE_INTERFACE = "ni.measurementlink.measurement.v2.MeasurementService"
 
 _INVALID_CHARS = "`~!@#$%^&*()-+={}[]\\|:;',<>.?/ \n"
 
 _XY_DATA_IMPORT = "from ni_measurement_plugin_sdk_service._internal.stubs.ni.protobuf.types.xydata_pb2 import DoubleXYData"
-_PATH_IMPORT = "from pathlib import Path"
+_PATH_IMPORT = "import pathlib"
```
jayaseelan-james, line 36: We hard-coded the import pathlib in #871 but continued iterating the `built_in_import_modules` dict until it got removed in #900. Therefore, this _PATH_IMPORT is not used/required anymore. Now, the `built_in_import_modules` is only used to check if it contains the "import pathlib" string literal.

Also, the `built_in_import_modules` currently only holds the "import pathlib" string literal. So, possibly in a separate PR can we replace the `built_in_import_modules` with a boolean which could be used to check if the measurement has a "Path" type configuration or output?

packages/service/ni_measurement_plugin_sdk_service/measurement/client_support.py
```diff
@@ -1,15 +1,82 @@
...
+) -> List[Any]:
+    """Convert path parameters from strings to Path objects.
+
+    Args:
+        parameter_metadata_dict: Parameter metadata by ID.
+
+        parameter_values: The parameter values.
+
+    Returns:
+        The parameter values with strings converted to Path objects, where approriate.
```
jayaseelan-james, line 68: typo.
```suggestion
        The parameter values with strings converted to Path objects, where appropriate.
```

jayaseelan-james (approved):

packages/service/ni_measurement_plugin_sdk_service/measurement/client_support.py
bkeryan: On second thought, the codegen template shouldn't be aware of this conversion. The client_support functions should handle it automatically. I'll change client_support to export functions that do what we need, instead of exporting internal functions that do 95% of what we need.

bkeryan: `client_support` now defines its own versions of `deserialize_parameters` and `serialize_parameters` that do path/str conversions and `dict[int, Any]`/`Sequence[Any]` conversions. This hides as much of the complexity as possible from the codegen template.

packages/service/ni_measurement_plugin_sdk_service/measurement/client_support.py
bkeryan: This function has been deleted.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
bkeryan: I think I removed the `built_in_import_modules` checks but I will leave further cleanup for a separate PR, because it's not critical to do that before release.

packages/service/ni_measurement_plugin_sdk_service/measurement/client_support.py
```diff
@@ -13,3 +24,89 @@
...
+        parameter_metadata_dict, parameter_bytes, message_name
+    )
+
+    for id in parameter_values.keys():
+        metadata = parameter_metadata_dict[id]
+        if (
+            convert_paths
+            and metadata.type == FieldDescriptorProto.TYPE_STRING
+            and metadata.annotations
+            and metadata.annotations.get("ni/type_specialization") == "path"
```
jayaseelan-james, line 60: Likewise, here as well, can we use the TYPE_SPECIALIZATION_KEY constant from [_annotations.py](https://github.com/ni/measurement-plugin-python/blob/801220f17686e1e1b9acf27444078b2e15f69361/packages/service/ni_measurement_plugin_sdk_service/_annotations.py#L4) here?

jayaseelan-james (approved):

packages/service/ni_measurement_plugin_sdk_service/measurement/client_support.py
```diff
@@ -13,3 +24,89 @@
...
+    """
+    new_parameter_values = list(parameter_values)
+
+    for id in parameter_metadata_dict.keys():
+        index = id - 1
+        metadata = parameter_metadata_dict[id]
+        if (
+            metadata.type == FieldDescriptorProto.TYPE_STRING
+            and metadata.annotations
+            and metadata.annotations.get("ni/type_specialization") == "path"
```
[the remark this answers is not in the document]
bkeryan: Thanks, I forgot there were constants for these.

packages/service/ni_measurement_plugin_sdk_service/measurement/client_support.py
```diff
@@ -13,3 +24,89 @@
...
+            else:
+                parameter_values[id] = Path(parameter_values[id])
+
+    if parameter_metadata_dict:
+        result = [None] * max(parameter_metadata_dict.keys())
+    else:
+        result = []
+
+    for k, v in parameter_values.items():
+        result[k - 1] = v
```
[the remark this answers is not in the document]
bkeryan: If `parameter_metadata_dict` and `parameter_values` are consistent with each other, it shouldn't matter.

If `parameter_metadata_dict` is empty and `parameter_values` is not, moving it into the `if` block will hide the error. I'd rather let it fail.

PR #944, opened 2024-09-25
generator: Add dependency on service

bkeryan:

Update `ni-measurement-plugin-sdk-generator` to depend on `ni-measurement-plugin-sdk-service`

Update the generator's Poetry dev dependencies to install service in develop mode, in order to avoid saving the service source in the cached venv.

It was relying on Poetry dev dependencies. No one tried installing the generator in isolation.

```
cd packages/generator
pipx install .
mkdir ../examples/client
cd ../examples/client
ni-measurement-plugin-client-generator -a
```


jayaseelan-james (approved):
Should we consider adding the "develop=true" attribute to the NIMS package specified under the dev.dependencies?

bkeryan:
> Should we consider adding the "develop=true" attribute to the NIMS package specified under the dev.dependencies?

Good point. I think that will help. Poetry will install a `.pth` link file into lib/site-packages instead of installing the whole package.

We were previously doing it this way, but we stopped because it didn't catch erroneous code that accessed files that only exist in Git and not in the distribution package. I guess we'll have to be more careful about that in the future, or reorganize the repo (again) to use the "src" package layout instead of the "flat" package layout.

Issue #946, opened 2024-09-26
Measurement client Generator does not support backward compatability for Older libraries


Avinash2Suresh:
# Bug Report

The Measurement Plugin client lacks backward compatibility for measurements that use older libraries.
Python measurements with ni-measurement-plugin-sdk-service version <= 2.0.0
LabVIEW measurements with ni-measurement-plugin-sdk-service version <= 3.0.0.3

The same issue is applicable for LabVIEW client generator as well.

## Repro or Code Sample
- Modify the py project.toml to point ni-measurement-plugin-sdk-service version = 2.0.0
- Generate .venv
- Run the measurement
- Generate client using ni-measurement-plugin-sdk-generator
- Run measure call

## Expected Behavior
Should give the result as expected

## Current Behavior
Failed to resolve service with provided interface & service. Not matches found error.

## Possible Solution
Store the version as 0.0.1 for older libraries irrespective of the version available in those measurements.

## Context
- The discovery service does not store the versions for measurements that use older libraries, as the version parameter did not exist in those libraries during the RegisterService process. As a result, the version is set to 0.0.1 by default for those measurements, regardless of the version specified in the .serviceconfig file or the Measurement service constructor.
- However, in our client generator, when creating the client, the version is fetched from the measurement using GetMetadata, and this version is stored in the generated client. Therefore, when we attempt to communicate with the measurement using the actual version, the discovery service throws an error because it is unaware of the actual version and instead holds the version as 0.0.1.

[AB#2869356](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2869356)

Avinash2Suresh:
FYI, @bkeryan, @jasonmreding

bkeryan:
Another potential solution is to ask the discovery service what the service's version is (by calling ResolveServiceWithInformation) instead of asking the service itself.

Cc: @nick-beer

jasonmreding:
I'm not sure I understand the issue at hand. Are you saying the `GetMetadata` API from the measurement service is returning a different version than what the discovery service returns via `ResolveServiceWithInformation` or `EnumerateServices`? I'm not sure I understand how that would happen unless you are using an older version of the discovery service. However, in that case, I thought the discovery service would ignore the version field in the request and just always resolve the latest/last version that was registered for that `service_class`.

Avinash2Suresh:
[...]

Issue #941, opened 2024-09-25
PR build's virtualenv caching does not handle file dependencies like `../../packages/service` correctly


bkeryan:
## Tech Debt
The PR build uses actions/cache to cache virtualenvs based on the hash of `poetry.lock`. For external dependencies, this works great, but when using file dependencies to reference other packages in the same repo such as `../../packages/service`, this doesn't work right because the contents of `../../packages/service` are not reflected in the hash of `poetry.lock`.

PR #949, opened 2024-09-26
[High Priority] Client generator: provide support backward compatibility for Older libraries

Jotheeswaran-Nandagopal:

- Replace `resolve_service` API with `resolve_service_with_information` API, to get the version of the measurement along with service location.

- The Measurement Plug-In client lacks backward compatibility for measurements that use older libraries, as the versions are set to `0.0.1` by the recent updates in the discovery service. 
- This implements contains the fetching of the measurement version from the `ServiceDescriptor` of `resolve_service_with_information` API, which solves the backward compatibility issue.
Context - #946

- Done manual testing.
- Existing automated tests passed.


packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -64,14 +64,14 @@
...
 
 
-def get_measurement_service_stub(
+def get_measurement_service_stub_and_version(
     discovery_client: DiscoveryClient,
     channel_pool: GrpcChannelPool,
     service_class: str,
-) -> v2_measurement_service_pb2_grpc.MeasurementServiceStub:
+) -> Tuple[v2_measurement_service_pb2_grpc.MeasurementServiceStub, str]:
     """Returns the measurement service stub of the given service class."""
```
jayaseelan-james, line 72: Update the documentation.

jayaseelan-james (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Jotheeswaran-Nandagopal: Thanks for pointing. I have updated it.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -81,8 +81,17 @@ def get_measurement_service_stub(
             )
         else:
             raise
-    channel = channel_pool.get_channel(service_location.insecure_address)
-    return v2_measurement_service_pb2_grpc.MeasurementServiceStub(channel)
+    channel = channel_pool.get_channel(resolve_service_response.service_location.insecure_address)
+    version = (
+        resolve_service_response.service_info.versions[0]
```
bkeryan, line 86: FYI: I assume this call only returns one version, not all available versions.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -64,14 +64,14 @@
...
     discovery_client: DiscoveryClient,
     channel_pool: GrpcChannelPool,
     service_class: str,
-) -> v2_measurement_service_pb2_grpc.MeasurementServiceStub:
-    """Returns the measurement service stub of the given service class."""
+) -> Tuple[v2_measurement_service_pb2_grpc.MeasurementServiceStub, str]:
+    """Returns the measurement service stub and version of the given service class."""
     try:
-        service_location = discovery_client.resolve_service(
+        resolve_service_response = discovery_client.resolve_service_with_information(
```
bkeryan, line 74:
```suggestion
        service_location, service_info = discovery_client.resolve_service_with_information(
```

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
bkeryan, line 74: > fix: set version to empty if old discovery server is used

What does this mean? Did we previously ship a DiscoveryService that supports ResolveServiceWithInformation but which returns an empty list of versions?

@MounikaBattu17 added ResolveServiceWithInformation in June: https://github.com/ni/ni-apis/pull/51

We should either document that the client generator requires InstrumentStudio Pro 2024Q4 or whatever, or we should fall back to the old ResolveService if the new one throws RpcError with code UNIMPLEMENTED. However, the latter seems like overdoing it.

Cc: @jasonmreding

Jotheeswaran-Nandagopal: >What does this mean? Did we previously ship a DiscoveryService that supports ResolveServiceWithInformation but which returns an empty list of versions?

Previously, there is no versions in the[ service descriptor](https://github.com/ni/ni-apis/blob/78351723cfea5f751f77dfdd8ef845a7a174c12a/ni/measurementlink/discovery/v1/discovery_service.proto#L67), when Mounika added `ResolveServiceWithInformation`. The versions are added recently in https://github.com/ni/ni-apis/pull/54/files, for versioning support.

>We should either document that the client generator requires InstrumentStudio Pro 2024Q4 or whatever

We can go with the approach of throwing click error stating, the client generator requires InstrumentStudio Pro 2024 Q4.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Jotheeswaran-Nandagopal: yes, only one version will be returned.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -82,11 +82,15 @@ def get_measurement_service_stub_and_version(
         else:
             raise
     channel = channel_pool.get_channel(service_location.insecure_address)
-    version = service_info.versions[0] if service_info.versions else ""
+
+    if not service_info.versions:
```
bkeryan, line 86: The old DiscoveryService doesn't support ResolveServiceWithInformation, so if you have an older version, I don't think you'll even get this far.

I was expecting you to add a case to the `except grpc.RpcError` block for `elif e.code() == grpc.StatusCode.UNIMPLEMENTED`.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Jotheeswaran-Nandagopal: Updated to throw error.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Jotheeswaran-Nandagopal: Thanks for mentioning, I have added the check for `grpc.StatusCode.UNIMPLEMENTED`. 

When the `ni_measurement_plugin_sdk_service` is higher than 2.0.0 i.e., latest release, the `service_info.versions[0]` will throw index out of range error. To catch that `if not service_info.versions:` check was added.

Jotheeswaran-Nandagopal:
Pipeline is failing, because the `ni_measurement_plugin_sdk_service` dependency is taken from cache, instead of installing a updated service (#941), which throws `grpc.StatusCode.UNIMPLEMENTED` .

bkeryan (approved):

bkeryan:
> Pipeline is failing, because the `ni_measurement_plugin_sdk_service` dependency is taken from cache, instead of installing a updated service (#941), which throws `grpc.StatusCode.UNIMPLEMENTED` .

I expected https://github.com/ni/measurement-plugin-python/pull/944 to fix this. Please merge/rebase the latest main into your branch to be sure.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
bkeryan: I don't understand why that would happen. Have you tested with the previous version of InstrumentStudio installed?

That said, an extra error case that doesn't get hit is not a big deal, so I approved the PR.

Jotheeswaran-Nandagopal:
> I expected #944 to fix this. Please merge/rebase the latest main into your branch to be sure.

Yes, I have merged the latest main and the changes in #944 , are added to this branch.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Jotheeswaran-Nandagopal: >Have you tested with the previous version of InstrumentStudio installed?

Yes, I have tested with previous version of InstrumentStudio.

bkeryan: I see. Mounika's PR to add ResolveServiceWithInformation to DiscoveryService is in the releases/24C2 branch, so we previously shipped a DiscoveryService with incomplete support for this API.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
jasonmreding: Either way, the first version returned should be the newest version which generally seems like what the client generator should use given the UI doesn't present the user a choice.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
jasonmreding: The other option would be to remember the version from `EnumerateServices` and then pass that into the client generation rather than relying on calling `ResolveServiceWithInformation`. The advantage to this approach is the `EnumerateServices` API has always existed, it just didn't always have a `Versions` field. So if you are talking to an older version of the discovery service, this will just come back as an empty array. In that case, you could just capture the version in the client as the empty string which means it will always connect to the latest version of the service.

The down side to that approach is if you were relying on the version to indicate the generated client is incompatible with the current measurement's signature, that will no longer work.

bkeryan:
> > I expected #944 to fix this. Please merge/rebase the latest main into your branch to be sure.
> 
> Yes, I have merged the latest main and the [changes](https://github.com/ni/measurement-plugin-python/blob/c52ddef9c61bcd160372cf73dc176739fe0d8bf8/packages/generator/pyproject.toml#L40C4-L40C15) in #944 , are added to this branch.

I think the problem is that measlinkbot is not running the latest release of InstrumentStudio. @MounikaBattu17 rebuilt the measlinkbot images in late August but only deployed the lv64 one. The py32 and py64 images from August have never been launched, and the runner is still using an image from May. I will rebuild **and deploy** the py32 and py64 images.

PR #971, opened 2024-10-09
Remove requirement for check_nimg to run the unit tests

dixonjoel:

Removes the need for check_nimg to pass before running the unit tests.

We need the unit tests to run for the pre-publish steps even if check_nimg is failing.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,7 +21,7 @@ jobs:
   run_unit_tests:
     name: Run unit tests
     uses: ./.github/workflows/run_unit_tests.yml
-    needs: [check_nims, check_nimg]
+    needs: [check_nims]
   run_system_tests:
     name: Run system tests
     uses: ./.github/workflows/run_system_tests.yml


bkeryan (approved):

bkeryan:
@dixonjoel I don't think this is going to work. run_system_tests.yml is going to fail when it tries to install ni-measurement-plugin-sdk-generator. We're probably going to have to disable all checks when building against packages that don't exist, at least until we have a better way to handle workspaces in Poetry.

bkeryan:
> @dixonjoel @rajatkumar87 I don't think this is going to work. run_system_tests.yml is going to fail when it tries to install ni-measurement-plugin-sdk-generator. We're probably going to have to disable all checks when building against packages that don't exist, at least until we have a better way to handle workspaces in Poetry.

Actually, it worked: 

> Updating ni-measurement-plugin-sdk-service (2.1.0.dev2 C:/ar/_work/measurement-plugin-python/measurement-plugin-python/packages/service -> 2.1.0 C:/ar/_work/measurement-plugin-python/measurement-plugin-python/packages/service)

Because we didn't comment out the dev-dependency referencing ../../packages/service

Issue #974, opened 2024-10-10
Client generator should handle unsupported message types more gracefully


bkeryan:
# Bug Report

The client generator does not support https://github.com/ni/ni-apis/blob/main/ni/protobuf/types/waveform.proto yet. If you try to generate a client for a plugin that uses DoubleAnalogWaveform, you get a couple of misleading low-level errors.

Also, when using `--all`, this terminates code generation early, so it might not get to all of the other services.

## Repro or Code Sample

Run ASW internal build of InstrumentStudio, then run `ni-measurement-plugin-generator --all`

## Expected Behavior

`--all` skips `ni.examples.WaveformMeasurement_NET`  and prints a warning that this plug-in was skipped because it uses `ni.protobuf.types.DoubleAnalogWaveform`, which is not supported.

`-s ni.examples.WaveformMeasurement_NET` does the same, but it's an error instead of a warning.

## Current Behavior
```(.venv) PS D:\dev\measurement-plugin-python\examples\client> ni-measurement-plugin-client-generator --all
The measurement plug-in client for the service class 'ni.examples.EnumMeasurement' has been successfully created as 'enum_measurement_client.py'.
...
The measurement plug-in client for the service class 'ni.examples.UIProgressUpdates_NET' has been successfully created as 'ui_progress_updates_net_client.py'.
Traceback (most recent call last):
  File "D:\dev\measurement-plugin-python\examples\client\.venv\lib\site-packages\ni_measurement_plugin_sdk_service\_internal\parameter\serialization_descriptors.py", line 115, in create_file_descriptor
    pool.FindFileByName(service_name)
KeyError: "Couldn't find file ni.examples.WaveformMeasurement_NET"

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Administrator\.pyenv\pyenv-win\versions\3.9.13\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\Administrator\.pyenv\pyenv-win\versions\3.9.13\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "D:\dev\measurement-plugin-python\examples\client\.venv\Scripts\ni-measurement-plugin-client-generator.exe\__main__.py", line 7, in <module>
  File "D:\dev\measurement-plugin-python\examples\client\.venv\lib\site-packages\click\core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
[...]

Issue #906, opened 2024-09-20
Improve Measurement Client Generator to independently generate clients


MounikaBattu17:
## Tech Debt
Currently, in the Measurement PlugIn Client Generator, if the creation of one measurement client fails, subsequent clients fail as well. We need to modify the client creation process to make it independent for each measurement service, ensuring that a failure in one measurement client does not affect the creation of others.

[AB#2863186](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2863186)

PR #984, opened 2024-10-11
Handle unsupported message type errors in client generator

Jotheeswaran-Nandagopal:

- Updated the client generator to raise a click exception for unsupported datatypes, by validating the message type after getting the metadata.
- Added `try-except` block to catch the exceptions raised during the creation of multiple clients and for all registered measurements. Also, show warning message about client creation failure and continue client creation for subsequent measurements.

- This implementation fixes the following issues,
    - #974 
    - #906 

- Done manual testing
- Existing automation test passed

### DIFF
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
@@ -31,7 +31,7 @@ from ni_measurement_plugin_sdk_generator.client._support import (
 )
 
 
-CLIENT_CREATION_ERROR_MESSAGE = "The measurement plug-in client creation for the service class '{}' has failed. Possible reason(s): {}"
+_CLIENT_CREATION_ERROR_MESSAGE = "Client creation failed for '{}'. Possible reason(s): {}"
 
 
 def _render_template(template_name: str, **template_args: Any) -> bytes:
@@ -117,6 +117,7 @@ def _create_all_clients(directory_out: Optional[str]) -> None:
     discovery_client = DiscoveryClient(grpc_channel_pool=channel_pool)
 
     generated_modules: List[str] = []
+    errors: List[str] = []
     directory_out_path = resolve_output_directory(directory_out)
     measurement_service_classes, _ = get_all_registered_measurement_info(discovery_client)
     validate_measurement_service_classes(measurement_service_classes)
@@ -139,7 +140,10 @@ def _create_all_clients(directory_out: Optional[str]) -> None:
                 generated_modules=generated_modules,
             )
         except Exception as e:
-            click.echo(CLIENT_CREATION_ERROR_MESSAGE.format(service_class, e))
+            errors.append(_CLIENT_CREATION_ERROR_MESSAGE.format(service_class,e))
+
+    if errors:
+        raise click.ClickException(errors[0])
 
 
 def _create_clients_interactively() -> None:
@@ -196,7 +200,7 @@ def _create_clients_interactively() -> None:
                 generated_modules=generated_modules,
             )
         except Exception as e:
-            click.echo(CLIENT_CREATION_ERROR_MESSAGE.format(service_class, e))
+            click.echo(_CLIENT_CREATION_ERROR_MESSAGE.format(service_class, e))
 
 
 def _create_clients(
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
@@ -31,7 +31,7 @@ from ni_measurement_plugin_sdk_generator.client._support import (
 )
 
 
-CLIENT_CREATION_ERROR_MESSAGE = "The measurement plug-in client creation for the service class '{}' has been failed. Possible reason(s): {}"
+CLIENT_CREATION_ERROR_MESSAGE = "The measurement plug-in client creation for the service class '{}' has failed. Possible reason(s): {}"
 
 
 def _render_template(template_name: str, **template_args: Any) -> bytes:
@@ -138,7 +138,6 @@ def _create_all_clients(directory_out: Optional[str]) -> None:
             generated_modules.append(module_name)
         except Exception as e:
             click.echo(CLIENT_CREATION_ERROR_MESSAGE.format(service_class, e))
-            continue
 
 
 def _create_clients_interactively() -> None:
@@ -232,7 +231,6 @@ def _create_clients(
             generated_modules.append(module_name)
         except Exception as e:
             click.echo(CLIENT_CREATION_ERROR_MESSAGE.format(service_class, e))
-            continue
 
 
 @click.command()
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
@@ -142,8 +142,14 @@ def _create_all_clients(directory_out: Optional[str]) -> None:
         except Exception as e:
             errors.append(_CLIENT_CREATION_ERROR_MESSAGE.format(service_class, e))
 
-    if errors:
+    if len(errors) == 1:
         raise click.ClickException(errors[0])
+    elif len(errors) > 1:
+        raise click.ClickException(
+            "Client creation failed for multiple measurement plug-ins:\n\n{}".format(
+                "\n\n".join(errors)
+            )
+        )
 
 
 def _create_clients_interactively() -> None:
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
@@ -15,16 +15,18 @@ from ni_measurement_plugin_sdk_service.discovery import DiscoveryClient
 from ni_measurement_plugin_sdk_service.grpc.channelpool import GrpcChannelPool
 
 from ni_measurement_plugin_sdk_generator.client._support import (
-    get_class_name,
+    create_class_name,
+    create_module_name,
+    extract_base_service_class,
     get_configuration_and_output_metadata_by_index,
     get_configuration_parameters_with_type_and_default_values,
     get_measurement_service_stub_and_version,
-    get_module_name,
     get_output_parameters_with_type,
     get_all_registered_measurement_info,
     get_selected_measurement_service_class,
     to_ordered_set,
     resolve_output_directory,
+    validate_identifier,
     validate_measurement_service_classes,
 )
 
@@ -121,8 +123,11 @@ def _create_all_clients(directory_out: Optional[str]) -> None:
 
     for service_class in measurement_service_classes:
         try:
-            module_name = get_module_name(service_class, generated_modules)
-            class_name = get_class_name(service_class)
+            base_service_class = extract_base_service_class(service_class)
+            module_name = create_module_name(base_service_class, generated_modules)
+            class_name = create_class_name(base_service_class)
+            validate_identifier(module_name, "module")
+            validate_identifier(class_name, "class")
 
             _create_client(
                 channel_pool=channel_pool,
@@ -165,8 +170,21 @@ def _create_clients_interactively() -> None:
         )
 
         try:
-            module_name = get_module_name(service_class, generated_modules, True)
-            class_name = get_class_name(service_class, True)
+            base_service_class = extract_base_service_class(service_class)
+            default_module_name = create_module_name(base_service_class, generated_modules)
+            module_name = click.prompt(
+                "Enter a name for the Python client module, or press Enter to use the default name.",
+                type=str,
+                default=default_module_name,
+            )
+            validate_identifier(module_name, "module")
+            default_class_name = create_class_name(base_service_class)
+            class_name = click.prompt(
+                "Enter a name for the Python client class, or press Enter to use the default name.",
+                type=str,
+                default=default_class_name,
+            )
+            validate_identifier(class_name, "class")
 
             _create_client(
                 channel_pool=channel_pool,
@@ -179,7 +197,6 @@ def _create_clients_interactively() -> None:
             )
         except Exception as e:
             click.echo(CLIENT_CREATION_ERROR_MESSAGE.format(service_class, e))
-            break
 
 
 def _create_clients(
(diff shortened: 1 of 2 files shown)


packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -161,30 +168,34 @@ def _create_clients_interactively() -> None:
...
+                discovery_client=discovery_client,
+                measurement_service_class=service_class,
+                module_name=module_name,
+                class_name=class_name,
+                directory_out=directory_out_path,
+            )
+        except Exception as e:
+            click.echo(CLIENT_CREATION_ERROR_MESSAGE.format(service_class, e))
+            break
         generated_modules.append(module_name)
```
Avinash2Suresh, line 199: The current logic appends the module name even when client generation fails. Therefore, this should be moved inside the try block.

Avinash2Suresh (approved):
Approved with suggestion.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -201,23 +212,27 @@ def _create_clients(
...
-            directory_out=directory_out_path,
-        )
-        generated_modules.append(module_name)
+        try:
+            base_service_class = extract_base_service_class(service_class)
+            if has_multiple_service_classes or module_name is None:
+                module_name = create_module_name(base_service_class, generated_modules)
+            if has_multiple_service_classes or class_name is None:
+                class_name = create_class_name(base_service_class)
+            validate_identifier(module_name, "module")
```
dixonjoel, line 221: I think the 12 or so lines from 221 - 232 can be factored into a sub-method and called from these 3 places.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -148,6 +153,11 @@ def get_configuration_and_output_metadata_by_index(
 
     output_parameter_list = []
     for output in metadata.measurement_signature.outputs:
+        if output.message_type and output.message_type != "ni.protobuf.types.DoubleXYData":
+            raise click.ClickException(
+                "Measurement outputs with message datatype other than DoubleXYData are not supported."
```
dixonjoel, line 158: Include `output.message_type` similar to this:
f"Measurement outputs do not support {output.message_type}. Only DoubleXYData is supported."

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -123,6 +123,11 @@ def get_configuration_and_output_metadata_by_index(
     """Returns the configuration and output metadata of the measurement."""
     configuration_parameter_list = []
     for configuration in metadata.measurement_signature.configuration_parameters:
+        if configuration.message_type:
+            raise click.ClickException(
+                "Measurement configurations with message datatype are not supported."
```
dixonjoel, line 128: Include bad message_type like below.
f"Measurement configurations do not support message data types ({configuration.message_type} is unsupported)."

dixonjoel (changes requested):
Couple of code changes I'll look at after they're done.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -201,23 +212,27 @@ def _create_clients(
...
+                discovery_client=discovery_client,
+                measurement_service_class=service_class,
+                module_name=module_name,
+                class_name=class_name,
+                directory_out=directory_out_path,
+            )
+            generated_modules.append(module_name)
+        except Exception as e:
+            click.echo(CLIENT_CREATION_ERROR_MESSAGE.format(service_class, e))
+            continue
```
jayaseelan-james, line 235: nit: The `continue` statement skips the execution of the code block after the continue statement and forces the control to start the next iteration. So, having it at the end of the iteration has no effect.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -117,21 +120,25 @@ def _create_all_clients(directory_out: Optional[str]) -> None:
...
```
jayaseelan-james, line 141: nit: The `continue` statement skips the execution of the code block after the continue statement and forces the control to start the next iteration. So, having it at the end of the iteration has no effect.

Jotheeswaran-Nandagopal: Thanks for pointing. I missed to remove it.
I have updated now.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Jotheeswaran-Nandagopal: I have updated the error messages.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
Jotheeswaran-Nandagopal: Sure, I have refactored the code as follows:
- Added a method to `get_module_name` that either generates the module name from the service class or receives it as input from the user, depending on the client generation mode. This method also validates the module name.
- Added a method to `get_class_name`, which functions similarly to `get_module_name`, generating or obtaining the class name and validating it.
- Also, I have moved the `generated_modules.append(module_name)` to `_create_client` method.

Jotheeswaran-Nandagopal:
@dixonjoel , @bkeryan 
I have refactored the code and implemented the requested changes. Please help me reviewing the changes.

dixonjoel (approved):

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -320,46 +325,43 @@ def resolve_output_directory(directory_out: Optional[str] = None) -> pathlib.Pat
...
+    base_service_class = _extract_base_service_class(service_class)
+    module_name = _create_module_name(base_service_class, generated_modules)
+
+    if is_interactive_mode:
+        module_name = click.prompt(
+            "Enter a name for the Python client module, or press Enter to use the default name.",
+            type=str,
+            default=module_name,
         )
+    _validate_identifier(module_name, "module")
```
bkeryan, line 343: Please don't scatter the interactive code all over the place. See previous discussion:

https://github.com/ni/measurement-plugin-python/pull/883#discussion_r1759405388

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -148,6 +153,11 @@ def get_configuration_and_output_metadata_by_index(
 
     output_parameter_list = []
     for output in metadata.measurement_signature.outputs:
+        if output.message_type and output.message_type != "ni.protobuf.types.DoubleXYData":
+            raise click.ClickException(
+                f"Measurement outputs do not support {output.message_type}. Only DoubleXYData is supported."
+            )
```
bkeryan, line 159: Ok, I think this should prevent `TypeError: Couldn't build proto file into descriptor pool` but I haven't tested it yet.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -201,23 +195,23 @@ def _create_clients(
...
+                channel_pool=channel_pool,
+                discovery_client=discovery_client,
+                measurement_service_class=service_class,
+                module_name=module_name,
+                class_name=class_name,
+                directory_out=directory_out_path,
+                generated_modules=generated_modules,
+            )
+        except Exception as e:
+            click.echo(CLIENT_CREATION_ERROR_MESSAGE.format(service_class, e))
```
bkeryan, line 214: This treats all errors as success. 

Batch mode should still report errors as errors, but it should try to generate all of the files first and then report the errors (or the first error).

It's ok for interactive mode to print errors without exiting.

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
Avinash2Suresh: >Batch mode should still report errors as errors, but it should try to generate all of the files first and then report the errors (or the first error).

During Batch mode, the user can provide multiple measurement service classes to create multiple clients at the same time. Are you expecting to create clients for all the successful cases first and then throw an error for the failed cases? If that’s the case, and we encounter errors for more than one measurement, are you expecting to report multiple errors based on the number of failed client creations?

Avinash2Suresh: I see that Batch mode is quite similar to the --all command, as the user can create clients for multiple measurements by specifying each class using -s command. That’s why we followed the same approach for both Batch mode and -all mode.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Avinash2Suresh: Yes, both the above handling and https://github.com/ni/measurement-plugin-python/pull/984/files/78b383f49f860c8c2a4ba923c301790fe8301339#:~:text=.message_type%3A-,raise%20click.ClickException(,-f%22Measurement%20configurations handling will prevent the TypeError. I’ve also verified this with the waveform measurement in .NET.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
bkeryan: I was using "batch mode" to refer to both -a and -s, since neither is interactive.

I guess the "best effort" behavior is mainly important for -a. You asked for all services, and some encountered errors, but here are the ones that worked.

For -s, I guess it's ok to fail fast on the 1st error. You asked for N specific services, so there's not as much benefit in making the best effort to continue after an error.

Jotheeswaran-Nandagopal: I have remove the try except of "-s" mode and allowed to fail. Only "--all" and "--interactive" mode will print warning and continue next iteration.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Jotheeswaran-Nandagopal: I have reverted the refactored methods that contains the scattering of interactive code.

Jothees07:
@bkeryan, I have updated the changes. Kindly, help us reviewing the changes at your earliest convenience.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
```diff
@@ -31,6 +31,9 @@
 )
 
 
+CLIENT_CREATION_ERROR_MESSAGE = "The measurement plug-in client creation for the service class '{}' has failed. Possible reason(s): {}"
```
bkeryan, line 34: This shouldn't be public.

bkeryan, line 34: Copy the 1st sentence of this because it's more concise and readable: https://github.com/ni/measurement-plugin-python/blob/b9efedf4df35ef75a557fc91446f44c246d66cb6/packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py#L338

> f"Client creation failed for '{service_class}'.\nEither provide a module name or update the measurement with a valid service class."

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
bkeryan: > it should try to generate all of the files first and then report the errors (or the first error).

My expectation is that it will still report the errors in the normal way. For a non-interactive program, printing the error to the console and returning an exit code of 0 (success) does not really report the error because the calling process does not know there was an error.

For --all, please do something like this:

```
errors = []
for ...:
    try:
       ...
    except Exception as e:
       errors.append(e)

if errors:
   raise errors[0]
```

Printing the error to the console without exiting is appropriate for interactive mode.

bkeryan (changes requested):

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
Jotheeswaran-Nandagopal: Updated to private.

packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py
Jotheeswaran-Nandagopal: I have updated the code as suggested. 
Now, the error will be thrown for the first failure, along with its service class, to clearly indicate which measurement has failed.

Jotheeswaran-Nandagopal: If multiple clients creation fails, only the first error will be indicated to the user. 
Should we have to concatenate the errors and throw a single error stating all the failures, to provide more clarity on the errors. May I know your thoughts?

bkeryan: After testing your change, I agree, it's better to report all of the errors.

Let's do this. Leaving a blank line between errors makes them easier to distinguish.
```
    if len(errors) == 1: 
        raise click.ClickException(errors[0])
    elif len(errors) > 1:
        raise click.ClickException("Client creation failed for multiple measurement plug-ins:\n\n{}".format("\n\n".join(errors)))
```

Output:
```
PS D:\dev\measurement-plugin-python\packages\generator> poetry run ni-measurement-plugin-client-generator --all
The measurement plug-in client for the service class 'ni.examples.EnumMeasurement' has been successfully created as 'enum_measurement_client.py'.
...
The measurement plug-in client for the service class 'ni.examples.UIProgressUpdates_NET' has been successfully created as 'ui_progress_updates_net_client.py'.
Error: Client creation failed for multiple measurement plug-ins:

Client creation failed for 'ni.examples.WaveformMeasurement_NET'. Possible reason(s): Measurement outputs do not support ni.protobuf.types.DoubleAnalogWaveform. Only DoubleXYData is supported.

Client creation failed for 'ni.tests.LoopbackMeasurement_NET'. Possible reason(s): Measurement configurations do not support message data types (ni.protobuf.types.DoubleXYData is unsupported).
```

Jotheeswaran-Nandagopal: Thanks for your suggestion. I have updated the changes as suggested.

bkeryan (approved):

Issue #851, opened 2024-08-28
Measurement client generator does not work with C# services


bkeryan:
# Bug Report

The generated measurement client gets a gRPC UNKNOWN error from C# services.

The generated measurement client generator does not specify `MeasureRequest.configuration_parameters.type_url`. C# measurement services use `Any.Unpack<T>`, which validates the type URL.

## Repro or Code Sample

Run the C# GameOfLife example.

```
PS D:\dev\measurement-plugin-python\packages\generator> poetry run ni-measurement-plugin-client-generator ni.examples.GameOfLife
PS D:\dev\measurement-plugin-python\packages\generator> poetry run python
Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import game_of_life_client
>>> client = game_of_life_client.GameOfLifeClient()
>>> client.measure()
```

## Expected Behavior

It returns a measurement.

## Current Behavior
```
>>> client.measure()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\dev\measurement-plugin-python\packages\generator\game_of_life_client.py", line 192, in measure
    for response in self._get_stub().Measure(request):
  File "D:\dev\measurement-plugin-python\packages\generator\.venv\lib\site-packages\grpc\_channel.py", line 543, in __next__
    return self._next()
  File "D:\dev\measurement-plugin-python\packages\generator\.venv\lib\site-packages\grpc\_channel.py", line 969, in _next
    raise self
grpc._channel._MultiThreadedRendezvous: <_MultiThreadedRendezvous of RPC that terminated with:
        status = StatusCode.UNKNOWN
        details = "Exception was thrown by handler."
        debug_error_string = "UNKNOWN:Error received from peer  {grpc_message:"Exception was thrown by handler.", grpc_status:2, created_time:"2024-08-28T18:45:14.0185954+00:00"}"
>
```
C:\ProgramData\National Instruments\Plug-Ins\Logs\DiscoveryService.txt:
```
===== 08/28/2024 13:45:09 ===== Information: ResolveService "ni.measurementlink.measurement.v2.MeasurementService": Local IP: "127.0.0.1" Remote IP: "127.0.0.1" DeploymentTarget: ""
[...]

PR #1002, opened 2024-10-21
Fix config/output type URLs

### LINKED ISSUE #850: Python measurement services tolerate incorrect type_url
# Bug Report

Python measurement services ignore `MeasureRequest.configuration_parameters.type_url`. C# measurement services validate this field. As a result, some clients work with Python services but not C# services.

The service should advertise the required configuration/output types in the [MeasurementSignature](https://github.com/ni/ni-apis/blob/main/ni/measurementlink/measurement/v2/measurement_service.proto#L78) message, and the client should use these type names when sending/receiving messages.

## Repro or Code Sample

Send a `MeasureRequest` with an incorrect or empty `configuration_parameters.type_url`. https://github.com/ni/measurement-plugin-python/pull/841 currently does this.

## Expected Behavior

It returns an error.

## Current Behavior

It ignores the incorrect value.

## Possible Solution

Start using `Any.unpack`.

## Context

https://github.com/ni/measurement-plugin-python/pull/841 

## Your Environment

* OS & Device: Windows
* `ni-measurement-plugin-sdk-service` version: 2.1 main
* MeasurementLink version: 2024 Q3
* Python version: 3.9

AB#2841886

### LINKED ISSUE #890: Python services return incorrect message type names from `GetMetadata` RPC
# Bug Report

`grpc_servicer.py`'s `GetMetadata` method hardcodes incorrect message type names in the `MeasurementSignature` message: https://github.com/ni/measurement-plugin-python/blob/47721338763d400476edb3fd61e142e32dde8a4e/packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py#L193

This value gets saved in each `.sfp` file: https://github.com/search?q=repo%3Ani%2Fmeasurement-plugin-python%20MeasurementConfigurations&type=code

And now we are going to save this value in the generated Python client class: #889 

## Repro or Code Sample

Inspect generated client's configuration type_url.

## Expected Behavior

The message type matches reality (i.e. `serialization_descriptors.py`): https://github.com/ni/measurement-plugin-python/blob/47721338763d400476edb3fd61e142e32dde8a4e/packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/serialization_descriptors.py

## Current Behavior

The message type is the same for all Python services and does not match reality.

Advertised message type names:
```
        measurement_signature = v1_measurement_service_pb2.MeasurementSignature(
            configuration_parameters_message_type="ni.measurementlink.measurement.v1.MeasurementConfigurations",
            outputs_message_type="ni.measurementlink.measurement.v1.MeasurementOutputs",
        )
```

Real message type names: Configurations and Outputs with some namespace:
```
        _create_message_type(input_metadata, "Configurations", file_descriptor)
        _create_message_type(output_metadata, "Outputs", file_descriptor)
```
## Context

#889 

## Your Environment

N/A

[AB#2858106](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2858106)

DavidCurtiss:

* Updates the type_url of the Configurations and Outputs of the measurement, as returned by GetMetadata
* Checks the type_url of the request's connection_parameters, and prints a warning to stderr if there's a mismatch

Fixes #890 

* Tested with GameOfLife_Python and the C# GameOfLife that the measurements work and do not print warnings
* Tested modifying the client and server side of GameOfLife_Python to expect a different type_url, and saw the warnings


packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
```diff
@@ -307,7 +320,8 @@ def GetMetadata(  # noqa: N802 - function name should be lowercase
         )
 
         measurement_signature = v2_measurement_service_pb2.MeasurementSignature(
-            configuration_parameters_message_type="ni.measurementlink.measurement.v2.MeasurementConfigurations",
+            configuration_parameters_message_type=self._service_info.service_class
+            + ".Configurations",
```
bkeryan, line 324: This changes the message type advertised in GetMetadata. This message type is stored in .sfp files and generated LV/Python client code. Do previously saved files still work after this change?

bkeryan: > I tried a previously saved .sfp that uses the SampleMeasurement, and it still worked fine. (I updated grpc_servicer.py in-place, and killed NI.Discovery.Service.)

https://github.com/ni/measurement-plugin-python/blob/main/examples/sample_measurement/SampleMeasurement.sfp contains a JSON-serialized copy of the configuration parameters:

`<Screen ClientId="{c67a0b30-2590-499f-9b92-dbd2ef02233c}" ConfigurationParameters="{&quot;@type&quot;:&quot;type.googleapis.com/ni.measurementlink.measurement.v1.MeasurementConfigurations&quot;,&quot;float in&quot;:0.06,&quot;double array in&quot;:[0.1,0.2,0.3],&quot;...`

If the `.sfp` still works, then I think this may be what's happening:
- InstrumentStudio ignores the type URL when deserializing the JSON configuration parameters.
- InstrumentStudio doesn't use this type URL when sending Measure requests; it calls GetMetadata and uses the message names returned in the measurement signature.

> But for the Python client code, yes, it would have to get regenerated. And presumably the generated LV client code, too.
>
> The reason I had to change it is because the new name is what's actually there in the descriptor_pool.
> 
> So what should I do?

First of all, #890 and #850 are separate issues. It sounds like you're trying to fix both, but you only referenced one in your PR description.

When I filed #850, I listed `Any.unpack` as a possible solution, but on further thought, it might not be one. If we use the configuration/output message names advertised in the measurement signature when we register messages with the Protobuf descriptor pool, then we are basically requiring all measurement plug-ins to advertise unique configuration/output message names. Besides #890, there may be other reasons for multiple services to advertise the same message names:
- Multiple service versions with the same service class
- Copy/pasted code (if the message name isn't derived from the service class)
- Deliberate reuse of shared message types

Also, it sounds like we should reconsider baking the configuration/output message names into the generated code. Perhaps the generated code should call `GetMetadata` and then use the message names returned in the measurement signature. However, we have to balance this against the desire for the generated code to be a snapshot of the measurement metadata & version at the time it was generated.

DavidCurtiss: > First of all, https://github.com/ni/measurement-plugin-python/issues/890 and https://github.com/ni/measurement-plugin-python/issues/850 are separate issues. It sounds like you're trying to fix both, but you only referenced one in your PR description.

I just discovered #890 today, and I've so far only fixed half of it here. If I fix the other half, too, I'll reference it.

> When I filed https://github.com/ni/measurement-plugin-python/issues/850, I listed Any.unpack as a possible solution, but on further thought, it might not be one.

Note that this is how the C# plugins behave. So is it okay there but not here?

> If we use the configuration/output message names advertised in the measurement signature when we register messages with the Protobuf descriptor pool, then we are basically requiring all measurement plug-ins to advertise unique configuration/output message names. ... Also, it sounds like we should reconsider baking the configuration/output message names into the generated code.

I think we _should_ use the advertised message names (and probably bake them into the generated code), but we need to fix it to advertise the _correct_ message names, if the user provided their own descriptor instead of using a generated one.

My concern for that, though, is if they change the contents of the message without versioning the message name. But that's more of a best practice that _they'd_ need to follow. (Though I'm now wondering if we should also auto-version the generated message names. Is it allowed to have multiple versions of the same Measurement running in parallel in the discovery service? If so then we'd need to version the message. And then we'd have to figure out how specific we need to be with the version. Just major version??)

bkeryan: > > When I filed https://github.com/ni/measurement-plugin-python/issues/850, I listed Any.unpack as a possible solution, but on further thought, it might not be one.

>
> Note that this is how the C# plugins behave. So is it okay there but not here?

Writing C# plugins is not officially supported yet. Also, there is no C# client generator yet.

My concern about using `Any.Unpack` and the measurement signature message names is for the client generator, not the service. 

If the client generator registers the measurement signature message names in the descriptor pool, then it may not work properly when there is a name collision. It may be acceptable to allow collisions as long as the descriptors are identical, but a small change to the service's metadata will change the serialized descriptor, making it incompatible with clients for other services.

Most services only use their own messages, and services that use a generated client to communicate with other services should not conflict with that generated client (aside from the "all Python services use the same names" issue).

> I think we should use the advertised message names (and probably bake them into the generated code), but we need to fix it to advertise the correct message names, if the user provided their own descriptor instead of using a generated one.

If we do this I think we should include the service version in addition to the service class, or allow clients to include a version.

DavidCurtiss: So, it's sounding like "Python measurement services tolerate incorrect type_url" is actually a _feature_ rather than a bug. Which suggests we should just reject #850, and that #890 is little more than an internal naming inconsistency.

And perhaps we should even revert #889 that implemented #851, so there's no impact on the generated client code. (And instead we would need to make our C# services more tolerant of an incorrect/missing type_url.)

bkeryan: That's one way to look at it, but I think it's a stretch to say that "multiple measurement clients share a single protobuf descriptor pool" implies "measurement clients should send incorrect type URLs."

If you follow the published APIs in ni-apis, you can write a measurement service without reusing any of NI's C#, Python, or LV code. I've done it before. :) If you do this, then it's natural to unpack the protobuf `Any` type in the normal way supported by your programming language's protobuf library.

So I think it is 100% reasonable for services to expect clients to send the type URL advertised in the service's measurement signature.

#850 goes beyond this and suggests that the Python service package should validate the type URL. I could be convinced that this lack of validation is indeed a feature. After all, Postel's Law: be conservative in what you send, be liberal in what you accept.

However, following Postel's Law also suggests that clients should set the type URL to the message type advertised in the measurement signature. I do not think intentionally sending the wrong type URL is a feature. Also, if clients don't use the message type in the measurement signature as the type URL, then what is it there for?

Also, all of this is completely orthogonal to whether clients use the message types in the measurement signature when they register messages in the protobuf descriptor pool. I think it is 100% possible for clients to follow Postel's Law but also do a custom mapping between what the service thinks the message type is and what protobuf thinks the message type is.

DavidCurtiss: Alright, I updated it to fix #890 instead, and modified the validation I added for #850 to just be a warning printed to STDERR.

packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
```diff
@@ -271,10 +275,21 @@ def _serialize_response(
...
+        pool = descriptor_pool.Default()
+        configuration_proto = pool.FindMessageTypeByName(
+            self._configuration_parameters_message_type
         )
+        configuration_message = message_factory.GetMessageClass(configuration_proto)()
+        if not request.configuration_parameters.Is(configuration_message.DESCRIPTOR):
+            expected = "type.googleapis.com/" + configuration_message.DESCRIPTOR.full_name
+            actual = request.configuration_parameters.type_url
+            sys.stderr.write(f"Note: Wrong message type. Expected {expected!r} but got {actual!r}")
+        elif not request.configuration_parameters.Unpack(configuration_message):
```
bkeryan, line 291: Is this unpacking the message and throwing it away, for validation purposes? I don't think that's necessary. All `Unpack` does is use `Is` to validate the type URL then `ParseFromString`, which I assume we are still doing elsewhere since this is a validate function and it doesn't return the unpacked message.

https://github.com/protocolbuffers/protobuf/blob/9f08ec543be4a38f7831b5b7cea07fe9015cb3c2/python/google/protobuf/internal/well_known_types.py#L60

packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
```diff
@@ -271,10 +275,21 @@ def _serialize_response(
...
+    def _validate_parameters(self, request: v1_measurement_service_pb2.MeasureRequest) -> None:
+        pool = descriptor_pool.Default()
+        configuration_proto = pool.FindMessageTypeByName(
+            self._configuration_parameters_message_type
         )
+        configuration_message = message_factory.GetMessageClass(configuration_proto)()
+        if not request.configuration_parameters.Is(configuration_message.DESCRIPTOR):
+            expected = "type.googleapis.com/" + configuration_message.DESCRIPTOR.full_name
+            actual = request.configuration_parameters.type_url
+            sys.stderr.write(f"Note: Wrong message type. Expected {expected!r} but got {actual!r}")
```
bkeryan, line 290: Please update the client to warn if the output message has the wrong type URL.

packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
```diff
@@ -181,6 +182,8 @@ def __init__(
         self._measure_function = measure_function
         self._owner = weakref.ref(owner) if owner is not None else None  # avoid reference cycle
         self._service_info = service_info
+        self._configuration_parameters_message_type = service_info.service_class + ".Configurations"
```
bkeryan, line 185: I'm ok with fixing this assuming that the user impact is limited to emitting warnings in some cases.

bkeryan, line 185: We should think through what happens with multiple versions of the same service class, and potentially write some tests. This can be deferred to a separate PR / work item though.

bkeryan (changes requested):

packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
DavidCurtiss: If I just directly check the type_url of the received object, is that just as good as calling `Is`? Because that would remove any issues with conflicts when trying to instantiate a message via the message_factory. Though I suppose there might still be some problems when trying to load the measurement in the first place (wherein it loads its types into the descriptor_pool).

packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
DavidCurtiss: Yeah, I was just doing it for validation purposes, thinking it would validate the parameters. But since it does nothing more than `Is` and `ParseFromString`, I'll drop it.

packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
bkeryan: Yes, it's fine. At this point, I'm convinced that it's a feature for the client to insulate itself & the descriptor pool from the service's advertised message names.

packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako
```diff
@@ -242,13 +244,25 @@ class ${class_name}:
...
                 self._output_metadata,
                 response.outputs.value,
                 f"{self._service_class}.Outputs",
             )
         )
+
+    def _validate_response(
+        self, response: v2_measurement_service_pb2.MeasureResponse
+    ) -> None:
+        expected_type = f"type.googleapis.com/{self._service_class}.Outputs"
```
bkeryan, line 259: This should expect the advertised message type, not the message type we made up for the descriptor pool. For Python services, these are now the same thing, but for LV and C# services they aren't. Please generate a client for a C# service and manually test it. (You can load the client in the REPL and call update_pin_map() and measure().)

bkeryan (changes requested):

packages/service/tests/acceptance/test_logging.py
```diff
@@ -105,9 +105,12 @@ def test___streaming_data_measurement___measure___server_call_logged(
         )
         num_responses = 10
 
+        metadata = stub_v2.GetMetadata(v2_measurement_service_pb2.GetMetadataRequest())
```
bkeryan, line 108: These tests have arrange/act/assert blocks separated by one blank line. Remove the blank line above metadata so it's part of the arrange block.
```suggestion
        )
        num_responses = 10
        metadata = stub_v2.GetMetadata(v2_measurement_service_pb2.GetMetadataRequest())
```

packages/service/tests/acceptance/test_measurement_service.py
```diff
@@ -154,6 +160,8 @@ def test___measurement_service_v1___measure_with_large_array___returns_output(
     protobuf_enum_in = ProtobufColor.WHITE
     string_array_in = ["", "TestString1", "#$%!@<*(&^~`"]
 
+    metadata = stub_v1.GetMetadata(v1_measurement_service_pb2.GetMetadataRequest())
```
bkeryan, line 163: These tests have arrange/act/assert blocks separated by one blank line. Remove the blank line above metadata so it's part of the arrange block.

packages/service/tests/acceptance/test_session_management.py
```diff
@@ -20,6 +20,7 @@
 from tests.utilities.stubs.pinaware.types_pb2 import Configurations, Outputs
 
 
+@pytest.mark.usefixtures("filter_wrong_configurations_message_type_warnings")
```
bkeryan, line 23: FYI, you can single-source this for the file with `pytestmark = pytest.mark.usefixtures("filter_wrong_configurations_message_type_warnings")`

packages/service/tests/acceptance/conftest.py
```diff
@@ -9,3 +13,10 @@
 def pin_map_directory(test_assets_directory: pathlib.Path) -> pathlib.Path:
     """Test fixture that returns the pin map directory."""
     return test_assets_directory / "acceptance" / "session_management"
+
+
+@pytest.fixture
+def filter_wrong_configurations_message_type_warnings() -> Generator:
```
bkeryan, line 19: There's already a mark for this: https://docs.pytest.org/en/6.2.x/warnings.html#pytest-mark-filterwarnings

bkeryan (changes requested):

packages/service/tests/acceptance/test_session_management.py
bkeryan: The docs for pytest.mark.filterwarnings also point this out.

packages/service/tests/acceptance/conftest.py
DavidCurtiss: Yes, but I didn't want to have this repeated everywhere:
```py
@pytest.mark.filterwarnings(
    "ignore:.*Configurations.*:ni_measurement_plugin_sdk_service.measurement.WrongMessageTypeWarning"
)
```
(You have to use the fully-qualified name when it's not a built-in warning, and the line then becomes too long to fit onto one line.)

If you disagree, I can switch it.

bkeryan: Ok, that seems like a good reason.

bkeryan (approved):

PR #1026, opened 2024-11-29
Input Support for Double2dArray for Client Generator and Service and updating Sample Measurement

Blahpapto1:

Adding Input Support for Measurment Client Generator and Service and updating Sample Measurement.

Adding Input Support for Client Generator, Service and Sample Measurement

- Automated tests are passing after adding Double2DArray Input parameter.
- Measurement is running fine through commandline.
- Tested on MeasurementUI Editor with Sample Measurement.

### DIFF
--- a/examples/sample_measurement/measurement.py
+++ b/examples/sample_measurement/measurement.py
@@ -52,6 +52,11 @@ class Color(Enum):
 @measurement_service.configuration(
     "String Array In", nims.DataType.StringArray1D, ["String1", "String2"]
 )
+@measurement_service.configuration(
+    "Double 2D Array In",
+    nims.DataType.Double2DArray,
+    array_pb2.Double2DArray(rows=2, columns=3, data=[1, 2, 3, 4, 5, 6]),
+)
 @measurement_service.output("Float out", nims.DataType.Float)
 @measurement_service.output("Double Array out", nims.DataType.DoubleArray1D)
 @measurement_service.output("Bool out", nims.DataType.Boolean)
@@ -70,6 +75,7 @@ def measure(
     enum_input: Color,
     protobuf_enum_input: color_pb2.ProtobufColor.ValueType,
     string_array_in: Iterable[str],
+    double_2d_array_input: array_pb2.Double2DArray,
 ) -> Tuple[
     float,
     Iterable[float],
@@ -95,7 +101,7 @@ def measure(
     enum_output = enum_input
     protobuf_enum_output = protobuf_enum_input
     string_array_output = string_array_in
-    double_2d_array_output = array_pb2.Double2DArray()
+    double_2d_array_output = double_2d_array_input
     logging.info("Completed measurement")
 
     return (
--- a/packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
+++ b/packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
@@ -124,7 +124,7 @@ def get_configuration_and_output_metadata_by_index(
     """Returns the configuration and output metadata of the measurement."""
     configuration_parameter_list = []
     for configuration in metadata.measurement_signature.configuration_parameters:
-        if configuration.message_type:
+        if configuration.message_type and configuration.message_type != "ni.protobuf.types.Double2DArray":
             raise click.ClickException(
                 f"Measurement configurations do not support message data types ({configuration.message_type} is unsupported)."
             )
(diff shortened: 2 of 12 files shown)


Blahpapto1:
For parsing Double2DArray value we are converting the default_value to string in `_support.py`. Also, handling repr(value) conversions in `get_configuration_metadata_with_index_values`. This is one of the approach that we think is the best way to handle these conversions.

Another way, we can handle the parsing of Double2DArray value is to handle it after render_template method creates the output then we fix all the instance in the generated template output through regex. But this happens during the `_create_file` method call, which may or may not be the best place to handle this case.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
```diff
@@ -209,6 +212,18 @@ def get_configuration_and_output_metadata_by_index(
             default_value = default_value.value
         elif isinstance(default_value, list) and any(isinstance(e, Enum) for e in default_value):
             default_value = [e.value for e in default_value]
+        elif isinstance(default_value, str):
+            default_value = repr(default_value)
+        elif configuration_metadata[id].message_type == "ni.protobuf.types.Double2DArray":
```
dixonjoel, line 217: Can you give an example of what 'default_value' is when it comes in here and what it looks like after you reconstruct it?

dixonjoel: I think this approach is ok. I'd like to hear what @bkeryan or @mshafer-NI think.

examples/sample_measurement/measurement.py
bkeryan, line 1: > - [ ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/measurement-plugin-python/blob/main/CONTRIBUTING.md).
>
> ### What does this Pull Request accomplish?
>
> TODO: Include high-level description of the changes in this pull request.
> 
> ### Why should this Pull Request be merged?
> 
> TODO: Justify why this contribution should be part of the project.
> 
> ### What testing has been done?
> 
> TODO: Detail what testing has been done to ensure this submission meets requirements.

Fill this out.

packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py
Blahpapto1: Default Value for Double2DArray before reconstruct -
rows: 2
columns: 3
data: 1
data: 2
data: 3
data: 4
data: 5
data: 6

After Reconstruct -
'Double2DArray(rows=2, columns=3, data=[1, 2, 3, 4, 5, 6])'

Blahpapto1: Now, the error (without reconstruct) is thrown because the generated template has Double2DArray value as `default_value=rows: 2\ncolumns: 3\ndata: 1\ndata: 2\ndata: 3\ndata: 4\ndata: 5\ndata: 6\n,\r\n`

and it is passed to this [method](https://github.com/ni/measurement-plugin-python/blob/7ea52e5a2b45da2af95e3f66e0563a41239ccfdd/packages/generator/ni_measurement_plugin_sdk_generator/client/__init__.py#L50)

Blahpapto1:
```
AssertionError: assert 1 == 0
E        +  where 1 = <Result InvalidInput('Cannot parse: 242:34:                 default_value=rows: 2')>.exit_code

tests\acceptance\test_client_generator.py:44: AssertionError
```

clintverghese: @bkeryan If I understand correctly, your point here is that `default_value` should contain the python data type (`Double2DArray` in this case) and any formatting should be done in the `_format_default_value` method defined in the template file.
`_format_default_value` should be checking if `default_value` is an instance of Double2DArray and do the necessary steps to convert it to the `'Double2DArray(rows=2, columns=3, data=[1, 2, 3, 4, 5, 6])'` string. 

That makes sense. 

If the above understanding is correct then. one concern is that currently the `default_value` is added to the generated file not only through the `_format_default_value` method. [Here](https://github.com/ni/measurement-plugin-python/blob/7ea52e5a2b45da2af95e3f66e0563a41239ccfdd/packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py#L263C81-L263C94) the default value is being used directly and the reason that things work fine is because [this piece of code](https://github.com/ni/measurement-plugin-python/blob/7ea52e5a2b45da2af95e3f66e0563a41239ccfdd/packages/generator/ni_measurement_plugin_sdk_generator/client/_support.py#L229C1-L230C48) incorporates the [special string logic in _format_default_value](https://github.com/ni/measurement-plugin-python/blob/7ea52e5a2b45da2af95e3f66e0563a41239ccfdd/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako#L10C1-L11C31).
Is that intentional? 
If yes then, whatever `Double2DArray` conversion logic that is added into `_format_default_value` will need to be added here as well, or we would need this code to be passed through the `_format_default_value` workflow.

bkeryan: @clintverghese It may look like this is duplicated code, but `get_configuration_parameters_with_type_and_default_values` and `_format_default_value` are doing slightly different things.

`get_configuration_parameters_with_type_and_default_values`  generates Python code for the client API class. In some cases we use more Pythonic data types in the client API. For example, path parameters are `pathlib.Path` in the client API and `str` in protobuf. 

`_format_default_value` is used when generating Python code for the service's metadata structs. This tends to use the protobuf types or basic Python types, not stuff like `pathlib.Path`.

I think it would be fine to single-source `_format_default_value` by moving it into `_support.py` so that `get_configuration_parameters_with_type_and_default_values` can call it for non-path cases. Also, it would be helpful to update the docstrings to clarify this difference.

Blahpapto1: Moved the `_format_default_value` method to `_support.py` and added logic for Double2DArray. Now, we are calling this method from `get_configuration_parameters_with_type_and_default_values` for string and Double2dArray parameter types in `_support.py` and for the default value conversions in the template. Also, updated Docstring for both the methods.

Issue #335, opened 2023-07-17
`grpc_servicer` should use `*args` to pass positional parameters


bkeryan:
## Tech Debt

The `MeasurementService.configuration_parameter` decorator describes positional parameters in order, but `grpc_servicer` uses `inspect.signature` to look up the parameter names so that it can use `**kwargs` to pass them:

https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/_internal/grpc_servicer.py#L98
https://github.com/ni/measurementlink-python/blob/main/ni_measurementlink_service/_internal/grpc_servicer.py#L245

Passing positional parameters with `*args` would be simpler and have slightly lower overhead.

[AB#2457623](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2457623)

[AB#3037070](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3037070)

bkeryan:
This also affects whether the measurement function's parameter list is allowed to contain keyword-only parameters, positional-only parameters, `*args`, or `**kwargs`.

It could make sense to continue using `inspect.signature`, but build up separate `args` and `kwargs` based on `parameter.kind`:
- `POSITIONAL_ONLY`, `POSITIONAL_OR_KEYWORD` -> `args`
- `KEYWORD_ONLY` -> `kwargs`
- `VAR_POSITIONAL`, `VAR_KEYWORD` -> ignored?

PR #1070, opened 2025-02-26
Allow default arguments to be used for measure() method

chadallee:
I want this to work (default parameter value):

```python
@measurementservice.output("my_output", nims.DataType.Double)
@measurementservice.configuration("my_param_1", nims.DataType.Double)
#@measurementservice.configuration("my_param_2_still_debugging_might_not_use_this", nims.DataType.Double)
def measure(my_param_1,
            my_param_2_still_debugging_might_not_use_this = None):
    return (0.0,)
```

Allows default parameters to be used from the `measure()` method for the measurement plugin instead of throwing an exception.

Currently, the above plugin script will throw an exception (and not a particularly helpful one):
> 2025-02-26 14:35:42,820 ERROR: Exception iterating responses: 2
  File ".venv\Lib\site-packages\grpc\_server.py", line 654, in _take_response_from_response_iterator
    return next(response_iterator), True
           ^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv\Lib\site-packages\ni_measurement_plugin_sdk_service\grpc\loggers.py", line 391, in __next__
    response = next(self._inner_iterator)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv\Lib\site-packages\ni_measurement_plugin_sdk_service\_internal\grpc_servicer.py", line 381, in Measure
    mapping_by_variable_name = _get_mapping_by_parameter_name(
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv\Lib\site-packages\ni_measurement_plugin_sdk_service\_internal\grpc_servicer.py", line 139, in _get_mapping_by_parameter_name
    mapping_by_variable_name[parameter.name] = mapping_by_id[i]
                                               ~~~~~~~~~~~~~^^^
KeyError: 2

This problem is admittedly easy to workaround (comment out both the `@measurementservice.configuration()` AND the method parameter), but this fixes a corner case that really doesn't have to error -- and it makes the exception that gets thrown when there is no default much better .

My measurement plugin works and uses default parameters for any arguments that weren't specified with `@measurementservice.configuration()`.

Also tested that any arguments **without** a default parameter that also weren't configured with `@measurementservice.configuration()` does cause an exception to be thrown -- and the exception is more useful:

> 2025-02-26 14:23:23,980 ERROR: Exception iterating responses: measure() missing 1 required positional argument: 'my_param_2_still_debugging_might_not_use_this'

### DIFF
--- a/packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
@@ -134,7 +134,8 @@ def _get_mapping_by_parameter_name(
     signature = inspect.signature(measure_function)
     mapping_by_variable_name = {}
     for i, parameter in enumerate(signature.parameters.values(), start=1):
-        mapping_by_variable_name[parameter.name] = mapping_by_id[i]
+        if i in mapping_by_id:
+            mapping_by_variable_name[parameter.name] = mapping_by_id[i]
     return mapping_by_variable_name
 
 


packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
bkeryan, line 1: Would this also solve your problem?

https://github.com/ni/measurement-plugin-python/issues/335

bkeryan, line 1: If we're going to allow this use case, we should update one of the test services to use it.

https://github.com/ni/measurement-plugin-python/tree/main/packages/service/tests/utilities/measurements

bkeryan, line 1: What happens if the service defines a keyword-only or positional-only parameter?

Perhaps we should filter parameters based on `default` and `kind`.
https://docs.python.org/3/library/inspect.html#inspect.Parameter

chadallee: Yes, absolutely. That would be great.

chadallee: Interesting! I was not aware of keyword-only or positional-only parameters.

With this change, the behavior of position-only arguments is that you get the following exception:
> 2025-02-27 12:41:16,764 ERROR: Exception iterating responses: measure() got some positional-only arguments passed as keyword arguments: 'In_Boolean'
  File ".\Lib\site-packages\grpc\_server.py", line 654, in _take_response_from_response_iterator
    return next(response_iterator), True
           ^^^^^^^^^^^^^^^^^^^^^^^
  File ".\.venv\Lib\site-packages\ni_measurement_plugin_sdk_service\grpc\loggers.py", line 391, in __next__
    response = next(self._inner_iterator)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".\.venv\Lib\site-packages\ni_measurement_plugin_sdk_service\_internal\grpc_servicer.py", line 389, in Measure
    return_value = self._measure_function(**mapping_by_variable_name)
    
So I guess the alternative you're suggesting would be to filter out that positional-only argument, which would stop the exception from being thrown, but now we're just silently passing nothing to `measure()` function and allowing it to use its default value instead of the actual input.

I feel like throwing the exception seems better in this case. And as you said, using *args would be even better still.

chadallee:
> ```
> @measurementservice.configuration("my_param_1", nims.DataType.Double)
> #@measurementservice.configuration("my_param_2_still_debugging_might_not_use_this", nims.DataType.Double)
> def measure(my_param_1,
>             my_param_2_still_debugging_might_not_use_this : None):
> ```
> 
> @chadallee This is a type hint that says the type of `my_param_2_still_debugging_might_not_use_this ` is `None`. Did you mean `= None`?
> 
> There are two types of default parameters in Python and they are often confused.
> 
> 1. The parameter can have a default value.
> 
> ```
> def measure(my_param: str = "", my_param_2: float = 0.0):
> ```
> 
> 2. The parameter's type can accept None in addition to another type.
> 
> ```
> # Python 3.8
> def measure(my_param: typing.Optional[str] = None, my_param_2: typing.Optional[float] = 0.0):
> # Python 3.9+
> def measure(my_param: str | None = None, my_param_2: float | None = None):
> ```
> 
> BTW, this package has its own way of specifying default values, which allows the service to return the default values over gRPC so that they can be displayed in InstrumentStudio or TestStand:
> 
> ```
> @measurement_service.configuration("Float In", nims.DataType.Float, 0.06)
> @measurement_service.configuration("String In", nims.DataType.String, "sample string")
> def measure(
>     float_input: float,
>     string_input: str,
> )
> ```

Oops, good catch. Sorry, I hastily trimmed down the script I was actually testing with and changed the `=` to `:`.
I've fixed my initial PR description

packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
bkeryan: Yeah, I agree filtering out parameters doesn't make sense.

Switching to `*args` would break compatibility with keyword-only parameters.

I think the most compatible approach would be to use `kind` to decide whether to add the parameter to `args` or `kwargs`. I guess that's orthogonal to ignoring function parameters that don't correspond to "configurations", though.

packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
```diff
@@ -134,7 +134,8 @@ def _get_mapping_by_parameter_name(
     signature = inspect.signature(measure_function)
     mapping_by_variable_name = {}
     for i, parameter in enumerate(signature.parameters.values(), start=1):
-        mapping_by_variable_name[parameter.name] = mapping_by_id[i]
+        if i in mapping_by_id:
+            mapping_by_variable_name[parameter.name] = mapping_by_id[i]
```
bkeryan, line 138: You can avoid an extra map lookup here:
```
_UNSET_MAPPING = object()
...
def _get_mapping_by_parameter_name(...):
    ....
    for i, parameter in enumerate(signature.parameters.values(), start=1):
        mapping = mapping_by_id.get(i, _UNSET_MAPPING)
        if mapping is not _UNSET_MAPPING:
            mapping_by_variable_name[parameter.name] = mapping
```

packages/service/ni_measurement_plugin_sdk_service/_internal/grpc_servicer.py
bkeryan: For example, add something like `extra_param: bool = False` to https://github.com/ni/measurement-plugin-python/blob/main/packages/service/tests/utilities/measurements/loopback_measurement/__init__.py

bkeryan (changes requested):

PR #1092, opened 2025-03-21
Update to Python 3.9 syntax

bkeryan:

Use [PEP 585 – Type Hinting Generics In Standard Collections](https://peps.python.org/pep-0585/) and do other Python 3.9 syntax upgrades:
- Ran `Get-ChildItem -Recurse "*.py" | ForEach-Object { pyupgrade --py39-plus $_ }` using the latest [pyupgrade](https://github.com/asottile/pyupgrade) (installed via pipx).

Use [PEP 604 – Allow writing union types as X | Y](https://peps.python.org/pep-0604/) where possible:
- Added `from __future__ import annotations` to most files containing `from typing import ...`. This allows using type hint syntax from the future.
- Re-ran pyupgrade. This uses `X | Y` syntax if the file already has `from __future__ import annotations`.
- Note: pyupgrade leaves type aliases alone. I think this is because these are not parsed as annotations. As a result, there are still a couple of references to `typing.Union`.

Manually updated codegen templates.

Reduce number of `typing` imports.

Remove usage of `typing.Optional`, which people often confuse with default arguments.

Ran ni-python-styleguide, mypy, and pytest locally. (I did not have NI-SWITCH installed, so the switch tests failed.)


packages/service/ni_measurement_plugin_sdk_service/_internal/parameter/metadata.py
```diff
@@ -20,7 +21,7 @@
 if TYPE_CHECKING:
     from google.protobuf.internal.enum_type_wrapper import _EnumTypeWrapper
 
-    SupportedEnumType = Union[Type[Enum], _EnumTypeWrapper]
+    SupportedEnumType = Union[type[Enum], _EnumTypeWrapper]
```
bkeryan, line 24: Here is an example of Union used for a type alias.

bkeryan:
To add `from __future__ import annotations`, I used `Get-ChildItem -Recurse "*.py" | ForEach-Object { python path\to\add_future.py $_ }` with this script:

add_future.py:
```
#!/usr/bin/env python3

import sys

filename = sys.argv[1]
with open(filename, "rt") as f:
    text = f.read()
if "from typing" in text and "from __future__" not in text:
    print(filename)
    text = "from __future__ import annotations\n" + text
    with open(filename, "wt") as f:
        f.write(text)
```

and I manually fixed up any files that have a module docstring.

PR #1093, opened 2025-03-21
Fix incorrect reporting of non-ASCII characters in the measurement display name

jonlugoatx:

Fixes an internationalization issue where foreign characters in the measurement display name get mangled when displayed in InstrumentStudio:
image

This display name is loaded from the .serviceconfig file and reported through the gRPC `GetMetadataResponse`. The problem was that we were loading it without specifying the encoding. We should load the .serviceconfig as utf-8 to allow for non-ASCII characters in the display name.

Ran a python measurement with a non-ascii (tm) symbol in the display name in the serviceconfig file and loaded it in InstrumentStudio:
image


packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
bkeryan, line 1: Does the client generator handle non-ASCII characters properly? It uses `repr`:

https://github.com/ni/measurement-plugin-python/blob/main/packages/generator/ni_measurement_plugin_sdk_generator/client/templates/measurement_plugin_client.py.mako#L17

bkeryan, line 1: Please add a test, for both the generator and service packages.

You can fork https://github.com/ni/measurement-plugin-python/tree/main/packages/service/tests/utilities/measurements/v2_only_measurement into a "localized_measurement" or something.

bkeryan (changes requested):

packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
```diff
@@ -204,7 +204,7 @@ def __init__(
         if not path.exists(service_config_path):
             raise RuntimeError(f"File does not exist. {service_config_path}")
 
-        with open(service_config_path) as service_config_file:
+        with open(service_config_path, "r", encoding="utf-8") as service_config_file:
```
bkeryan, line 207: Does our C# implementation accept a BOM? You can do this with `encoding="utf-8-sig"`.

jonlugoatx: The discovery service does accept the BOM, but the measurement service fails to load if it's there. I put in the `-sig`.

packages/service/ni_measurement_plugin_sdk_service/measurement/service.py
jonlugoatx: Done.

jonlugoatx: Short answer: no

I updated it to save the black cleanup results as utf-8. When writing the auto test, I realized that the client generator doesn't support converting localized configuration/output display names because those can't be converted to valid protobuf field names. This is a known issue with there not being a separation between display name and protobuf name. I'm not fixing that here.

packages/generator/tests/utilities/measurements/localized_measurement/_stubs/color_pb2.py
bkeryan, line 1: These stubs are regenerated by https://github.com/ni/measurement-plugin-python/blob/main/packages/generator/scripts/generate_grpc_stubs.py

IMO we should just delete the ProtobufColor stuff and the stubs from the localized test service.

bkeryan (approved):
Approved with suggestions

packages/generator/tests/utilities/measurements/localized_measurement/_stubs/color_pb2.py
jonlugoatx: Done.

PR #1194, opened 2025-07-04
feat: Add support for Measurement Remote Execution by populating deployment_target

divakar-selvam07:

This Pull Request implements support for the [Feature 3069888](https://dev.azure.com/ni/DevCentral/_workitems/edit/3069888): Execute measurement plug-ins using remote instruments with Session Management​.
It enables accessing instrument resources hosted on a remote test bench machine from developer machines. To achieve this, the `EnumerateComputeNodes` API that returns the remote test bench URL as a compute node is implemented in the `DiscoveryClient` and integrated into the service clients:

* gRPC Device Server Activation service
* Session Management service
* Pin map service

These clients use the `EnumerateComputeNodes` API to detect remote compute nodes and populate the `deployment_target` accordingly. As a result, all service requests are resolved and sent to the appropriate remote services instead of the local host. If no remote compute nodes are found, the system falls back to the current local behavior.

This change is essential for enabling seamless execution of measurement plug-ins using remote instrument resources. By allowing service clients to resolve and communicate with remote services transparently, developers can work remotely without changing existing workflows.

TODO: Add automation tests.

### Additional references:
* [Pull Request 948366](https://dev.azure.com/ni/DevCentral/_git/ASW/pullrequest/948366): HLD to execute measurement plug-ins using remote instruments with Session Management​
* [Pull Request 1004993](https://dev.azure.com/ni/DevCentral/_git/ASW/pullrequest/1004993): Service clients: Populate the respective clients with the deployment target

### DIFF
--- a/packages/service/ni_measurement_plugin_sdk_service/_drivers/_grpcdevice.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/_drivers/_grpcdevice.py
@@ -36,8 +36,13 @@ def get_grpc_device_server_location(
         )
         return service_location
 
+    compute_nodes = discovery_client.enumerate_compute_nodes()
+    remote_nodes = [node for node in compute_nodes if not node.is_local]
+    first_remote_node_url = remote_nodes[0].url if len(remote_nodes) == 1 else ""
+    
     service_location = discovery_client.resolve_service(
         provided_interface=provided_interface,
+        deployment_target=first_remote_node_url,
         service_class=SERVICE_CLASS,
     )
     _logger.debug(
--- a/packages/service/ni_measurement_plugin_sdk_service/discovery/_client.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/discovery/_client.py
@@ -24,6 +24,7 @@ from ni_measurement_plugin_sdk_service.grpc.channelpool import GrpcChannelPool
 from ni_measurement_plugin_sdk_service.measurement.info import (
     MeasurementInfo,
     ServiceInfo,
+    ComputeNodeDescriptor,
 )
 
 _logger = logging.getLogger(__name__)
@@ -304,3 +305,15 @@ class DiscoveryClient:
         response = self._get_stub().EnumerateServices(request)
 
         return [ServiceInfo._from_grpc(service) for service in response.available_services]
+
+    def enumerate_compute_nodes(self) -> Sequence[ComputeNodeDescriptor]:
+        """Enumerates all compute nodes registered with the discovery service.
+
+        Returns:
+            The list of information describing the compute nodes.
+        """
+        request = discovery_service_pb2.EnumerateComputeNodesRequest()
+
+        response = self._get_stub().EnumerateComputeNodes(request)
+
+        return [ComputeNodeDescriptor._from_grpc(node) for node in response.compute_nodes]
--- a/packages/service/ni_measurement_plugin_sdk_service/session_management/_client.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/session_management/_client.py
@@ -75,8 +75,14 @@ class SessionManagementClient:
                         grpc_channel_pool=self._grpc_channel_pool
                     )
                 if self._stub is None:
+                    
+                    compute_nodes = self._discovery_client.enumerate_compute_nodes()
+                    remote_nodes = [node for node in compute_nodes if not node.is_local]
+                    first_remote_node_url = remote_nodes[0].url if len(remote_nodes) == 1 else ""
+                    
                     service_location = self._discovery_client.resolve_service(
                         provided_interface=GRPC_SERVICE_INTERFACE_NAME,
+                        deployment_target=first_remote_node_url,
                         service_class=GRPC_SERVICE_CLASS,
                     )
                     channel = self._grpc_channel_pool.get_channel(service_location.insecure_address)
--- a/packages/service/ni_measurement_plugin_sdk_service/measurement/info.py
+++ b/packages/service/ni_measurement_plugin_sdk_service/measurement/info.py
@@ -80,6 +80,23 @@ class ServiceInfo(NamedTuple):
             versions=list(other.versions),
         )
 
+class ComputeNodeDescriptor(NamedTuple):
+    """A named tuple providing information about a compute node.
+
+    This class is used with the NI Discovery Service when enumerating compute nodes.
+    """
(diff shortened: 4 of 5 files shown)


jayaseelan-james (commented):
Apart from these comments, I'll let @Avinash2Suresh review the automated tests once they're added for these updates.

packages/service/ni_measurement_plugin_sdk_service/_drivers/_grpcdevice.py
```diff
@@ -36,8 +36,12 @@ def get_grpc_device_server_location(
         )
         return service_location
 
+    compute_nodes = discovery_client.enumerate_compute_nodes()
+    remote_nodes = [node for node in compute_nodes if not node.is_local]
+    first_remote_node_url = remote_nodes[0].url if len(remote_nodes) == 1 else ""
```
jayaseelan-james, line 41: On second thought, I believe we should add a comment in the code explaining why we do this, as it doesn't seem to be obvious.

packages/service/ni_measurement_plugin_sdk_service/_drivers/_grpcdevice.py
```diff
@@ -36,8 +36,12 @@ def get_grpc_device_server_location(
         )
         return service_location
 
+    compute_nodes = discovery_client.enumerate_compute_nodes()
+    remote_nodes = [node for node in compute_nodes if not node.is_local]
```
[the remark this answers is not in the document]
divakar-selvam07: Renamed to `remote_compute_nodes` in all instances.

packages/service/ni_measurement_plugin_sdk_service/session_management/_client.py
```diff
@@ -75,8 +75,16 @@ def _get_stub(self) -> session_management_service_pb2_grpc.SessionManagementServ
                         grpc_channel_pool=self._grpc_channel_pool
                     )
                 if self._stub is None:
+                    compute_nodes = self._discovery_client.enumerate_compute_nodes()
+                    remote_compute_nodes = [node for node in compute_nodes if not node.is_local]
+                    first_remote_node_url = (
+                        remote_compute_nodes[0].url if len(remote_compute_nodes) == 1 else ""
+                    )
+                    # Assigning the first available remote compute node to deployment_target
+                    # to enable remote service resolution if connected.
```
bkeryan, line 84: https://peps.python.org/pep-0020/ 

> In the face of ambiguity, refuse the temptation to guess.

Why is this class choosing the deployment target for the user? That seems like guessing.

I realize you are not trying to support the multi-test-bench-PC scenarios yet, but I wonder if you will break compatibility in order to support them.

Do the client APIs for other programming languages also do this?

I think adding the deployment target as a constructor parameter to the PinMapClient and SessionManagementClient classes would be a better API. If getting the first remote node is a very common use case, it would be reasonable to have a method on the discovery client to get it.

Cc: @jasonmreding @JasonC512 @ccifra

bkeryan (changes requested):

packages/service/ni_measurement_plugin_sdk_service/session_management/_client.py
jayaseelan-james: > I realize you are not trying to support the multi-test-bench-PC scenarios yet, but I wonder if you will break compatibility in order to support them.

So far, we haven’t received any feedback or customer use cases that require support for multi-test-bench PC setups. As a result, we haven’t yet thought or designed a solution for this scenario. However, we did align at a high level that introducing support in the future may involve breaking compatibility in certain areas.

> Do the client APIs for other programming languages also do this?

Yes, we followed the same approach for clients in both C# and LabVIEW to ensure consistency across tech stacks.

> I think adding the deployment target as a constructor parameter to the PinMapClient and SessionManagementClient classes would be a better API.

According to the current design, we treat the response from the `enumerate_compute_nodes` API as the source of truth for determining the status of the remote feature toggle. That is, `enumerate_compute_nodes` will return remote compute nodes only when the "Remote test bench" toggle is enabled (without considering the 'remote compute nodes' feature).

If we begin accepting these addresses through client class constructors, we might lose the ability to enforce this constraint. For instance, `PinMapClient` is usually used alongside measurement clients, and users could pass any address, regardless of whether it was added via InstrumentStudio. Additionally, `SessionManagementClient` is instantiated in all our `teststand_nifoo.py` example files. This could result in bypassing the feature toggle check and cause inconsistent behavior.

> Putting node configuration in the .env file may also be an option.

This feature is already controlled by a feature toggle exposed via an environment variable. Introducing an additional configuration variable specifically for Python may not offer a good user experience.

bkeryan: > This feature is already controlled by a feature toggle exposed via an environment variable. 

Okay, I did not know about this. The fact that this feature is opt-in resolves my concern.

> There was some prior discussion about throwing an exception if more than one remote compute node exists vs. reverting back to existing behavior of only targeting the localhost. I had a slight preference for just reverting back to localhost only behavior.

As implemented, the code does neither.

packages/service/ni_measurement_plugin_sdk_service/discovery/_client.py
```diff
@@ -304,3 +307,15 @@ def enumerate_services(self, provided_interface: str) -> Sequence[ServiceInfo]:
...
+
+    def enumerate_compute_nodes(self) -> Sequence[ComputeNodeDescriptor]:
+        """Enumerates all the compute nodes registered with the discovery service.
+
+        Returns:
+            The list of information describing the compute nodes.
+        """
+        request = discovery_service_pb2.EnumerateComputeNodesRequest()
+
+        response = self._get_stub().EnumerateComputeNodes(request)
```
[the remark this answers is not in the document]
nick-beer: @bkeryan - `EnumerateComputeNodes` has always effectively been behind a toggle.  With the toggle disabled, it's never returned more than the current, local instance of DiscoveryService.  I'm not sure what you're referencing when you ask "Is 2024 Q2 still supported"?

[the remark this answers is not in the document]
bkeryan: Oops, I was off by a quarter. EnumerateComputeNodes does not exist in ASW releases/23C4 (a.k.a. 2024Q1). Do we support using the latest Python package with the DiscoveryService from 2024Q1?

PR #1216, opened 2025-08-18
Consume ni.measurementlink.discovery.v1.proto package and remove _internal.stubs for discovery

dixonjoel:

Pull in the new ni.measurementlink.discovery.v1.proto package and eliminate the copy of the grpc stubs in this repo.

Single source the generated grpc Python code.

Existing tests


packages/generator/pyproject.toml
```diff
@@ -32,6 +32,7 @@ black = ">=24.8.0"
 click-option-group = ">=0.5.6"
 ni-measurement-plugin-sdk-service = {version = "^2.2.0"}
 ni-protobuf-types = { version = ">=0.1.0dev2", allow-prereleases = true }
+ni-measurementlink-discovery-v1-proto = { version = ">=0.1.0dev0", allow-prereleases = true }
```
mjohanse-emr, line 35: Is this dependency needed in `generator`? I doesn't look like you changed anything in the source code to use it.

packages/generator/poetry.lock
csjall, line 1: Regen/delete this lock file.

packages/service/ni_measurement_plugin_sdk_service/discovery/_client.py
csjall, line 1: In a future PR we want to move this discovery client from measurement-plugin-python repo to ni-apis-python repo. I did no realize that we have imports (SERVICE_PROGRAMMINGLANGUAGE_KEY) other than the stub files. What is the plan to ensure we don't have a circular dependency between measurement-plugin-python repo and ni-apis-python repo. 

Options:
1. We move the annotations to shared location
2. We hard code the SERVICE_PROGRAMMINGLANGUAGE_KEY in ni-apis-python
3. ???

csjall (approved):

packages/service/ni_measurement_plugin_sdk_service/discovery/_client.py
bkeryan: I recommend adding the annotation constants as a public submodule of `ni.measurementlink.discovery.v1.proto` because they can be used by both clients and servers.

These are also a problem:
```
from ni_measurement_plugin_sdk_service.measurement.info import (
    MeasurementInfo,
    ServiceInfo,
)
```
- MeasurementInfo is measurement-specific and is only used by a deprecated method. 
  - The deprecated method could be removed. It was deprecated before the measurement link -> measurement plug-in rename, so I don't know if you would need to bump major version.
  - You can keep the deprecated method around by having the measurement plug-in service package provide a DiscoveryClient that subclasses the extracted DiscoveryClient.
- ServiceInfo is discovery-specific and probably belongs in `ni.measurementlink.discovery.v1.proto` or `ni.measurementlink.discovery.v1.client` (I'm not sure which).

dixonjoel: Also, at least the _client.py contents would go into a new `ni.measurementlink.discovery.v1.client` package right? And the other stuff may go into the .client or .proto package depending what we decide.

bkeryan: > Also, at least the _client.py contents would go into a new `ni.measurementlink.discovery.v1.client` package right? And the other stuff may go into the .client or .proto package depending what we decide.

Yes, I agree with this.

dixonjoel: Keeping this comment in mind for a future ni.measurementlink.discovery.v1.client package - no action right now.

packages/generator/poetry.lock
dixonjoel: Not intending any action here.

packages/service/ni_measurement_plugin_sdk_service/_internal/stubs/ni/measurementlink/discovery/v1/__init__.py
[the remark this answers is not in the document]
dixonjoel: I don't see or know of any external usages so I'm not planning to redirect them.

packages/generator/poetry.lock
bkeryan, line 1: This one too

bkeryan (changes requested):

packages/generator/poetry.lock
dixonjoel: Updated. I don't know why the 'poetry lock --no-update' seems to have updated the version of ni-measurement-plugin-sdk-service.

PR #1304, opened 2026-05-13
poetry.toml: Set min-release-age to 2 weeks

bkeryan:

Configure Poetry to use a minimum release age of 2 weeks, the same as in https://github.com/ni/python-renovate-config

Update poetry.lock to reference Poetry 2.4.0 or above, which is when the `solver.min-release-age` setting was added.

`poetry update` doesn't support specifying a version number, which causes problems when Renovate tries to use it to update to a version that is not the latest. Configuring both Renovate and Poetry to use the same minimum release age should improve this behavior. It should also enforce the minimum release age for lockFileMaintenance.

I tested the approach by running Renovate in a personal fork of the nitypes-python repo.


packages/sdk/poetry.toml
```diff
@@ -1,2 +1,6 @@
 [virtualenvs]
-in-project = true
\ No newline at end of file
+in-project = true
+
+[solver]
+# Set min-release-age to 2 weeks, same as in https://github.com/ni/python-renovate-config
+min-release-age = 14
```
dixonjoel: You mentioned in one thread restricting this to external dependencies and allowing internal dependencies to be consumed right away. Is that necessary here and if not why not?

bkeryan: Thanks, I forgot about that.

bkeryan: Fixed

dixonjoel (approved):

bkeryan:
Bypassing rules because the py32 runner failed to start.

